PR #81, opened 2020-11-18
feat: add CentOS and RHEL infra recipes

jbeveland27:


Julien4218 (commented):
@jbeveland27 I've deployed a redhat 8 and centos 8. Here are few notes:
- I notice the /etc/newrelic-infra.yml is getting multiple license_key entries if I redeploy. We should probably make the execution idempotent.
- It's unclear how the `os` variant value will be used, but one potential downside, if it's matching on exact value, the recipe would only be looked up for those precise centos/redhat version (7 and 8). However, I suspect the recipe would just work when the next version of centos/redhat (let's say 9) comes along. Should we just have `os` has the linux name, and leverage another field for the actual version?
- The test definition file is only installing the infra-agent. Should we have a separate test definition file for installing flex and logs, or should this be bundled in the same test definition file for centos/redhat as well? We may have to make the other test definition files consistent if we chose not to for those linux versions?

jbeveland27:
> @jbeveland27 I've deployed a redhat 8 and centos 8. Here are few notes:
> 
> * I notice the /etc/newrelic-infra.yml is getting multiple license_key entries if I redeploy. We should probably make the execution idempotent.

Good catch - fixed the idempotency issue.

> * It's unclear how the `os` variant value will be used, but one potential downside, if it's matching on exact value, the recipe would only be looked up for those precise centos/redhat version (7 and 8). However, I suspect the recipe would just work when the next version of centos/redhat (let's say 9) comes along. Should we just have `os` has the linux name, and leverage another field for the actual version?

I looked at this from the other direction in that I _know_ it works on 7/8, but less so on 5, 6, or future versions. IMO I'd rather be more explicit with the `os` being a `distro-version` pair.

> * The test definition file is only installing the infra-agent. Should we have a separate test definition file for installing flex and logs, or should this be bundled in the same test definition file for centos/redhat as well? We may have to make the other test definition files consistent if we chose not to for those linux versions?

These were intentionally removed to keep these deploy configs focused on just testing the infra recipe.

PR #84, opened 2020-11-20
Adding recipe for OpenSuse and infrastructure agent install

Julien4218:


jbeveland27 (approved):
The only thing I might remove is the logs and flex recipes from the deploy configs:

```
"https://raw.githubusercontent.com/newrelic/open-install-library/main/recipes/newrelic/logs/logs.yml",
"https://raw.githubusercontent.com/newrelic/open-install-library/main/recipes/newrelic/flex/flex.yml"
```

My thinking is these recipes should get tested in separate deploy configs once they're more fully developed (the existing version of these recipes were more so contrived for demo purposes).

Julien4218:
> The only thing I might remove is the logs and flex recipes from the deploy configs:
> 
> ```
> "https://raw.githubusercontent.com/newrelic/open-install-library/main/recipes/newrelic/logs/logs.yml",
> "https://raw.githubusercontent.com/newrelic/open-install-library/main/recipes/newrelic/flex/flex.yml"
> ```
> 
> My thinking is these recipes should get tested in separate deploy configs once they're more fully developed (the existing version of these recipes were more so contrived for demo purposes).

I could see several tests, some specific to the targeted recipes, and some others more integrated where we tests all the recipes that will be installed. I don't really want to maintain multiple set of definition files at this time given a linux distro and all that can be installed, and I don't want to lose the coverage to make sure all the pieces install correctly at this time either.
This will likely change overtime, I suspect as soon as we can just run the cli to install all the recipes that can be installed.
Would it be ok to proceed as-is for now and revisit later?

jbeveland27:
Yep, that makes sense 👍

PR #96, opened 2020-12-03
Adding nginx suse recipe and test

Julien4218:


recipes/newrelic/nginx/suse.yml
```diff
@@ -0,0 +1,123 @@
...
+      - /net.connectionsDroppedPerSecond/
+      - /net.connectionsReading/
+      - /new.connectionsWaiting/
+      - /net.connectionsWriting/
+      - /net.requestsPerSecond/
+      - /software.edition/
+      - /software.version/
+  logging:
+    pattern:
+      - /http/
```
zlesnr, line 36: I think just `http` is sufficient here.  The CLI doesn't yet implement this, but I'm expecting that we pass the string directly to infra agent config.

Julien4218: Hey @zlesnr , thanks for the feedback.
Yeah I actually have no idea about all those meltMatch data. I've copied it from another definition I believe. I'll cut most of it out.

jbeveland27 (approved):
Awesome work @Julien4218! I'll address the `meltMatch`/`logMatch` changes in an upcoming PR, so this is good to go :shipit:

PR #98, opened 2020-12-03
chore(spec): update recipe-spec for logMatch and remove meltMatch

jbeveland27:


recipes/newrelic/nginx/debian.yml
```diff
@@ -16,24 +16,25 @@ keywords:
...
-meltMatch:
-  logging:
-    pattern:
-      - /http/
-    files:
-      - /var/log/nginx/
+# Matches partial list of the Log forwarding parameters
+logMatch:
+  - name: nginx
+    file: /var/log/nginx/access.log*  # assumes log rotation like access.log.0, access.log.1, etc.
```
Julien4218, line 26: Can we just specify the folder instead of all the files in the folder? `file: /var/log/nginx`

recipes/newrelic/infra-agent/amazonlinux2.yml
```diff
@@ -20,28 +20,13 @@ keywords:
...
-      - /system.cpu.usage/
-  # logging:
-  #   pattern:
-  #     - /http/
-  #   files:
-  #     - /var\/log\/system.log
-
 validationNrql: "SELECT count(*) from SystemSample where hostname like 'HOSTNAME' SINCE 10 minutes ago"
 
+# go-task yaml definition goes here
```
Julien4218, line 25: We dont have any logMatch for the regular infra-agent recipes? I thought we were going to remove the log recipe potentially, maybe another story?

Julien4218 (approved):

recipes/newrelic/nginx/debian.yml
jbeveland27: This is specifically targeted at access logs because of the use of `logtype: nginx` for ruleset parsing. Likewise there's another logMatch entry under this for `nginx-error`.

recipes/newrelic/infra-agent/amazonlinux2.yml
jbeveland27: That's correct, no `logMatch` for the infra-agent recipes. Rishi had a thought to use `logMatch` for Docker use case, but my preference is to omit it from infra-agent recipes until we deem it necessary (and I think we should have a separate recipe for Docker use case). 

The CLI will kick off the Log recipe after it receives a list of `logMatch` from the recipe service, so that will still hang around.

PR #99, opened 2020-12-04
Cassandra recipe

setamyDG:
still in progress.


recipes/newrelic/cassandra/amazonLinux2.yml
```diff
@@ -0,0 +1,118 @@
+name: Cassandra Open Source Integration
```
Julien4218, line 1: I think this file should be named "rhel.yml" since the platform it targets covers all the red hat linux platforms (amazon, centos, redhat).

recipes/newrelic/cassandra/amazonLinux2.yml
```diff
@@ -0,0 +1,118 @@
...
+  - Cassandra
+
+# Examine Infrastructure events for correlated data
+processMatch:
+  - cassandra
+
+# Matches partial list of the Log forwarding parameters
+logMatch: 
+  - name: cassandra
+    file: /var/log/cassanra/access.log*
```
Julien4218, line 30: typo for `cassanra`

recipes/newrelic/cassandra/amazonLinux2.yml
```diff
@@ -0,0 +1,118 @@
...
+        - |
+          sudo mkdir -p "/etc/newrelic-infra/integrations.d"
+        - |
+          sudo yum install nri-cassandra -y
+        - |
+          if [ -f /etc/newrelic-infra/integrations.d/cassandra-config.yml ]; then
+            sudo rm /etc/newrelic-infra/integrations.d/cassandra-config.yml;
+          fi
+
+          sudo touch /etc/newrelic-infra/integrations.d/cassnadra-config.yml;
```
Julien4218, line 83: filename typo 'cassnadra'

recipes/newrelic/cassandra/debian.yml
```diff
@@ -0,0 +1,116 @@
...
+       - |
+         sudo apt-get update
+       - |
+         sudo apt-get install nri-cassandra -y
+       - |
+          if [ -f /etc/newrelic-infra/integrations.d/cassandra-config.yml ]; then
+            sudo rm /etc/newrelic-infra/integrations.d/cassandra-config.yml;
+          fi
+
+          sudo touch /etc/newrelic-infra/integrations.d/cassnadra-config.yml;
```
Julien4218, line 81: typo

recipes/newrelic/cassandra/suse.yml
```diff
@@ -0,0 +1,113 @@
...
+       - |
+         sudo zypper -n ref -r newrelic-infra
+       - |
+         sudo zypper -n install nri-cassandra
+       - |
+          if [ -f /etc/newrelic-infra/integrations.d/cassandra-config.yml ]; then
+            sudo rm /etc/newrelic-infra/integrations.d/cassandra-config.yml;
+          fi
+
+          sudo touch /etc/newrelic-infra/integrations.d/cassnadra-config.yml;
```
Julien4218, line 78: typo

test/definitions/ohi/linux/cassandra-debian.json
```diff
@@ -0,0 +1,63 @@
...
+        }
+      },
+      {
+        "id": "nr_infra_is_having_data",
+        "resource": ["host1"],
+        "provider": "newrelic",
+        "source_repository": "https://github.com/newrelic/open-install-library.git",
+        "deploy_script_path": "test/deploy/assertions/recipe-is-valid/roles",
+        "params": {
+          "nrql_query": ""
```
Julien4218, line 48: check the most recent nginx test definition, we should use a params element to use the recipe validationNrql https://github.com/newrelic/open-install-library/blob/main/test/definitions/ohi/linux/nginx-debian.json#L49

test/definitions/ohi/linux/cassandra-debian.json
```diff
@@ -0,0 +1,63 @@
...
+        }
+      },
+      {
+        "id": "nr_infra_is_reporting",
```
Julien4218, line 52: This validation is not needed, you can remove this whole json section (line 51 to 60)

test/deploy/linux/cassandra/install/debian/roles/configure/tasks/main.yml
```diff
@@ -0,0 +1,26 @@
...
+  shell: "apt update"
+  become: true
+
+- name: Install cassandra
+  shell: "apt install cassandra -y"
+  become: true
+
+- name: Configure cassandra default route with status module
+  blockinfile:
+    path: /etc/cassandra/sites-enabled/default
```
Julien4218: This looks like nginx configuration, is this really how to configure cassandra?

setamyDG:
@Julien4218 @devfreddy We're experiencing an odd issue with the cassandra debian recipe. While running it from a local deployer it goes all the way to assert entity check (which fails, I suppose there might be something wrong still) but when we push it to github, it fails on login to cassandra check (which uses default cassandra values for login and works from local deployer). Do you have any inkling as to what might be causing this issue?

Julien4218:
> @Julien4218 @devfreddy We're experiencing an odd issue with the cassandra debian recipe. While running it from a local deployer it goes all the way to assert entity check (which fails, I suppose there might be something wrong still) but when we push it to github, it fails on login to cassandra check (which uses default cassandra values for login and works from local deployer). Do you have any inkling as to what might be causing this issue? There is also an error what we receiving after the test pass:
> image

@setamyDG I'll try to repro local to see what I find.
Have you seen all the comments I've put inline in the review of this PR? There are a few things needed to be updated.

setamyDG:
@Julien4218 The resolution for this particular issue was to add host and port to the cqlsh command

test/definitions/ohi/linux/cassandra-debian.json
```diff
@@ -0,0 +1,53 @@
...
+    "user_name": "admin",
+    "ami_name": "SupportedImages debian-10-amd64-*"
+  }],
+
+  "services": [{
+    "id": "cassandra1",
+    "destinations": ["host1"],
+    "source_repository": "https://github.com/newrelic/open-install-library.git",
+    "deploy_script_path": "test/deploy/linux/cassandra/install/debian/roles",
+    "port": 80
```
Julien4218, line 24: I don't believe we need port 80 to be opened for Cassandra. You should use the port 9999 instead (port is a required field, the value of 9999 is equivalent to no-port-required for this service).

test/definitions/ohi/linux/cassandra-linux2.json
```diff
@@ -0,0 +1,52 @@
...
+    "size": "t3.small",
+    "ami_name": "amazonlinux-2-base*"
+  }],
+
+  "services": [{
+    "id": "cassandra1",
+    "destinations": ["host1"],
+    "source_repository": "https://github.com/newrelic/open-install-library.git",
+    "deploy_script_path": "test/deploy/linux/cassandra/install/linux2/roles",
+    "port": 80
```
Julien4218, line 23: I don't believe we need port 80 to be opened for Cassandra. You should use the port 9999 instead (port is a required field, the value of 9999 is equivalent to no-port-required for this service).

test/definitions/ohi/linux/cassandra-suse.json
```diff
@@ -0,0 +1,52 @@
...
+    "size": "t2.small",
+    "ami_name": "openSUSE 15.2 Leap*"
+  }],
+
+  "services": [{
+    "id": "cassandra1",
+    "destinations": ["host1"],
+    "source_repository": "https://github.com/newrelic/open-install-library.git",
+    "deploy_script_path": "test/deploy/linux/cassandra/install/suse/roles",
+    "port": 80
```
Julien4218, line 23: I don't believe we need port 80 to be opened for Cassandra. You should use the port 9999 instead (port is a required field, the value of 9999 is equivalent to no-port-required for this service).

setamyDG:
@devfreddy @Julien4218  Regarding to the cassandra-suse I think it's close to end and I would appreciate any help. I have an issue like before, about that I cannot login to cassandra with default root username and password. I already tried a commands with the "wait_for" section and the loop what @Julien4218 included into debian. It receives the error "...Timeout...". Actually when I run the same ansible-playbook locally on the virtual machine, everything works correctly. It is probably again some port issue or there is too small ec2 instance... I'm not sure. Please take a look on this in your free time. Feel free to left some comment or push changes. Thank you in advance. 

GH check error: 
image

VM ansible-playbook
image

Julien4218 (approved):

PR #109, opened 2020-12-15
chore: include log validationNrql

zlesnr:


recipes/newrelic/logs/logs.yml
```diff
@@ -14,6 +14,8 @@ keywords:
 processMatch:
   - /logs/
 
+validationNrql: "SELECT count(*) from Log where hostname like '{{.HOSTNAME}}' SINCE 10 minutes ago"
```
Julien4218, line 17: We found out sometimes (for aws linux2), hostname is being replaced by short form, but the reported host name can be long form. So if you just add a '%' after the 2 closing brackets, that takes care of the issue.

Julien4218 (changes requested):
Just need a wildcard character after hostname

recipes/newrelic/logs/logs.yml
ctrombley: Hopefully in the future we can switch over to agent/entity ID for a more precise validation.  If this works for now then :shipit:

ctrombley (approved):

recipes/newrelic/logs/logs.yml
Julien4218: Would you like me to make the change?

PR #123, opened 2020-12-23
Apache Ohi recipe

JakubKotkowiak:


recipes/newrelic/apache/debian.yml
```diff
@@ -0,0 +1,117 @@
...
+        
+    assert_pre_req:
+      cmds:
+        - |
+          SERVICE_EXIST=$(sudo systemctl status newrelic-infra.service | grep "Active" | wc -l)
+          if [ $SERVICE_EXIST -eq 0 ]; then
+            echo "The newrelic-infra agent service is NOT installed on the host, but is required to install this integration." >> /dev/stderr
+            exit 1
+          fi
+
```
Julien4218, line 63: @JakubKotkowiak similarly to nginx, there is a statusURL needed so the integration can work properly.
Could you add a pre-req that test the statusURL responds correctly, and if not echo an explicit error message and return a different exit code?

recipes/newrelic/apache/rhel.yml
```diff
@@ -0,0 +1,103 @@
...
+        
+    assert_pre_req:
+      cmds:
+        - |
+          SERVICE_EXIST=$(sudo systemctl status newrelic-infra.service | grep "Active" | wc -l)
+          if [ $SERVICE_EXIST -eq 0 ]; then
+            echo "The newrelic-infra agent service is NOT installed on the host, but is required to install this integration." >> /dev/stderr
+            exit 1
+          fi
+
```
Julien4218, line 66: @JakubKotkowiak similarly to nginx, there is a statusURL needed so the integration can work properly.
Could you add a pre-req that test the statusURL responds correctly, and if not echo an explicit error message and return a different exit code?

Julien4218 (approved):

PR #132, opened 2021-01-13
Jmx recipe

setamyDG:


jbeveland27 (dismissed):
Leaving a note to say we should revisit allowing the `jvm-metrics.yml` file to be written interactively in the future. Otherwise LGTM 🚀

jbeveland27 (approved):

PR #140, opened 2021-01-20
feat(debian): add redis debian recipe

JakubKotkowiak:


jbeveland27 (commented):
@JakubKotkowiak - Question: Can the redis.conf template be copied over after redis is installed from apt? As in, is a sample config downloaded that we can just use or are there edits in that file? I worry having the config here in the repo might make it go stale over time as installing redis with the package manager will always pull the latest version.

jbeveland27 (approved):
LGTM 🚀

PR #145, opened 2021-01-22
Feat/docker log recipe

Julien4218:


.github/workflows/validation.yaml
```diff
@@ -111,12 +111,12 @@ jobs:
...
-                    contents.instrumentations.resources[0].params.recipe_content_url !== null)
-                {
-                  // Look at current recipe_content_url. If it contains any of recipeFiles, add this to the current accumulator
-                  recipes = contents.instrumentations.resources[0].params.recipe_content_url;
+                if (contents.instrumentations.resources) {
+                  contents.instrumentations.resources.forEach(resource => {
+                    if (resource.params && resource.params.recipe_content_url) {
+                      recipes = recipes.concat(resource.params.recipe_content_url);
+                    }
+                  });
```
jbeveland27, line 119: Ah, good catch!

recipes/newrelic/logs/logs.yml
```diff
@@ -58,4 +58,8 @@ install:
             fi
           done
         - echo -e "{{.DISCOVERED_LOG_FILES}}" | sudo tee /etc/newrelic-infra/logging.d/discovered.yml > /dev/null
+        - |
+          if [ -d /var/lib/docker/containers ]; then
+              echo -e "  - name: docker-logs\n    file: /var/lib/docker/containers/*/*.log" | sudo tee -a /etc/newrelic-infra/logging.d/logging.yml > /dev/null
+          fi
```
jbeveland27, line 64: @Julien4218 This works great. However, should we do something to indicate to the user that we're grabbing Docker logs? The way this runs now, there's nothing to indicate what we've done with adding this log entry to `logging.yml`. I wonder if `DOCKER_LOGS` should be a prompted inputVar, with `/var/lib/docker/containers/*/*.log` being the default, so that the user has to acknowledge that we're grabbing the Docker Logs prior to running the recipe.

Julien4218: Interesting, this wasn't specified though. I'll check with rishi.

.github/workflows/validation.yaml
Julien4218: Yeah I bumped into that a few times, so I thought it would be great to just fix.

recipes/newrelic/logs/logs.yml
Julien4218: @jbeveland27 can you check this PR, I've made the changes I believe you mentioned.

jbeveland27: Thanks @Julien4218. The only issue here is that the `docker-logs` recipe wasn't actually getting run because there was no match on the `recipe_content_url` in the test definition file. I added a param for hooking into the validation called `ci_validation_url` (https://github.com/newrelic/open-install-library/pull/145/commits/2db93bc74eb48f7db34f340a3f64a8982b697f0b), but then realized this isn't really any better, as process detection runs and the CLI can only pull recipes that are already in the service. Seems we have a bit of a chicken and egg problem with testing recipes that are reliant on process detection to work 👎🏼 

Without modifying the CLI to run process detection AND choose a specific recipe to run for a given process, I'm not sure of a feasible way to specifically target testing situations like this. However, it's pretty reasonable to assume that this works the way it's intended. I'll approve and we can brainstorm if it's worth trying to address testing this use case in a better way.

jbeveland27 (approved):
🚀

recipes/newrelic/logs/logs.yml
Julien4218: Correct @jbeveland27 I went through the same thought process, and it's not an easy thing to make it testable on a feature branch. Given the non-regression should test it, I've settled for that in the meantime.

PR #154, opened 2021-01-29
feat: add ohi-auto-config tests

jbeveland27:


jbeveland27:
Ran non-regression tests locally and they passed, so am reasonably confident that this works as intended.

PR #155, opened 2021-01-29
fix: use -y to accept prompts

jbeveland27:
Recent changes to the CLI messaging cause a situation where the deployer tests get stuck waiting for the user to confirm y/n to install integrations (even when `-c` flag is passed). 

image

This change uses the `-y` flag to bypass that restriction until we get a proper fix into the CLI (the original intent of the `-c` flag is prompt wouldn't be required)

### DIFF
--- a/test/deploy/linux/newrelic-cli/install-recipe/roles/configure/tasks/main.yml
+++ b/test/deploy/linux/newrelic-cli/install-recipe/roles/configure/tasks/main.yml
@@ -46,7 +46,7 @@
   shell: "curl -Ls https://raw.githubusercontent.com/newrelic/newrelic-cli/master/scripts/install.sh | sudo bash"
 
 - name: Run recipes
-  shell: "{{ env_vars }} /usr/local/bin/newrelic install -c {{ item }}"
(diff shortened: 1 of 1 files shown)


jbeveland27:
More detail:

[installRecipes](https://github.com/newrelic/newrelic-cli/blob/410039c5a4e644913f6d9e56e66d1f85dabd7682/internal/install/recipe_installer.go#L214) contains a check for [ShouldPrompt()](https://github.com/newrelic/newrelic-cli/blob/410039c5a4e644913f6d9e56e66d1f85dabd7682/internal/install/recipe_installer.go#L252), which is returning true. Then this drops into [userAcceptsInstall](https://github.com/newrelic/newrelic-cli/blob/410039c5a4e644913f6d9e56e66d1f85dabd7682/internal/install/recipe_installer.go#L501), which is springing the [Would you like to enable](https://github.com/newrelic/newrelic-cli/blob/410039c5a4e644913f6d9e56e66d1f85dabd7682/internal/install/recipe_installer.go#L511) message on the user. Setting `-y` short-circuits the y/n prompt, but we need to consider if it should even show in the first place.

PR #170, opened 2021-02-03
Recipes for installing .net agent (core) on linux

Julien4218:


recipes/newrelic/apm/.net/linux/net-debian.yml
```diff
@@ -0,0 +1,76 @@
...
+        - |
+          AGENTPROFILE=/etc/profile.d/newrelic-netcore20-agent-path.sh
+          if test -f "$AGENTPROFILE"; then
+              sudo rm $AGENTPROFILE
+          fi
+
+    install:
+      cmds:
+        - echo 'deb http://apt.newrelic.com/debian/ newrelic non-free' | sudo tee /etc/apt/sources.list.d/newrelic.list
+        - wget -O- https://download.newrelic.com/548C16BF.gpg | sudo apt-key add -
```
Julien4218, line 57: could we use curl instead of wget, we already use curl when we instruct the user to install the CLI, and wget might not be installed

recipes/newrelic/apm/.net/linux/net-debian.yml
```diff
@@ -0,0 +1,76 @@
...
+        - echo "New Relic .NET agent installed"
+
+    configure:
+      cmds:
+        - |
+          echo "Setup startup environment variables"
+          echo 'export NEW_RELIC_LICENSE_KEY={{.NEW_RELIC_LICENSE_KEY}}' | sudo tee -a /etc/profile.d/newrelic-netcore20-agent-path.sh
+          
+          echo "Setup session environment variables for child processes"
+          export NEW_RELIC_LICENSE_KEY={{.NEW_RELIC_LICENSE_KEY}}
```
Julien4218, line 69: I think those export need to be written into the ~/.bashrc file, so they persist on the next session. You can see an example here https://github.com/newrelic/open-install-library/blob/main/test/deploy/linux/haproxy/install/debian/roles/configure/tasks/main.yml#L36

recipes/newrelic/apm/.net/linux/net-debian.yml
jaffinito: This was pulled from our install docs. Sure, I can change that.

recipes/newrelic/apm/.net/linux/net-debian.yml
jaffinito: I was not sure we wanted to write those out to a place that loads them all be default.  We don't recommend enabling the agent to instrument all .NET Core apps globally. We do this due to how the Core parts of the profiler are designed.  On the Framework side, we require some configuration changes to tell the profiler to explicitly instrument apps outside of IIS.  The profiler loads briefly, checks if it should be there, and if not it unloads.  For .NET Core, this was never implemented - to keep it simple - so the profiler will stay loaded, watching methods, but never instrumenting them.

Julien4218: Ho, I don't quite fully understand, but here is what I know.
Those export would be gone the moment the recipe install is completed (like they were not issued).
To make those persistent those should be in the ~/.bashrc. And also, just noticed this morning, if this is done, it should be idempotent (only written at most once in the file).
I'll check if the .net agent instrument the demo apps, and will get back to you.

PR #183, opened 2021-02-09
Ensure IIS and .net 4.5 are installed when attempting to install .net apm agent

Julien4218:


Julien4218:
@jaffinito could you review this PR, I've added the pre-req step to ensure IIS and .net 4.5 are installed.

jaffinito (approved):
That looks great!

PR #196, opened 2021-02-12
feat(infra): align infra agent install

jbeveland27:
Also aligns the infra agent recipes to consistently use quiet/silent flags for their commands (but not full `stdout` suppression).

### DIFF
--- a/recipes/newrelic/infrastructure/ohi/mysql/debian.yml
+++ b/recipes/newrelic/infrastructure/ohi/mysql/debian.yml
@@ -55,26 +55,26 @@ inputVars:
   - name: "NR_CLI_DATABASE"
     prompt: "MySQL Database Name"
 
-preInstall:
-  prompt: |2
-      You'll need to have a MySQL user with replication permissions for enabling
-      this integration. If you'd like to create a new user please use these 
-      commands as reference:
-
-      > sudo mysql -e "CREATE USER 'newrelic'@'localhost' IDENTIFIED BY 'YOUR_SELECTED_PASSWORD';"
-      > sudo mysql -e "GRANT REPLICATION CLIENT ON *.* TO 'newrelic'@'localhost' WITH MAX_USER_CONNECTIONS 5;"
-      > sudo mysql -e "GRANT SELECT ON *.* TO 'newrelic'@'localhost' WITH MAX_USER_CONNECTIONS 5;"
-
 install:
+
   version: "3"
   silent: true
 
   tasks:
     default:
       cmds:
+        - task: create_user_notification
         - task: assert_pre_req
         - task: setup
         - task: restart
+
+    create_user_notification:
+      cmds:
+        - |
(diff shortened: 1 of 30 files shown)


recipes/newrelic/infrastructure/ohi/mysql/debian.yml
```diff
@@ -85,7 +85,7 @@ install:
             exit 1
           fi
         - |
-          CHECK_DB=$(mysql -u{{.NR_CLI_DB_USERNAME}} --port {{.NR_CLI_DB_PORT}} -p{{.NR_CLI_DB_PASSWORD}} -e "SHOW GRANTS FOR CURRENT_USER" | grep 'GRANT SELECT' | awk '{print $2, $3}')
+          CHECK_DB=$(mysql -u{{.NR_CLI_DB_USERNAME}} --port {{.NR_CLI_DB_PORT}} -p{{.NR_CLI_DB_PASSWORD}} -e "SHOW GRANTS FOR CURRENT_USER" 2>&1 | grep -v "mysql: [Warning] Using a password on the command line interface can be insecure." | grep 'GRANT ALL\|SELECT' | awk '{print $2, $3}')
```
jbeveland27, line 88: @gabeobrien @Julien4218 In the course of doing the `assert_pre_req` task, we call mysql from the command line and validate the user's grants. That causes this message to be displayed:

`mysql: [Warning] Using a password on the command line interface can be insecure.`

I added this code to suppress that message from the user. Does that seem appropriate or should we let the warning fly?

image

jbeveland27: This is done in the rhel and suse recipes as well.

Julien4218 (commented):
@jbeveland27 I think all the quieting of the output has to be run through rishi. I believe his expectation is now different than what the mockups are initially showing. We actually talked about this Tuesday.

recipes/newrelic/infrastructure/ohi/mysql/debian.yml
gabeobrien: Seems like suppression the message from MySQL is a good call, since it displays awkwardly (no line break) and we do not have control over what the message is.
Could we add our own message that mimics the warning?  That way we surface an issue that the user might care about, but we control what that message is and how it is displayed.

gabeobrien:
I haven't bee around for the history of what should or should not be logged.
I think for most users 'less is better' with regards to output.  
But for debugging/development/support tickets having the full output of all the steps could be helpful.

recipes/newrelic/infrastructure/ohi/mysql/debian.yml
jbeveland27: Since we have control of stdout in the recipe, we could do something like:

```
Verifying MySQL requirements...
Checks passed.
```

jbeveland27:
I've broken out the prompt changes to a separate PR so this is only focused on aligning the Infra Agent install to be consistent amongst the linux distros.

Julien4218 (approved):

PR #198, opened 2021-02-16
feat(mysql): preInstall info

jbeveland27:
* Update spec and schema validation to use `preInstall:info`
* Update MySQL recipes


devfreddy (approved):
LGTM 👍🏼

PR #207, opened 2021-02-22
fix(apache): adjust the systemctl command for Ubuntu hosts

zlesnr:


jbeveland27 (approved):
LGTM 🚀

recipes/newrelic/infrastructure/ohi/apache/debian.yml
```diff
@@ -61,8 +61,8 @@ install:
     assert_pre_req:
       cmds:
         - |
-          SERVICE_EXIST=$(sudo systemctl status newrelic-infra.service | grep "Active" | wc -l)
-          if [ $SERVICE_EXIST -eq 0 ]; then
+          systemctl is-active --quiet newrelic-infra
```
Julien4218, line 64: I assume this was failing before the grep "Active" was not latching correctly for Ubuntu based host.
If that's the case, we'll want to have this updated on all the Debian OHI recipes.

jbeveland27:
For ref, here's a Stack Overflow thread giving context to why this is the proposed change: https://unix.stackexchange.com/a/396638

Julien4218 (approved):

PR #209, opened 2021-02-22
feat: add node test definition

jbeveland27:


Julien4218 (approved):
Looks good to me.
You could also add a little cronjob to generate a bit of traffic, see https://github.com/newrelic/demo-catalog/blob/main/catalog/hello/hello.aws.json#L39-L47

jbeveland27:
@Julien4218 Good call - updated to add the cronjob

test/manual/definitions/apm/node/node-linux2.json
```diff
@@ -0,0 +1,37 @@
...
+      "destinations": [
+        "nodelinux2"
+      ]
+    },
+    {
+      "destination_filepath": "engine/cronjob.json",
+      "content": [
+        {
+          "frequency": "* * * * *",
+          "job": "for ((i=1;i<=30;i++)); do curl '[service:app1:url]/'; curl '[service:app1:url]/api/inventory'; sleep 1; done"
```
Julien4218, line 32: there is no `app1` in this config, so the field should be referenced using `nodetron` for the service_id [service:nodetron:url]

jbeveland27: Good catch. Fixed.

Julien4218 (approved):

PR #221, opened 2021-02-25
Remove setup of logs from within the infra recipe since it's done in the windows log recipe file

Julien4218:


jbeveland27 (approved):
LGTM 🚀

PR #229, opened 2021-03-01
Add escape char for windows process match regex

Julien4218:


gabeobrien (approved):
+1 now with more restrictive REGEX!

PR #233, opened 2021-03-05
Debian Oracle

JakubKotkowiak:


JakubKotkowiak:
Currently does not contain Ansible Play for installing OracleDb due to issues with account requirements. [WIP]

PR #235, opened 2021-03-05
Adding debian platform for logs

Julien4218:
It turns out, those special logs are also needed for Debian linux, I just tested on debian 10 and I don't see a /var/log/secure and see a /var/log/auth.log as well as a /var/log/alternatives.log


jbeveland27 (approved):
LGTM 🚀

PR #252, opened 2021-03-09
Java linux2 apm agent

gabeobrien:
This is a POC PR of the Java/APM instrumentation recipe for AWS Linux 2.


jbeveland27:
@ctrombley Would be good to update the [recipe schema](https://github.com/newrelic/open-install-library/blob/main/docs/recipe-spec/recipe-spec.md), [example spec file](https://github.com/newrelic/open-install-library/blob/main/docs/recipe-spec/spec.json), and [validation schema](https://github.com/newrelic/open-install-library/blob/main/validator/schema-v1.json) to reflect the addition of the new `successLink` field. 

[Schema validation](https://github.com/newrelic/open-install-library/blob/main/.github/workflows/validation.yaml#L34-L49) is run on every pull request to make sure the recipe is valid yaml, but the fields are ignored if they're missing.

test/definitions/apm/java/linux/javatomcat-debian.json
```diff
@@ -0,0 +1,56 @@
...
+  
+      }
+    ],
+    "instrumentations": {
+      "resources": [
+        {
+          "id": "javatomcat",
+          "resource_ids": ["javatomcat"],
+          "provider": "newrelic",
+          "source_repository": "-b java_linux2_apm_agent https://github.com/newrelic/open-install-library.git",
```
gabeobrien, line 46: Before this PR gets merged we need to remove all the reference to this branch in the recipes.

Julien4218 (commented):
@gabeobrien I'm not seeing anything that would prevent users from installing the java recipe without the feature flag being checked. Would we want to add some metadata on the recipe about feature flag and have some logic on the service to filter out recipes the user should not be able to install?

gabeobrien:
@Julien4218 I do not know enough about the FF and Recipe Service to really answer.  @michaelkro setup to FF.  @ctrombley knows more about the recipe service.

Julien4218:
BTW, this can also be done in the newrelic-cli, I'm actually not sure if the service should/should not do the FF work, I could see that shifted in the CLI just as well

ctrombley:
@Julien4218 We made a decision to gate the feature at the UI level, not in the recipe service.

recipes/newrelic/apm/java/rhel.yml
```diff
@@ -0,0 +1,66 @@
...
+  - type: host
+    os: linux
+    platform: amazon
+    platformVersion: "2"
+  - type: host
+    os: linux
+    platform: "centos"
+  - type: host
+    os: linux
+    platform: "redhat"
```
jbeveland27, line 19: This looks like a mistake using `type: host`. Think it should be `type: application` like in the Debian recipe.

gabeobrien: I made the changes you asked.  I also update the Debian recipe to have definition that are Debian and not redhat.

https://github.com/newrelic/open-install-library/pull/252/commits/161e2e69817d80bd0b65195e427e9756a94945ad

Julien4218 (changes requested):
Before pushing those changes to main, and be deployed to production since we have no way of stopping them afterward, I believe we should have a mechanism to make sure those recipes are not executed when the FF for java/apm is NOT enabled for the user/account.

We might need to keep a separate recipe for the process detection and recommendation, unless the FF check implementation is scoped to the recipe execution (not the process detection)

Julien4218 (approved):
Skipping FF validation for recipes for now

PR #253, opened 2021-03-09
Fix ubuntu installs for oldest version of ubuntu (<=14.04)

Julien4218:
This change works on the recent ubuntu and older versions

[ubuntu18.log](https://github.com/newrelic/open-install-library/files/6111946/ubuntu18.log)
[ubuntu14.log](https://github.com/newrelic/open-install-library/files/6111948/ubuntu14.log)


Julien4218:
Merging to validate the fix with merged recipe (cli installs from latest currently)

PR #254, opened 2021-03-10
Update infra service dependency

Julien4218:
Detect if infra agent service is started by detecting the process, to be compatible with all linux versions.
Update Debian recipe to restart the infra agent service through init.d if it exists (older ubuntu).


jbeveland27 (approved):
LGTM 🚀 

I think it'll be good to keep in mind that infra check is dependent on the infra-agent being installed, just in case something ends up shifting as we branch out into more areas.

PR #267, opened 2021-03-17
fix(logs): check for empty log files input

sanderblue:


Julien4218 (changes requested):
@sanderblue I like what you've done.
Could you make similar changes to the other logs recipe? I believe we have debian, docker and windows.
Also, could you use an exit code that is not 1? We typically use rc 1 when the infra agent is not installed, which by the way, it looks like those recipes dont check the infra agent is up and running and should.

sanderblue:
Yep, no problem.

sanderblue:
I updated the Debian recipe, but I'm not familiar with our Windows and Docker setups here (plus I'll be out tomorrow), so in the interest of incremental improvements, it might be best to scope this PR to just the Linux distros. Maybe we add an additional ticket to update those recipes so we don't forget. I'd be happy to pick that up on first thing next week 😎

Julien4218:
No problem, I'll continue the work for the remaining recipes

Julien4218 (approved):

PR #270, opened 2021-03-19
Linux2 apache aspnetcore

jaffinito:
- Adds Amazon Linux 2 def with ASP.NET Core and Apache
- Adds new deploy-application section for Apache and RHEL
- deploy-application contains 2 apps, one self contained and one framework dependent
- Both install services and apache confs
- Self contained just runs the executable directly
- Framework dependent uses the dotnet commend to run the DLL

**Note**: The two new apps live in a different S3 than the other apps.  They will need to be moved over at some point and the scripting updated.

### DIFF
--- a/test/manual/definitions/apm/.net/linux/linux2-aspnetcore5.json
+++ b/test/manual/definitions/apm/.net/linux/linux2-aspnetcore5.json
@@ -1,0 +1,51 @@
+{
+  "global_tags": {
+    "owning_team": "OpenSource",
+    "Environment": "development",
+    "Department": "Product",
+    "Product": "Virtuoso"
+  },
+
+  "resources": [{
+    "id": "host1",
+    "display_name": "AwsLinux2ASPNETCore",
+    "provider": "aws",
+    "type": "ec2",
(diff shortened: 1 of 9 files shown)


test/manual/definitions/apm/.net/linux/linux2-aspnetcore5.json
```diff
@@ -0,0 +1,51 @@
...
+    "Department": "Product",
+    "Product": "Virtuoso"
+  },
+
+  "resources": [{
+    "id": "host1",
+    "display_name": "AwsLinux2ASPNETCore",
+    "provider": "aws",
+    "type": "ec2",
+    "size": "t3.nano",
```
Julien4218, line 14: Is nano a little small for everything we do on the box?

jaffinito: No, I don't think it is.  I was able to run everything and it spun up in a reasonable amount of time.

jaffinito:
closing to move from fork to in-repo branch.

PR #272, opened 2021-03-19
Two ASP.NET Core applications, with nginx reverse proxy on Ubuntu 20.4

nr-ahemsath:
Note: this PR parallels @jaffinito's PR #270 

* Adds Ubuntu 20.4 def with ASP.NET Core and Nginx
* Adds new deploy-application section for Nginx and Ubuntu 20.4
* deploy-application contains 2 apps, one self contained and one framework dependent
* Both install systemd services and Nginx configurations
* Self contained just runs the executable directly
* Framework dependent uses the dotnet commend to run the DLL

Note: The two new apps live in a different S3 than the other apps. They will need to be moved over at some point and the scripting updated.


test/deploy/linux/.net/install/ubuntu20/roles/prepare/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
+---
+- debug:
+    msg: Install ASP.NET Core 5.0 Runtime
+
+- name: Add Microsoft package signing key and add Microsoft package repository for Ubuntu 20.4
+  shell: |
+    wget https://packages.microsoft.com/config/ubuntu/20.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb
```
Julien4218, line 7: Is this package available on earlier ubuntu versions? If possible we might like to test on earlier versions.
If it's not supported on earlier ubuntu versions please disregard

test/deploy/linux/nginx/deploy-application/.net/debian/roles/onbeforestart/tasks/main.yml
```diff
@@ -0,0 +1,78 @@
...
+    msg: Install .NET App
+
+- name: Set default is_selfcontained (default to false)
+  set_fact:
+    is_selfcontained: "false"
+  when: is_selfcontained is undefined
+
+- name: Set default download url for self contained web app
+  set_fact:
+    web_app_url: "https://virtuoso-testing.s3-us-west-2.amazonaws.com/selfcontained-net5webapplication.zip"
```
Julien4218, line 12: We need to consolidate those zips on the same S3 account we already use for the other tests: https://s3.console.aws.amazon.com/s3/buckets/open-install-library-artifacts?region=us-west-2

Not sure if you have access to our AWS account, otherwise @gabeobrien can

test/deploy/linux/nginx/deploy-application/.net/debian/roles/onbeforestart/tasks/main.yml
```diff
@@ -0,0 +1,78 @@
...
+  when: web_app_url is undefined and (is_selfcontained|bool)
+
+- name: Set default download url for framwork dependent web app
+  set_fact:
+    web_app_url: "https://virtuoso-testing.s3-us-west-2.amazonaws.com/frameworkdependent-net5webapplication.zip"
```
Julien4218, line 17: We need to consolidate those zips on the same S3 account we already use for the other tests: https://s3.console.aws.amazon.com/s3/buckets/open-install-library-artifacts?region=us-west-2

Not sure if you have access to our AWS account, otherwise @gabeobrien can

test/deploy/linux/nginx/deploy-application/.net/debian/roles/onbeforestart/templates/frameworkdependent-net5webapplication-nginx-site
```diff
@@ -0,0 +1,13 @@
+server {
+    listen        8081;
```
Julien4218, line 2: Can we drive that port number from the deploy config value? @gabeobrien feel free to show the specifics

test/deploy/linux/nginx/deploy-application/.net/debian/roles/onbeforestart/templates/selfcontained-net5webapplication-nginx-site
```diff
@@ -0,0 +1,13 @@
+server {
+    listen        8080;
```
Julien4218, line 2: Can we drive that port number from the deploy config value? @gabeobrien feel free to show the specifics

test/manual/definitions/apm/.net/linux/ubuntu20-aspnetcore.json
```diff
@@ -0,0 +1,50 @@
...
+        "ami_name": "ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-????????.1",
+        "user_name": "ubuntu"
+    }],
+
+    "services": [{
+        "id": "aspnetcore",
+        "destinations": ["ubuntunetcore1"],
+        "source_repository": "https://github.com/newrelic/open-install-library.git",
+        "deploy_script_path": "test/deploy/linux/.net/install/ubuntu20/roles",
+        "port": 9998
```
Julien4218, line 24: 9999 (which is not what is used here, but close) is a special value for the deployer, could we update the port value to be similar to the other 2 services in the 8xxx range

test/manual/definitions/apm/.net/linux/ubuntu20-aspnetcore.json
```diff
@@ -0,0 +1,50 @@
...
+        "Department": "Product",
+        "Product": "Virtuoso"
+    },
+
+    "resources": [{
+        "id": "ubuntunetcore1",
+        "display_name": "NetCoreUbuntu20Host",
+        "provider": "aws",
+        "type": "ec2",
+        "size": "t3.nano",
```
Julien4218, line 14: I'm surprised we can fit so much on a nano, could you double check newrelic to make sure we're not in the 80%+ range for memory usage?

Julien4218 (commented):
Good stuff @nr-ahemsath I've noticed a few things on the PR

test/deploy/linux/.net/install/ubuntu20/roles/prepare/tasks/main.yml
nr-ahemsath: MSFT has different versions of their .deb repository setup package for different versions of the OS: https://docs.microsoft.com/en-us/dotnet/core/install/linux

If we do want to test multiple Ubuntu versions, I think we should stick to LTS versions, so we would only need to add 16.04 and 18.04.

test/manual/definitions/apm/.net/linux/ubuntu20-aspnetcore.json
nr-ahemsath: The `aspnetcore` "service" just installs the .NET runtime, so the port number is arbitrary.

test/deploy/linux/.net/install/ubuntu20/roles/prepare/tasks/main.yml
Julien4218: Ho perfect, yeah if we could do with 16.04 and 18.04 that's more than enough.
So maybe we could update the role path to be /ubuntu instead of /ubuntu20, and in the implementation reflect from the /etc/os-release file to pull the correct version number (18.04) for example, and use that for the wget.
BTW, I think we typically use curl instead of wget... But i think curl is not standard on ubuntu? So perhaps fine to keep wget

Julien4218: I think there is also an ansible command for curl/wget `uri`, which should do the right thing under the hood. https://docs.ansible.com/ansible/latest/collections/ansible/builtin/uri_module.html

test/manual/definitions/apm/.net/linux/ubuntu20-aspnetcore.json
nr-ahemsath: Looks like it's using less than half of the memory:

```
ubuntu@ip-172-31-16-179:~$ free
              total        used        free      shared  buff/cache   available
Mem:         467076      216548       20560         788      229968      225988
Swap:             0           0           0
```

test/deploy/linux/nginx/deploy-application/.net/debian/roles/onbeforestart/tasks/main.yml
nr-ahemsath: I don't appear to have the necessary permissions to add new content to the S3 bucket.  We do want to consolidate the zips, though.

test/manual/definitions/apm/.net/linux/ubuntu20-aspnetcore.json
Julien4218: Ok, so if we don't need that specific port to be opened on the host, I would suggest using the reserved value of 9999. This tells the deployer not to open that port to the public when creating the ec2.

test/deploy/linux/.net/install/ubuntu20/roles/prepare/tasks/main.yml
nr-ahemsath: Given that our current effort on these test aps is a timeboxed spike, I'd prefer to hold off on trying to test different versions.

test/manual/definitions/apm/.net/linux/ubuntu20-aspnetcore.json
nr-ahemsath: Done!

Julien4218:
@nr-ahemsath could you move this work to a branch on the https://github.com/newrelic/open-install-library repository.
There are a couple of reasons. Our UAT tests use some secrets for our AWS account, and those tests cannot run on a PR from another repo (secrets are scoped to the repo).
Also, I wouldn't mind spending the extra time to help, for example updating the ubuntu specifics, which might also work for all Debians. I could do this on a sub-PR and review with you.

Spikes are great, but if we're merging to main on the repo I would say this is no longer a spike and becomes prod quality. 
If we want to keep things in a branch without merging that would be fine. With that said I think this work seems good to me that it should be merged into main.

nr-ahemsath:
@Julien4218 Sounds good, will do.

PR #277, opened 2021-03-22
Validate inputs connect

Julien4218:


recipes/newrelic/infrastructure/ohi/couchbase/debian.yml
```diff
@@ -45,7 +45,7 @@ inputVars:
     prompt: "Couchbase Query Port (default: 8093)"
     default: 8093
   - name: "NR_CLI_API_USE_SSL"
-    prompt: "Use SSL to connect? (true/false)"
+    prompt: "Use SSL to connect? (false/true)"
```
sanderblue, line 48: Why rearrange this? Is it due to how default is `false`? Although minor, this might throw some users off (me being one of them)

recipes/newrelic/infrastructure/ohi/postgres/rhel.yml
```diff
@@ -86,6 +88,29 @@ install:
             echo "The infrastructure agent is required to install this integration, we recommend going through our guided install path for this pre-requisite which can be found here:  https://docs.newrelic.com/docs/new-relic-guided-installation-overview" >> /dev/stderr
             exit 1
           fi
+        - |
+          IS_DATABASE_VALID=$(PGPASSWORD={{.NR_CLI_PASSWORD}} psql -U {{.NR_CLI_USERNAME}} -w {{.NR_CLI_DATABASE}} -h localhost -p {{.NR_CLI_PORT}} -c "select version();" 2>&1 | grep -i PostgreSQL | grep -v grep | wc -l)
+          if [ $IS_DATABASE_VALID -eq 0 ] ; then
+            echo "[Error]: The provided database name is not accessible with the provided username and password and port. See https://docs.newrelic.com/docs/integrations/host-integrations/host-integrations-list/postgresql-monitoring-integration#comp-req for more info." >> /dev/stderr
```
devfreddy, line 94: @Julien4218 Would it be easier to read these messages as they are output if we split the docs link onto another line?

recipes/newrelic/infrastructure/ohi/couchbase/debian.yml
Julien4218: Correct, I re-ordered to match our default (false), so the first value is the default. It seemed backward otherwise.

devfreddy (approved):

recipes/newrelic/infrastructure/ohi/couchbase/debian.yml
devfreddy: I agree, but also, its a small issue. What if we did:

`prompt: "Use SSL to connect? (default: false)"`
or
`prompt: "Use SSL to connect? (true/false, default: false)`

Julien4218: I don't really favor any of those 2 alternatives. I think the user should know what are all the possible choices (discard first suggestion).
The 2nd one is quite verbose, we know users don't like to read text. My concern is they stop to read at the first param and use that one, which has plenty of other implications and would put them on a path to fail because they don't have all the required certificate and pem key files (3 files in total).

recipes/newrelic/infrastructure/ohi/postgres/rhel.yml
Julien4218: We could. I've heard we were thinking of capturing the last line in stderr in our events, so we could know what was the last error. If we do that, we would miss on the root cause error message.

PR #278, opened 2021-03-23
update recipe schema with successLink field

ctrombley:
Updating `successLink` -> `successLinkConfig` and updating docs per UAPI feedback.

cc @michaelkro


jbeveland27 (dismissed):
LGTM

PR #279, opened 2021-03-23
Install two ASP.NET Core test apps on an Ubuntu 20.4 system with nginx reverse proxy

nr-ahemsath:
Note: this PR parallels @jaffinito's PR #270

Adds Ubuntu 20.4 def with ASP.NET Core and Nginx
Adds new deploy-application section for Nginx and Ubuntu 20.4
deploy-application contains 2 apps, one self contained and one framework dependent
Both install systemd services and Nginx configurations
Self contained just runs the executable directly
Framework dependent uses the dotnet commend to run the DLL
Note: The two new apps live in a different S3 than the other apps. They will need to be moved over at some point and the scripting updated.


Julien4218:
@nr-ahemsath I've made a change to the .net/install/ubuntu20 roles for downloading and installing microsoft package.
I've tested local, and the ubuntu 20.04 is still working. I believe now this role can be used on all debians.

Let me know what you think, given the change was isolated in a single file, I didn't create a sub-branch and PR.
If you like it, we can then rename the role to be /debian instead of /ubuntu20

Julien4218:
@nr-ahemsath I've made a couple mistakes on what I pushed yesterday, so I've added this commit today https://github.com/newrelic/open-install-library/pull/279/commits/6e4729f7c5452d0bc75410a8860eecc385c9ec22

I've reverted to use wget instead of the ansible command. Somehow the ansible command was giving me some weird ssl issue when downloading (some odd python dependency perhaps).
I've moved the "apt update" from the onbeforestart step to the configure step. The onbeforestart step is not synchronized for all the services that run on the host, the configure step is. This avoid the deadlock situation inherent to "apt update" when more than 1 service runs on the same host.
I've updated a couple of places that had the `ubuntu` username and use the ansible {{ remote_user }} instead, which provides the same value for ubuntu.
Finally I've added a Debian 10 manual test, and validated both deploy config file deploy fine and have their dotnet sites up and running.

Lastly, I'm seeing an issue when the deploy don't work when re-run. We typically develop deploys so they can be idempotent. I'll be looking into what's going on there and fixing any issue.

Julien4218:
@nr-ahemsath I think I'm done making changes with this last commit https://github.com/newrelic/open-install-library/commit/e125bff1bd660a2dcc04ad6007d921d99e9b5928

We can now re-run deploy without errors. I've also factored some of the implementations, taking advantage of ansible template for re-use between the 2 kind of app we have (self contained versus framework dependent).
I've also added some stop, start, and after start processing.
Feel free to review and let me know if you have questions or feedback.

Julien4218 (approved):

nr-ahemsath:
@Julien4218 Thanks for all your help!  Looks like you'll have to merge the PR; I don't have permissions.

PR #280, opened 2021-03-23
Wrap application.RemoveAll() in if to check for child nodes

jaffinito:
Fixes the following error by checking if the `application` element has any children and skip if it doesn't:
```
Method invocation failed because [System.String] does not contain a method named 'RemoveAll'.
At line:7 char:1
+ $xdoc.configuration.application.RemoveAll();
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [], RuntimeException
    + FullyQualifiedErrorId : MethodNotFound
```

### DIFF
--- a/recipes/newrelic/apm/dotNet/windows/net-install.yml
+++ b/recipes/newrelic/apm/dotNet/windows/net-install.yml
@@ -81,7 +81,9 @@ install:
           $xdoc = New-Object System.Xml.XmlDocument;
           $xdoc.PreserveWhitespace = $true;
           $xdoc.Load($file);
-          $xdoc.configuration.application.RemoveAll();
+          if ($xdoc.GetElementsByTagName("application")[0].HasChildNodes){
+            $xdoc.configuration.application.RemoveAll();
+          }
           if ("{{.NEW_RELIC_REGION}}" -like "STAGING"){
             $xdoc.configuration.service.SetAttribute("host", "staging-collector.newrelic.com")
           }


recipes/newrelic/apm/dotNet/windows/net-install.yml
```diff
@@ -81,7 +81,9 @@ install:
           $xdoc = New-Object System.Xml.XmlDocument;
           $xdoc.PreserveWhitespace = $true;
           $xdoc.Load($file);
-          $xdoc.configuration.application.RemoveAll();
+          if ($xdoc.GetElementsByTagName("application")[0].HasChildNodes){
```
Julien4218, line 84: should we check the call to GetElementsByTagName( ) returns at least 1 element, to avoid having another (index out of range) exception?

jaffinito: Nice catch! Let me fix that.

jaffinito: After looking it over and running some tests, I don't think we need to make any changes.
- The `application` element is a required element so it should never be removed and will prevent the agent from starting if it is.
- I did some testing and `$xdoc.GetElementsByTagName("application")[0].HasChildNodes` does not throw an error if the element does not exist.  It just returns nothing which powershell treats as false.

Julien4218: We can leave it like that, but the moment `application` is no longer a required element (schema change), this will blow up. Perhaps we'd want it to be?

Julien4218 (approved):

recipes/newrelic/apm/dotNet/windows/net-install.yml
jaffinito: It should fail gracefully and not error out, but yeah it will stop doing its job.

Julien4218 (approved):

PR #281, opened 2021-03-24
Replace Get-WmiObject for uninstall

jaffinito:
There are a few reasons to replace Get-WmiObject with the included script.
- Eliminates the extra output we have been seeing when installing the infra agent (below)
```
__GENUS          : 2
__CLASS          : __PARAMETERS
__SUPERCLASS     :
__DYNASTY        : __PARAMETERS
__RELPATH        :
__PROPERTY_COUNT : 1
__DERIVATION     : {}
__SERVER         :
__NAMESPACE      :
__PATH           :
ReturnValue      : 0
PSComputerName   :
```
- Get-WmiObject  is very slow compared to the script
```
Basic test deploy with minimal apps:
Find-UninstallCommand	252.3892 ms
Get-WmiObject		2855.464 ms

Dev laptop with lots of apps:
Find-UninstallCommand	5094.1015 ms
Get-WmiObject		122040.7047 ms
```
- Find-UninstallCommand can be adjusted to change the amount of data is writes out to the console


recipes/newrelic/apm/dotNet/windows/net-install.yml
```diff
@@ -61,7 +61,51 @@ install:
...
+              return ""
+            }
+
+            return "msiexec /x """ +$uninstallIds[0] + """ /qn"
+          }
+
+          $uninstallCommand = Find-UninstallCommand -Match "New Relic .NET Agent"
+          if ($uninstallCommand) {
+            Invoke-Expression $uninstallCommand
+            Start-Sleep -s 15
```
Julien4218, line 106: @jaffinito why are we sleeping 15 seconds?

recipes/newrelic/apm/dotNet/windows/net-install.yml
```diff
@@ -61,7 +61,51 @@ install:
...
+                $uninstallIds.Add($keyId) | Out-Null
+              }
+            }
+
+            if ($uninstallIds.Count -gt 1) {
+              Write-Host "ERROR: More than one product was detected for the Match ""$Match"".  Please narrow the match so that only one product is found."
+              #exit 10022
+            }
+
+            if ($uninstallIds.Count -eq 0) {
```
Julien4218, line 95: Good idea to put a warning, however I think we should remove not to confuse users that something could be wrong (it's a legitimate use case to not have to uninstall)

recipes/newrelic/apm/dotNet/windows/net-install.yml
```diff
@@ -61,7 +61,51 @@ install:
...
+            foreach ($key in $allKeys) {
+              $keyData = Get-Item -Path HKLM:\$key
+              $name = $keyData.GetValue("DisplayName")
+              if ($name -and $name -match $Match) {
+                $keyId = Split-Path $key -Leaf
+                $uninstallIds.Add($keyId) | Out-Null
+              }
+            }
+
+            if ($uninstallIds.Count -gt 1) {
```
Julien4218, line 90: This is a little puzzling. I would like to find a way to proceed with the installation without surfacing an error. Could we return the first one that matched and just un-install that one?

Julien4218 (dismissed):

Julien4218 (approved):

PR #283, opened 2021-03-25
Add post install check to ensure infra agent is started

Julien4218:


jbeveland27 (approved):
Once checks pass, LGTM 🚀

Julien4218:
The test failing on the PR test seems to be a false positive. I'm going to merge this PR and watch for the non-regression test suite run after merging.

Julien4218:
I've run the test mysql-rhel.json locally and it passed.

PR #289, opened 2021-03-30
Initial node agent install recipe

carlo-808:
This is a WIP recipe for installing the Node.js agent on multiple node application on a single linux host.


carlo-808:
Going to resubmit PR from branch rather than fork because slack secret not available from forked PR.

PR #290, opened 2021-03-31
Update case of type for successLinkConfig

gabeobrien:
This PR update the type from explorer to EXPLORER to match expectation from GraphQL.


ctrombley (approved):
:shipit:

PR #292, opened 2021-04-01
Adds recipe for .NET on RHEL-based linux using systemd

jaffinito:
DRAFT

This PR is meant to be a draft PR to see what people think of the approach taken to get the .NET Agent install on a Linux server without any user interaction.  I don't expect this to be merged.

This recipe will install the .NET Agent and configure any systemd services that are directly calling the .NET app (executable or via dotnet ...).

Highlights

- Use the .net introspect to find the .net processes running on the system
- stores data in /tmp/dotnet-introspector/ for use by the recipe
- Uses a systemd drop in file to add the agent env vars to the service for easy cleanup

Below is some information on task...

task: setup_pre_req

- removes the temp directory the install uses to make sure we are using fresh data
- installs unzip and jq since these are used by the install later on

task: introspector

- downloads the .net introspector  from the .NET teams S3
- Uses it to get a list of .NET process which are cleaned up using jq

task: gather_services

- gets a list of all systemd services on the system and saves them to /tmp/dotnet-introspector/all_services.txt
- Using the the introspector data it pulls out the services that are running .NET apps
- The way this script works will require the service to be calling the .net app directly

task: stop

- Stops the .net services

task: remove_any_previous

- Attempts to uninstall the .net agent and remove any left over files
- Checks for our systemd drop in file and removes it if present
- Will also remove the empty drop in directory to allow the service to start.

task: install

- sets up our agent repo and installs the agent

task: configure

- Creates the systemd drop in directories and adds our override file with env vars
- Removes the default app name from the newrelic.config
- Enables DT support in the newrelic.config
- Can detect and configure the staging collector

task: start

- reloads the systemd daemons
- Starts the .net services

task: ensure_sites_started

- uses apachectl and some sed magic to get the ports the virtualhosts are listening on
- Uses curl to ping each site at the port 10 times over 10 seconds to get the agent reporting

task: cleanup_temp_files

- Removes /tmp/dotnet-introspector to keep things clean and tidy

### DIFF
--- a/recipes/newrelic/apm/dotNet/linux/rhel_apache.yml
+++ b/recipes/newrelic/apm/dotNet/linux/rhel_apache.yml
@@ -1,0 +1,178 @@
+# Visit our schema definition for additional information on this file format
+# https://github.com/newrelic/open-install-library/blob/main/docs/recipe-spec/recipe-spec.md#schema-definition
+
+name: dotnet-agent-installer
+displayName: .NET Agent
+description: New Relic install recipe for .NET applications on Linux
+repository: https://github.com/newrelic/newrelic-dotnet-agent
+
+installTargets:
+  - type: application
+    os: linux
+
+keywords:
+  - Apm
+  - .NET
+  - dotnet
+  - aspnet
+  - core
+
+processMatch: []
+
+validationNrql: "SELECT count(*) from Transaction WHERE host like '{{.HOSTNAME}}%' facet entityGuid since 10 minutes ago"
+
+successLinkConfig:
+   type: explorer
+   filter: '"`tags.language` = ''dotnet''"'
+
+install:
+
+  version: "3"
+  silent: true
+
+  tasks:
+    default:
+      cmds:
+        - task: setup_pre_req
+        - task: introspector
+        - task: gather_services
+        - task: stop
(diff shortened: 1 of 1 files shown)


Julien4218:
@jaffinito I really like what you've done here, it's quite simple from a user perspective, I really like it.

A minor adjustments We typically don't install 3rd party dependencies, but rather ensure they are there (pre-req check). Maybe we could prompt the user if we see something is not installed (unzip/jq) and ask the user either to install them and re-run, or potentially confirm we can install those for them.

jaffinito:
@Julien4218 Thank you!  I was concerned about just installing the packages as well.  I like the idea of prompting the user to approve the installation.

One approarch I didn't try since I was limited on time, was to detect if the packages are present and if not install them and then remove them later on in cleanup task.  There could be dependency issues, maybe, with this approach, however.

Julien4218:
@jaffinito Yeah I think it would be nice to detect if any of those dependencies are not there and propose to the user to install them. I also don't think we should un-install those dependencies at the end, but I guess we could have logic to ask and do so if the user likes

jaffinito:
Now that we have the real PR (https://github.com/newrelic/open-install-library/pull/331) we can close this out.

PR #293, opened 2021-04-01
fix(docker-logs): build discovered logs config file using a bash array to fix "-" executable error

sanderblue:
The underlying `sh` library used by `go-task` has some subtle differences from a native shell. Glob paths such as `/var/lib/docker/containers/*/*.log` will be expanded to the full file path such as `/var/lib/docker/containers/9483459940/9483459940.log` when trying to use with `echo`. When we try to build the discovered.yml file, we need the glob path and using a bash array avoids using `echo` and keeps the glob path intact.

This will require an update in the CLI as well since we also can't inject a multiple string into the `discoveredLogFiles` bash variable because it causes the `sh` interpreter to silently error.

Depends on: [newrelic/newrelic-cli #792](https://github.com/newrelic/newrelic-cli/pull/792)

### DIFF
--- a/recipes/newrelic/infrastructure/logs/logs.yml
+++ b/recipes/newrelic/infrastructure/logs/logs.yml
@@ -79,8 +79,8 @@ install:
           fi
         - |
           # First check on empty inputs
-          logFiles=$(echo -n {{.LOG_FILES}})
-          discoveredLogFiles=$(echo -n {{.DISCOVERED_LOG_FILES}})
+          logFiles=$(echo -n "{{.LOG_FILES}}")
+          discoveredLogFiles=$(echo -n "{{.DISCOVERED_LOG_FILES}}")
           logsStringLength=${#logFiles}
           discoveredLogsStringLength=${#discoveredLogFiles}
           if [ $logsStringLength -eq 0 ] && { [ $discoveredLogsStringLength -eq 0 ] || [ "$discoveredLogFiles" = "logs: []" ]; }; then
@@ -105,10 +105,8 @@ install:
           echo -e "verbose: 3\nlog_format: json" | sudo tee -a /etc/newrelic-infra.yml > /dev/null
           sudo sed -i '/^$/d' /etc/newrelic-infra.yml
         - |
-          logFiles=$(echo -n {{.LOG_FILES}})
+          logFiles=$(echo -n "{{.LOG_FILES}}")
           logsStringLength=${#logFiles}
-          discoveredLogFiles=$(echo -n {{.DISCOVERED_LOG_FILES}})
-          discoveredLogsStringLength=${#discoveredLogFiles}
 
           # Keep a count of the valid log files the integration will use.
           foundLogFilesCount=0
@@ -126,6 +124,9 @@ install:
             done
           fi
 
+          discoveredLogFiles=$(echo -n "{{.DISCOVERED_LOG_FILES | trim}}")
+          discoveredLogsStringLength=${#discoveredLogFiles}
+
           # If the user provides log files that don't exist and we haven't discovered any log files,
           # we cannot proceed with the installation.
           if [ $foundLogFilesCount -eq 0 ] && { [ $discoveredLogsStringLength -eq 0 ] || [ "$discoveredLogFiles" = "logs: []" ]; }; then
@@ -133,8 +134,18 @@ install:
             exit 16
           fi
 
+          # Build the discovered.yml logs config file used for Docker logs
+          IFS=', ' read -r discoveredLogFilesArray <<< "$discoveredLogFiles"
+
           if [ $discoveredLogsStringLength -gt 0 ]; then
-            echo -e "{{.DISCOVERED_LOG_FILES}}" | sudo tee /etc/newrelic-infra/logging.d/discovered.yml > /dev/null
+            echo "logs:" | sudo tee /etc/newrelic-infra/logging.d/discovered.yml > /dev/null
+
+            for filePath in "${discoveredLogFilesArray[@]}"
+            do
+              if [ -f $filePath ]; then
+                echo -e "  - name: docker log\n    file: $filePath" | sudo tee -a /etc/newrelic-infra/logging.d/discovered.yml > /dev/null
+              fi
+            done
           fi
 
     restart:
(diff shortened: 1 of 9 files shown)


recipes/newrelic/infrastructure/logs/logs.yml
```diff
@@ -126,15 +124,28 @@ install:
             done
           fi
 
+          discoveredLogFiles=$(echo -n "{{.DISCOVERED_LOG_FILES | trim}}")
```
zlesnr, line 127: Where does trim come from?

recipes/newrelic/infrastructure/logs/logs.yml
```diff
@@ -126,15 +124,28 @@ install:
...
+
           # If the user provides log files that don't exist and we haven't discovered any log files,
           # we cannot proceed with the installation.
           if [ $foundLogFilesCount -eq 0 ] && { [ $discoveredLogsStringLength -eq 0 ] || [ "$discoveredLogFiles" = "logs: []" ]; }; then
             echo "ERROR Provided log files do not exist: $logFiles. No log file(s) found or discovered. Cannot proceed. Exiting installation." >> /dev/stderr
             exit 16
           fi
 
+          # Build the discovered.yml logs config file used for Docker logs
+          IFS=', ' read -r discoveredLogFilesArray <<< "$discoveredLogFiles"
```
zlesnr, line 138: Is modifying the IFS here necessary?

recipes/newrelic/infrastructure/logs/logs.yml
Julien4218: Also, why trim? I've seen error in the past with the pipe causing error in the case the variable is not defined or empty.
I would prefer to remove the trim.

recipes/newrelic/infrastructure/logs/logs.yml
Julien4218: Why do we need this? We should just need to write the config file?

recipes/newrelic/infrastructure/logs/logs.yml
```diff
@@ -126,15 +124,28 @@ install:
...
+
           if [ $discoveredLogsStringLength -gt 0 ]; then
-            echo -e "{{.DISCOVERED_LOG_FILES}}" | sudo tee /etc/newrelic-infra/logging.d/discovered.yml > /dev/null
+            echo "logs:" | sudo tee /etc/newrelic-infra/logging.d/discovered.yml > /dev/null
+
+            for filePath in "${discoveredLogFilesArray[@]}"
+            do
+              if [ -f $filePath ]; then
```
Julien4218, line 145: I dont think we should validate filepath here.
We just need to write the config, the path may exist later and the agent will pick it up, that's the main case actually for docker logs which are generated in a guid sub-folder

Julien4218 (changes requested):
There is also a debian-logs.yml recipe to update

recipes/newrelic/infrastructure/logs/logs.yml
sanderblue: Makes sense. I'll update.

recipes/newrelic/infrastructure/logs/logs.yml
sanderblue: In this case, probably not since it's only one path, but what if we have a string of comma-separated file paths? I guess modifying `IFS` for this can have side effects, so is there a better way to handle this? I'm not exactly a shell expert 😉

recipes/newrelic/infrastructure/logs/logs.yml
sanderblue: > All functions by the Go’s slim-sprig lib are available.

Reference: [`go-task` docs](https://taskfile.dev/#/usage?id=gos-template-engine) 🦙

Julien4218: Ho it's Go, then maybe it's fine, I think the pipe in the bash was causing error before, but probably not in go, I assume it handles undefined variable, or does it not?

zlesnr: Ohh, good call.

recipes/newrelic/infrastructure/logs/logs.yml
zlesnr: You might check the block at line 120 that we use for the other log file pattern.  It looks something like this:
```bash
for i in $(echo -n {{.LOG_FILES}} | sed "s/,/ /g"); do
done
```
So if there was a comma, it is now a space, and so the list would look like `/file1 /file2 /file3` which extrapolates to this: `for x in /file1 /file2 /file3; do...` which I think covers the bases.  It isn't very clean but it should work.  In that case, it doesn't become an array, but you are just passing a series of arguments to the for loop, which can be easier to work with (and read) than an array.  Just a thought though.

sanderblue: The description in the PR explains the oddity around this with the `sh` library that go-task uses.

zlesnr: Yeah, I wasn't thinking about expansion.

zlesnr (dismissed):

recipes/newrelic/infrastructure/logs/logs.yml
Julien4218: I'm not sure what this does, and therefore don't see why we need it. My incline would be to remove, particularly knowing that each entry may be a pattern and not an existing file path

sanderblue: If we have a csv list of paths (regardless if it contains patterns or full paths), we need to tell bash to use the comma as a separator when reading the string, otherwise the list must be space-delimited which introduces inconsistencies in how set these paths in the recipes. Another issue is that I we can't use the `-d` (delmiter) option with `read` because it's not supported by the `sh` library. Otherwise I could use that and it would read something similar to `read -d "," myArr <<< "$discoveredLogFiles`, but unfortunately it's not possible with the `sh` lib.

zlesnr (approved):

Julien4218 (approved):
This is looking good. Approving and starting testing (automated+staging)

PR #294, opened 2021-04-05
Disable logging for unsupported ubuntu versions

Julien4218:


jbeveland27 (commented):
Nice catch @Julien4218. Is it worth adding an ubuntu 14 test definition to run infra install + this `ubuntu-unsupported` recipe together?

Julien4218:
> Nice catch @Julien4218. Is it worth adding an ubuntu 14 test definition to run infra install + this `ubuntu-unsupported` recipe together?

Yeah, we already have a test definition for ubuntu 14, and it does pull the log recipe too since we install as guided install now for the infra tests.

zlesnr (dismissed):

jbeveland27:
Ah right, I forgot we switched that to guided install

jbeveland27 (approved):

Julien4218:
Related and dependent PR on the library service to be deployed before this change 
https://source.datanerd.us/open-source/open-install-library-service/pull/34

Julien4218:
@jbeveland27 I've started to test the service change on staging, and notice the centos/rhel recipes needed their platformVersion to be removed otherwise those recipes would now get filtered out by the service.
So I've removed those.

I have the service deployed on Staging, and at this point I feel confident merging this PR and deployed it to Staging to finish the manual non-regression testing and the ubuntu 14 log testing as well.

Let me know if you have any objections

PR #306, opened 2021-04-15
Updates .NET recipe to use 2 exit code to match docs; add error code …

gabeobrien:
Updates .NET recipe to use 2 exit code to match docs
Add error code when no APM processes found to instrument
Updates java recipe to exit with the correct code


recipes/newrelic/apm/dotNet/windows/net-install.yml
```diff
@@ -49,7 +49,7 @@ install:
           $iisEnabled = Get-WindowsOptionalFeature -Online -FeatureName IIS-WebServer | select State
           if($iisEnabled.State -ne "Enabled"){
             Write-Error "IIS is not installed."
-            exit 1
+            exit 2
```
Julien4218, line 52: @gabeobrien actually, we may have to update the doc, we want a different error code so we can distinguish between "IIS is not installed" and "ASP.NET is not installed". If we do this change we wouldn't know the difference
cc @jaffinito

Julien4218 (changes requested):
We should use a different error code between the IIS is not installed and ASP.NET is not installed so we can know which one is causing to error.

recipes/newrelic/apm/dotNet/windows/net-install.yml
gabeobrien: Agreed.  Updated!

Julien4218 (dismissed):
This looks good, assuming all the end-to-end tests are passing.
We'll want to review our Tessen dashboard to update some of the graph query with the new error code values.
Let me know when this is released, I can make those changes.

Julien4218 (approved):

PR #307, opened 2021-04-15
Run apm iis test as guided with default, and ping site upon install t…

Julien4218:
…o make sure w3wp is started


gabeobrien (approved):
+1

PR #310, opened 2021-04-16
fix(Ubuntu Infra-agent): Update Ubuntu to match Debian recipe to reconcile differences in apt

devfreddy:


recipes/newrelic/infrastructure/ubuntu.yml
```diff
@@ -136,14 +136,24 @@ install:
...
     install_infra:
       cmds:
-        - sudo curl -s https://download.newrelic.com/infrastructure_agent/gpg/newrelic-infra.gpg | sudo apt-key add -
-        - sudo rm -f /etc/apt/sources.list.d/newrelic-infra.list
-        - printf "deb [arch=amd64] https://download.newrelic.com/infrastructure_agent/linux/apt {{.DEBIAN_CODENAME}} main" | sudo tee -a /etc/apt/sources.list.d/newrelic-infra.list > /dev/null
-        - sudo apt-get update -yq
-        - sudo apt-get install newrelic-infra -y -qq
+        - |
+          sudo apt-get update -yq
+          sudo apt-get install gnupg2 -y
```
Julien4218, line 141: This is probably the culprit, but worth asking, do we need to install this?
I'm shy on installing things that are not newrelic on the user host.
We should perhaps, check that it's not installed before installing, and maybe confirm with the user to ask them if we can install this required dependency?

Julien4218: nevermind, it looks like we do this for debian too

Julien4218 (approved):

jbeveland27 (approved):

PR #313, opened 2021-04-20
feat(schema): add support for stability, dependencies, and quickstarts

jbeveland27:
`stability` is the current term that was chosen to signify which level of recipes we want to expose in the CLI via a flag like `--stability=stable`. Other options I thought of included `library` (`--library=stable`), `category` (`--category=stable`), and `channel` (`--channel=stable`). 

Does stability make the most sense?

### DIFF
--- a/docs/recipe-spec/recipe-spec.md
+++ b/docs/recipe-spec/recipe-spec.md
@@ -34,6 +34,35 @@ description: string, required
 # Example: https://github.com/newrelic/infrastructure-agent
 repository: string, required
 
+# Indicates stability level of recipe. Useful for gradually enabling the
+# availablity of a recipe to users as it is developed and tested. Can be
+# thought of as an indicator for how stable/ready a recipe is for general 
+# consumption and can be used to quickly disable problematic recipes.
+# 
+# Usage: # newrelic install --stability=stable
+# 
+# Levels:
+#   disabled - No availability (recipe will not return from NerdGraph)
+#   experimental - Non-backward compatible changes or removal may occur in any future release. Use of the feature is not recommended in production environments.
+#   stable - General availability - ongoing support and compatability is a high priority
+stability: enum, optional # One of [ disabled, experimental, stable ]
+
+# Dependency list for recipes (by name) that must be run successfully prior to attempting
+# the current recipe
+# ex: 
+# dependencies:
+#   - infrastructure-agent-installer
+dependencies: list, optional
+
+# Metadata used to recommend and install Quickstarts (dashboards, alerts, synthetics, etc.)
+# This is filtering criteria for the quickstartSearch endpoint in NerdGraph
+quickstarts: list (object), optional
+  - name: string, required
+    entities: object, optional
+      type: string, required
+      domain: string, required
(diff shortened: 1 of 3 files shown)


sanderblue (commented):
I like the idea behind these additions. Will give others a chance to chime in as well.

docs/recipe-spec/recipe-spec.md
```diff
@@ -34,6 +34,35 @@ description: string, required
...
+# thought of as an indicator for how stable/ready a recipe is for general 
+# consumption and can be used to quickly disable problematic recipes.
+# 
+# Usage: # newrelic install --stability=stable
+# 
+# Levels:
+#   disabled - No availability (recipe will not return from NerdGraph)
+#   experimental - Non-backward compatible changes or removal may occur in any future release. Use of the feature is not recommended in production environments.
+#   stable - General availability - ongoing support and compatability is a high priority
+stability: enum, optional # One of [ disabled, experimental, stable ]
```
Julien4218, line 48: I'm not onboard with this attribute, I don't want recipe out with less clarity on their reliability, if a customer tries out something that's disabled or experimental that is still something we'll have to support.
We don't need this for making Infra optional.

docs/recipe-spec/recipe-spec.md
```diff
@@ -34,6 +34,35 @@ description: string, required
...
+# Dependency list for recipes (by name) that must be run successfully prior to attempting
+# the current recipe
+# ex: 
+# dependencies:
+#   - infrastructure-agent-installer
+dependencies: list, optional
+
+# Metadata used to recommend and install Quickstarts (dashboards, alerts, synthetics, etc.)
+# This is filtering criteria for the quickstartSearch endpoint in NerdGraph
+quickstarts: list (object), optional
```
Julien4218, line 59: This is not required for the Infra MMF, please remove.

Julien4218 (changes requested):
Please only add the dependencies element for the Infra work and move the rest to a separate PR

docs/recipe-spec/recipe-spec.md
jbeveland27: @Julien4218 This doesn't have to do with the Infra MMF. It's for [Virutoso-616](https://newrelic.atlassian.net/browse/VIRTUOSO-616), and it's tangentially related to the Quickstarts work so that we have a way to start adding that into experimental recipes.

We've agreed previously that we should have some mechanism for gating recipes. I don't think this does anything w.r.t. clarity around reliability, it's just a way to limit what recipes are used in a given context via a CLI flag. It's mostly a way for us to experiment without making a recipe generally available.

docs/recipe-spec/recipe-spec.md
jbeveland27: Not related to the infra MMF. This is for [Virtuoso-649](https://newrelic.atlassian.net/browse/VIRTUOSO-649).

jbeveland27:
New PRs:

Dependencies:
https://github.com/newrelic/open-install-library/pull/314

Quickstarts:
https://github.com/newrelic/open-install-library/pull/315

Stability:
https://github.com/newrelic/open-install-library/pull/316

PR #314, opened 2021-04-20
feat: add support for dependencies

jbeveland27:
related: https://newrelic.atlassian.net/browse/VIRTUOSO-684


devfreddy (approved):

sanderblue (approved):

Julien4218 (approved):
@jbeveland27 we should have a separate task/PR to add a check when testing to make sure a dependency name exist in the library so we'll know before releasing that a recipe is not usable.

PR #315, opened 2021-04-20
feat: add support for quickstarts

jbeveland27:
Related: https://newrelic.atlassian.net/browse/VIRTUOSO-649


sanderblue (approved):

devfreddy (approved):

Julien4218 (commented):
I would have liked a meeting to discuss why this addition is necessary. There are a lot of new words/concepts here and at least to me it's not clear what those means and how they would be used in an installation context.

PR #316, opened 2021-04-20
feat: add support for stability

jbeveland27:
Related: https://newrelic.atlassian.net/browse/VIRTUOSO-616


sanderblue (approved):

Julien4218 (changes requested):
I'd like a meeting so we can discuss this need and the best way to implement it.

PR #318, opened 2021-04-20
Adds jq error code

jaffinito:
jq is going to be used in the .NET Agent install recipe for Linux


gabeobrien:
For the Java recipe I pulled JQ out and replaced it with 'sed'.  Not ideal but we opted to not deal with installing packages on the customer machine if we can help it.

Julien4218:
> For the Java recipe I pulled JQ out and replaced it with 'sed'. Not ideal but we opted to not deal with installing packages on the customer machine if we can help it.

Yeah if we can workaround with `sed` that would be a lot better than installing jq, we already have a pre check assertion that makes sure sed is installed, which it is usually

gabeobrien:
Here are the two ways that I pull data out of the JSON output from the Java CLI:
https://github.com/newrelic/open-install-library/blob/main/recipes/newrelic/apm/java/rhel.yml#L60
https://github.com/newrelic/open-install-library/blob/main/recipes/newrelic/apm/java/rhel.yml#L64
Happy to help with this as needed.

jaffinito:
Alex found a way to break things up in our current PR.  Going to close this out since we appear to be able to drop jq.

PR #319, opened 2021-04-20
APM install, .NET/Windows: check for Administrator permissions before attempting anything else

nr-ahemsath:
This fixes https://newrelic.atlassian.net/browse/VIRTUOSO-651 by:
1. Adding a new step at the beginning to make sure the recipe is being run from a shell with Administrator permissions and exit with an informative message and error code (7) if not
2. Update the error codes doc with the new error code and meaning.

I did also tweak the language a bit on some of the error code descriptions in the document to match how various .NET ecosystem things tend to be capitalized, I can revert if necessary.


nr-ahemsath:
Converted to draft since we have another story to make the IIS and ASP.NET checks also be warnings instead of errors and I can just do that in this PR.

Julien4218 (approved):
This looks good.
cc @gabeobrien

Julien4218 (approved):

PR #322, opened 2021-04-23
Improves the ability create initial transactions and improve install validation success 

jaffinito:
This does two things:

Adds inputVars and checks to go along with them. Input now asks for comma-seperated URLs.  There is no error if they just hit enter and move on.  ensure_sites_started will now check if URLs have been provided and use those instead of the IIS bindings.  Setup so that any errors are ignored and there should be no output.

Adds a new task that asks the user to exercise their web apps.  This task occurs right before ensure_sites_started.  It displays a rather larger "marquee" in yellow text.  The script will not move forward until the user presses enter specifically.  The URLs will still be hit automatically in the ensure_sites_started task.

Please feel free to improve the text for the input prompt and marquee!

Resolves  newrelic/newrelic-dotnet-agent#529


recipes/newrelic/apm/dotNet/windows/net-install.yml
```diff
@@ -20,12 +20,18 @@ keywords:
 processMatch:
   - w3wp.exe
 
-validationNrql: "SELECT count(*) from Transaction WHERE host like '{{.HOSTNAME}}%' facet entityGuid since 10 minutes ago"
+validationNrql: "SELECT count(*) from Transaction WHERE host like '{{ .HOSTNAME }}%' facet entityGuid since 10 minutes ago"
```
Julien4218, line 23: @jaffinito I dont think there should be spaces between the {{ and .VARNAME and .VARNAME }}

recipes/newrelic/apm/dotNet/windows/net-install.yml
```diff
@@ -20,12 +20,18 @@ keywords:
...
 
 successLinkConfig:
   type: EXPLORER
   filter: '"`tags.language` = ''dotnet''"'
 
+inputVars:
+  - name: "IIS_ROOT_URLS"
```
Julien4218, line 30: @jaffinito could you rename this input var with a NR_CLI_ prefix, this is a consistent prefix we use across all recipes

recipes/newrelic/apm/dotNet/windows/net-install.yml
```diff
@@ -129,7 +136,7 @@ install:
           powershell -command '
           [Net.ServicePointManager]::SecurityProtocol = "tls12, tls";
           (New-Object System.Net.WebClient).DownloadFile("https://download.newrelic.com/dot_net_agent/latest_release/NewRelicDotNetAgent_x64.msi", "$env:TEMP\NewRelicDotNetAgent_x64.msi");
-          msiexec.exe /qn /i "$env:TEMP\NewRelicDotNetAgent_x64.msi" NR_LICENSE_KEY={{.NEW_RELIC_LICENSE_KEY}} | Out-Null;'
+          msiexec.exe /qn /i "$env:TEMP\NewRelicDotNetAgent_x64.msi" NR_LICENSE_KEY={{ .NEW_RELIC_LICENSE_KEY }} | Out-Null;'
```
Julien4218, line 139: @jaffinito same comment with spaces for go-task template var

recipes/newrelic/apm/dotNet/windows/net-install.yml
```diff
@@ -141,10 +148,10 @@ install:
           $xdoc = New-Object System.Xml.XmlDocument;
           $xdoc.PreserveWhitespace = $true;
           $xdoc.Load($file);
-          if ($xdoc.GetElementsByTagName("application")[0].HasChildNodes){
+          if ($xdoc.GetElementsByTagName("application")[0].HasChildNodes) {
             $xdoc.configuration.application.RemoveAll();
           }
-          if ("{{.NEW_RELIC_REGION}}" -like "STAGING"){
+          if ("{{ .NEW_RELIC_REGION }}" -like "STAGING") {
```
Julien4218, line 154: no spaces

Julien4218 (commented):
@jaffinito I like the strategy to improve the handling of starting the client application to get NR traffic. 
I wonder if our question is clear enough though. When we prompt for URLs, I could see a user inputting a DNS or load balanced URL that would not hit this server.
Maybe there is a way to try multiple things, silencing any errors, something like this:
1) Start by asking the user to generate some traffic manually onto this server, just prompt for Y to continue
2) Silently iterate and attempt a Get-WebURL on all IIS default site as we were previously doing. Possibly only ask for 1) if we can detect somehow 2) didn't start any ASP.NET app (look for any w3wp.exe process?)

recipes/newrelic/apm/dotNet/windows/net-install.yml
jaffinito: Odd, I have always used spaces between the braces (for readability) and not seen an issue.

recipes/newrelic/apm/dotNet/windows/net-install.yml
jaffinito: Will do!

jaffinito:
We could do both (hitting provide URLs and the IIS provided ones) pretty easily.

As for 2.2 (detect app started), this is not the problem.  The IIS provided URLs will start the app even if there is no page at the specific URL. The problem is we need transactions in order to validate the install.   For our simple test apps, they are created from the root just fine.  For many (I would say most) production apps you need to hit specific urls to start a transaction.

Julien4218:
> We could do both (hitting provide URLs and the IIS provided ones) pretty easily.
> 
> As for 2.2 (detect app started), this is not the problem. The IIS provided URLs will start the app even if there is no page at the specific URL. The problem is we need transactions in order to validate the install. For our simple test apps, they are created from the root just fine. For many (I would say most) production apps you need to hit specific urls to start a transaction.

👍

Julien4218 (dismissed):
This looks good I'm approving. I think before merging we should have a demo with the product managers to give the final approval.
Probably good to also have the designers (Daniel Golden).
cc @gabeobrien

recipes/newrelic/apm/dotNet/windows/net-install.yml
```diff
@@ -158,7 +165,43 @@ install:
...
+          Write-Host -ForegroundColor Yellow "     =                                  Important                                   ="
+          Write-Host -ForegroundColor Yellow "     =                                                                              ="
+          Write-Host -ForegroundColor Yellow "     =   To validate the installation, the application(s) need to be sent traffic.  ="
+          Write-Host -ForegroundColor Yellow "     =                                                                              ="
+          Write-Host -ForegroundColor Yellow "     =              Please send a few requests to your application(s).              ="
+          Write-Host -ForegroundColor Yellow "     =                  Press ENTER key when ready to continue...                   ="
+          Write-Host -ForegroundColor Yellow "     =                                                                              ="
+          Write-Host -ForegroundColor Yellow "     ================================================================================"
+          Write-Host ""
+          '
```
jbeveland27, line 184: This is neat!

recipes/newrelic/apm/dotNet/windows/net-install.yml
```diff
@@ -47,12 +43,49 @@ install:
...
+          Write-Host -ForegroundColor Yellow "=                                   Warning                                    ="
+          Write-Host -ForegroundColor Yellow "=                                                                              ="
+          Write-Host -ForegroundColor Yellow "=       This installation will automatically restart IIS upon completion.      ="
+          Write-Host -ForegroundColor Yellow "=                                                                              ="
+          Write-Host -ForegroundColor Yellow "================================================================================"
+          Write-Host ""
+          Write-Host "If you are hosting your .NET application differently then check out our other installation options:"
+          Write-Host "https://docs.newrelic.com/docs/agents/net-agent/installation/install-net-agent-windows/."
+          Write-Host ""
+          if (! $env:NEW_RELIC_ASSUME_YES) {
```
Julien4218, line 62: I would have expected we use go-task var syntax here like this:
if (! {{.NEW_RELIC_ASSUME_YES}} ) {

recipes/newrelic/apm/dotNet/windows/net-install.yml
```diff
@@ -47,12 +43,49 @@ install:
...
+            WHILE ($true) {
+              Write-Host "Do you want to install the .NET Agent Y/N (defaults to Y)? " -NoNewLine
```
Julien4218, line 64: is it just me or should we update defaults => default ?

recipes/newrelic/apm/dotNet/windows/net-install.yml
```diff
@@ -47,12 +43,49 @@ install:
...
+
+              if ($NEW_RELIC_CONTINUE -ieq "n") {
+                Write-Host "Exiting the installation..."
+                exit 130 # often used to represent Ctrl+C.
+              }
+              elseif ($NEW_RELIC_CONTINUE -ieq "y") {
+                break
+              }
+
+              Write-Host "Please type y or n only."
```
Julien4218, line 79: ho wow, fancy retry! I love it

recipes/newrelic/apm/dotNet/windows/net-install.yml
```diff
@@ -156,9 +192,45 @@ install:
...
+          Write-Host -ForegroundColor Yellow "=    Your application needs to be sent traffic to validate the installation.   ="
+          Write-Host -ForegroundColor Yellow "=                                                                              ="
+          Write-Host -ForegroundColor Yellow "=         Please send approximately 3 requests to your application or          ="
+          Write-Host -ForegroundColor Yellow "=       enter the application URL and this tool will send the requests.        ="
+          Write-Host -ForegroundColor Yellow "=                                                                              ="
+          Write-Host -ForegroundColor Yellow "================================================================================"
+          Write-Host ""
+
+          # If NEW_RELIC_ASSUME_YES is set to true, will not prompt for input
+          if (! $env:NEW_RELIC_ASSUME_YES) {
```
Julien4218, line 219: {{.NEW_RELIC_ASSUME_YES}}

recipes/newrelic/apm/dotNet/windows/net-install.yml
jaffinito: I was assuming it was an env var. I will update things to capture the go-task value and handle it.

Julien4218 (approved):
This looks great. Thank You

PR #323, opened 2021-04-24
Create an install definition for Ubuntu 20.4 with Apache as reverse proxy to ASP.NET Core apps

nr-ahemsath:
This PR should resolve this story in the .NET Agent repo: https://github.com/newrelic/newrelic-dotnet-agent/issues/540

This creates an install definition for Ubuntu 20 with Apache as a reverse proxy to the same pair of ASP.NET Core web apps that we install in the other definitions.

Some other changes:
1. Renamed the existing definitions to account for a new naming scheme that contains the proxy server type.
2. Renamed some paths from `.net` to `dotNet`: the install definitions already expected this, and it's a good idea to avoid hidden directories.


Julien4218 (approved):
This looks good to me.
Looping @gabeobrien FYI

gabeobrien (approved):
+1 this looks good.

nr-ahemsath:
Thanks for the reviews; I'm ready to have this merged to main (I don't have the permissions to do so myself).

Julien4218:
@nr-ahemsath I've merged and closed the branch.
If you could, keep an eye on this non-regression build to make sure it passes, and does so within 90min
https://github.com/newrelic/open-install-library/actions/runs/789823595

PR #327, opened 2021-04-29
Normalize the env variables for username and password

gabeobrien:


Julien4218 (approved):

test/deploy/linux/couchbase/install/debian/roles/configure/tasks/main.yml
```diff
@@ -45,9 +45,9 @@
   when: initialize_cluster|bool
 
 - block:
-  - name: Export NR_CLI_USERNAME
+  - name: Export NR_CLI_DB_USERNAME
     shell: "echo export NR_CLI_API_USERNAME=newrelic >> ~/.bashrc"
-  - name: Export NR_CLI_PASSWORD
+  - name: Export NR_CLI_DB_USERNAME
```
jbeveland27, line 50: This looks like the right change, but is only specific to debian. The Couchbase recipes and the [ansible play for rhel](https://github.com/newrelic/open-install-library/blob/main/test/deploy/linux/couchbase/install/rhel/roles/configure/tasks/main.yml#L34-L37) should be updated to reflect this new name too.

Julien4218: Good catch. I've updated those too.

PR #331, opened 2021-04-30
Add install recipe for .NET Agent on Ubuntu with systemd and Nginx

nr-ahemsath:
This should satisfy the requirements of this story in the agent repo: https://github.com/newrelic/newrelic-dotnet-agent/issues/542

Note: the recipe is currently called `debian-systemd-nginx`, but:
1. There no `nginx` specific logic at this point, and we think this recipe will also work for the `apache` case as-is.  Once this is approved and merged we will test that case and rename the recipe to be more generic if it works for apache as expected.
2. At this point the only debian-specific logic is in the step which adds our `apt` package source to the system and installs the agent from there.  If we could switch this logic based on some way of knowing whether we're on a deb-based vs. an rpm-based distro we should be able to have one recipe file for both deb and rpm distros.


recipes/newrelic/apm/dotNet/debian-systemd-nginx.yml
```diff
@@ -0,0 +1,283 @@
...
+    platform: "ubuntu"
+
+keywords:
+  - Apm
+  - .NET
+  - dotnet
+  - aspnet
+  - core
+
+processMatch: []
```
Julien4218, line 24: @gabeobrien @nr-ahemsath @jaffinito 
There is a potential issue here. Without a processMatch this recipe is going to run for every ubuntu/debian linux, which may have nothing to do with .net core.
Is there a process we could latch on to target this recipe when it will be more proper to run?

recipes/newrelic/apm/dotNet/debian-systemd-nginx.yml
```diff
@@ -0,0 +1,283 @@
...
+          fi
+
+    assert_pre_req:
+      cmds:
+        - |
+          IS_UNZIP_INSTALLED=$(which unzip | wc -l)
+          if [ $IS_UNZIP_INSTALLED -eq 0 ]
+          then
+            echo "unzip is required to run the newrelic install. Please install unzip and re-run the installation." >> /dev/stderr
+            exit 17
```
Julien4218, line 99: Can we document this specific error code in the docs/error-codes.md

recipes/newrelic/apm/dotNet/debian-systemd-nginx.yml
```diff
@@ -0,0 +1,283 @@
...
+        - |
+          processCount=$(cat /tmp/dotnet-introspector/dotnet_processes.txt | wc -w)
+          if [[ "$processCount" -eq 0 ]]; then
+            YELLOW='\033[0;33m'
+            NOCOLOR='\033[0m'
+            echo -e "${YELLOW}
+            No dotnet processes were found.  Please ensure the services you want to instrument
+            are running and try running this installer again.
+            ${NOCOLOR}\n"
+            exit 3
```
Julien4218, line 120: I think this relates to the processMatch topic, we might need to do more work before considering merging this PR (broader detection mechanism in Virtuoso than just process detection).
cc @gabeobrien

recipes/newrelic/apm/dotNet/debian-systemd-nginx.yml
```diff
@@ -0,0 +1,283 @@
...
+                  fi
+              done
+          done
+        - echo "Done gathering services"
+
+    stop:
+      cmds:
+        - |
+          for sn in `cat /tmp/dotnet-introspector/dotnet_services.txt`; do
+            sudo systemctl stop $sn
```
Julien4218, line 145: We might need to handle other system command than just systemctl?
Here is an example https://github.com/newrelic/open-install-library/blob/main/recipes/newrelic/infrastructure/debian.yml#L185-L201

recipes/newrelic/apm/dotNet/debian-systemd-nginx.yml
```diff
@@ -0,0 +1,283 @@
...
+              echo Detected running in STAGING, configuring agent to report to staging
+              echo "Environment=\"NEW_RELIC_HOST=staging-collector.newrelic.com\"" |sudo tee -a "/etc/systemd/system/${sn}.d/nr_dotnet_agent_env.conf" >/dev/null
+            fi
+
+          done
+        - echo "New Relic .NET agent configured"
+
+    start:
+      cmds:
+        - sudo systemctl daemon-reload
```
Julien4218, line 203: Same comment for systemctl

Julien4218 (changes requested):
I'm just putting a hold on this PR to make sure we address the topic of process detection. We wouldn't want to merge this work, have it deployed without some ways to control it doesn't run on all ubuntu/debian linuxes.
cc @gabeobrien

recipes/newrelic/apm/dotNet/debian-systemd-nginx.yml
gabeobrien: I took a look at a few other recipes and it does seem that the '[]' process match will match in all cases where the host matches.  
I will have to deploy a .Netcore app and see what 'ps' shows to know for sure what value to use for processMatch.

recipes/newrelic/apm/dotNet/debian-systemd-nginx.yml
gabeobrien: I think the scope for the current work with .NET is only to support systemctl.  We can expand the support in the future.  I do think we would ensure that systemctl is on the host before getting to far into the recipe.  We did this in the Node recipes:
https://github.com/newrelic/open-install-library/blob/main/recipes/newrelic/apm/node/linux.yml#L49

Julien4218: Assuming we can target the recipe to linuxes we know have a .net core app, this might not be too much of an issue.
Yes, we would have to add some precheck assertions to throw a well known error code if we don't see systemctl is available and running.

recipes/newrelic/apm/dotNet/debian-systemd-nginx.yml
nr-ahemsath: Sure.  However, I'm first exploring the possibility of having our LSI made available in `gzip` format so it can be uncompressed using the gzip tool that is more generally available by default on Linux distros.

This is more a question for the owners of this repo, but it seems like if we create a specific error code for every potential tool needed by an install recipe we might run out of unique numbers in the 1-255 range.

recipes/newrelic/apm/dotNet/debian-systemd-nginx.yml
nr-ahemsath: I've added a step to make sure `systemctl` exists to our `assert_pre_reqs` step for this iteration of the recipe.

recipes/newrelic/apm/dotNet/debian-systemd-nginx.yml
nr-ahemsath: Same reply as the one above. 😄

recipes/newrelic/apm/dotNet/debian-systemd-nginx.yml
jaffinito: We could use `dotnet`, but that only covers cases where customers use `dotnet someawesome.dll` to start the app. If they start it directly from an executable, we won't pick it up.

recipes/newrelic/apm/dotNet/debian-systemd-nginx.yml
jaffinito: I think we are going to take a look at this requirement and try to work around it.

recipes/newrelic/apm/dotNet/debian-systemd-nginx.yml
jaffinito: The introspector can detect .NET apps no matter how they are started.

recipes/newrelic/apm/dotNet/debian-systemd-nginx.yml
jaffinito: True, but I think that would be future work or maybe even another recipe.

recipes/newrelic/apm/dotNet/debian-systemd-nginx.yml
Julien4218: I was thinking the same, and I think like you, think we shouldn't do this (targeting dotnet process).

recipes/newrelic/apm/dotNet/debian-systemd-nginx.yml
Julien4218: Good to hear, I was wondering about that.
I could see we would try to leverage the LSI to get that info, assuming we have the capability in Virtuoso to have more abstract detection, does this mean we would install the LSI for every user attempting to install on ubuntu/linux? Maybe this is still the right approach, just wondering at this time.

jaffinito: Well for ,NET we already download the lSI since it's required for this all to work.   We do remove it and any temp files after the install however.

recipes/newrelic/apm/dotNet/debian-systemd-nginx.yml
nr-ahemsath: This has been resolved by hosting a `gzip`ped version of our lsi in S3, so we don't need `unzip` anymore.

recipes/newrelic/apm/dotNet/debian-systemd-nginx.yml
jbeveland27: Just want to add a point of clarification for `processMatch: []`

This will _not_ match all cases, and actually does the opposite by matching nothing. The current usage of an empty `processMatch` is limited to the infra-agent and logs recipes, and those are special recipes that have a dedicated execution path through the CLI. The recipes that come after those are surfaced from the `recommendations` endpoint in GraphQL, and that only returns a match when the regex entries in `processMatch` match a given process command the CLI sends to that endpoint ([here](https://github.com/newrelic/newrelic-cli/blob/708c26bfd1fba7887a6fd06e95180fb1a0be00a1/internal/install/recipes/service_recipe_fetcher.go#L74) is where that input is created and a few lines down is where NerdGraph is called).

[Here's the service code](https://source.datanerd.us/open-source/open-install-library-service/blob/master/src/modules/recommendations/recommendations.service.ts#L22-L24) - notice it's designed so that it'll only return a recipe as "recommended" when a process command that the CLI sends matches an entry in `processMatch`. 

For a concrete example, view [GraphiQL query with processDetails](https://api.newrelic.com/graphiql?#query=%7B%0A%20%20docs%20%7B%0A%20%20%20%20openInstallation%20%7B%0A%20%20%20%20%20%20recommendations%28criteria%3A%20%7BprocessDetails%3A%20%5B%7Bname%3A%20%22java%22%7D%5D%2C%20installTarget%3A%20%7Btype%3A%20APPLICATION%2C%20os%3A%20LINUX%2C%20platform%3A%20DEBIAN%7D%7D%29%20%7B%0A%20%20%20%20%20%20%20%20count%0A%20%20%20%20%20%20%20%20results%20%7B%0A%20%20%20%20%20%20%20%20%20%20installTargets%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20os%0A%20%20%20%20%20%20%20%20%20%20%20%20type%0A%20%20%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%20%20%20%20name%0A%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D%0A) vs [GraphQL query without processDetails](https://api.newrelic.com/graphiql?#query=%7B%0A%20%20docs%20%7B%0A%20%20%20%20openInstallation%20%7B%0A%20%20%20%20%20%20recommendations%28criteria%3A%20%7BprocessDetails%3A%20%5B%5D%2C%20installTarget%3A%20%7Btype%3A%20APPLICATION%2C%20os%3A%20LINUX%2C%20platform%3A%20DEBIAN%7D%7D%29%20%7B%0A%20%20%20%20%20%20%20%20count%0A%20%20%20%20%20%20%20%20results%20%7B%0A%20%20%20%20%20%20%20%20%20%20installTargets%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20os%0A%20%20%20%20%20%20%20%20%20%20%20%20type%0A%20%20%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%20%20%20%20name%0A%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D%0A).

Not sure how helpful that is to the matter at hand, besides confirming that you _will_ need to include something in `processMatch` for this to work.

Julien4218: @jbeveland27 I guess I got the processMatch backward, but either way, something needs to be enhanced. We want the recipe to be picked up for the correct folks.
Thanks for clarifying.

recipes/newrelic/apm/dotNet/debian-systemd-nginx.yml
Julien4218: Closing this thread. An enhancement is now identified to support the kind of detection we would need for this recipe. So I think we could move forward, as long as we remove the `Apm` keyword to remove this recipe from the Guided install.
We would come back to update this recipe once the detection enhancement is ready in Virtuoso.

recipes/newrelic/apm/dotNet/debian-systemd-nginx.yml
```diff
@@ -0,0 +1,284 @@
...
+installTargets:
+  - type: application
+    os: linux
+    platform: "debian"
+  - type: application
+    os: linux
+    platform: "ubuntu"
+
+keywords:
+  - Apm
```
Julien4218, line 18: @nr-ahemsath  please remove `Apm` keyword for now, we'll add it back later to have this recipe included in guided install once the detection mechanism enhancement is ready in https://newrelic.atlassian.net/browse/VIRTUOSO-786

nr-ahemsath: Done, thanks.

Julien4218 (approved):

PR #332, opened 2021-04-30
Added several fixes for .net  windows install recipe

jaffinito:
Changed remove_any_previous task to precent bubbling up errors:
  added ErrorAction SilentlyContinue in a few places
  added Out-Null in one place
  wrapped both foreach interiors with try-catch
Fixed bug in configure task that created multiple distributedTracing elements
  Now checks if element exist and updates
  Will create if it does not exist
  Used Out-Null during creation to prevent leaking messages
Changed ensure_sites_started task to wrap sections in try-catch
  Each section is independently wrapped to allow one to function even if the other fails
Added message in verify_continue task to let user know we are starting the install
Removed all unecessary semi-colons from scripting
Improved formatting in script for readability

### DIFF
--- a/recipes/newrelic/apm/dotNet/windows/net-install.yml
+++ b/recipes/newrelic/apm/dotNet/windows/net-install.yml
@@ -75,6 +75,7 @@ install:
                 exit 130 # often used to represent Ctrl+C.
               }
               elseif ($NEW_RELIC_CONTINUE -ieq "y") {
+                Write-Host "Starting installation of the New Relic.NET Agent..."
                 break
               }
 
@@ -102,6 +103,7 @@ install:
             Write-Host -ForegroundColor Red "IIS is not installed."
             exit 5
           }
+
           $aspNet45Enabled = Get-WindowsOptionalFeature -Online -FeatureName IIS-ASPNET45 | select State
           if ($aspNet45Enabled.State -ne "Enabled") {
             Write-Host -ForegroundColor Red "ASP.NET is not installed."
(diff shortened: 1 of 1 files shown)


recipes/newrelic/apm/dotNet/windows/net-install.yml
```diff
@@ -226,14 +242,20 @@ install:
           }
 
           Write-Host "Sending requests to your applications..."
-          if ($NEW_RELIC_IIS_URL) {
-            Write-Host $NEW_RELIC_IIS_URL
-            while($val -ne 3) {
-              $val++
-              Invoke-WebRequest -UseBasicParsing -Method Get -TimeoutSec 5 -Uri $NEW_RELIC_IIS_URL -ErrorAction SilentlyContinue | Out-Null
+          try {
```
Julien4218, line 245: @jaffinito should we still try up to 3 times when it failed? It's most likely a timeout because the app doesn't start in 5 second, it might be good to try again?

recipes/newrelic/apm/dotNet/windows/net-install.yml
```diff
@@ -75,6 +75,7 @@ install:
                 exit 130 # often used to represent Ctrl+C.
               }
               elseif ($NEW_RELIC_CONTINUE -ieq "y") {
+                Write-Host "Starting installation of the New Relic.NET Agent..."
```
Julien4218, line 78: @jaffinito  a space between New Relic and .NET ?

jaffinito: :( fixing....

recipes/newrelic/apm/dotNet/windows/net-install.yml
jaffinito: I guess its a tradeoff.  If we check inside the while, we will always try 3 times, even if there are errors, but we take care of the slow start up problem.  If we check outside, we will only check once on errors, but miss the slow start up.  I was targeting the second use case initially, but I think give our goals with this work, the first case is the one to solve for.  I will put in a change to address it.

Julien4218 (approved):

PR #340, opened 2021-05-05
Fixed: infra agent install overwrites config files

noahmmcgivern:


recipes/newrelic/infrastructure/amazonlinux.yml
```diff
@@ -87,25 +87,23 @@ install:
...
+            sudo touch /etc/newrelic-infra.yml
+
+            sudo tee /etc/newrelic-infra.yml > /dev/null <<"EOT"
           license_key: {{.NEW_RELIC_LICENSE_KEY}}
           enable_process_metrics: true
           EOT
-        - |
-          if [ $(echo {{.NEW_RELIC_REGION}} | grep -i staging | wc -l) -gt 0 ]; then
-            echo 'staging: true' | sudo tee -a /etc/newrelic-infra.yml > /dev/null
+            if [ $(echo {{.NEW_RELIC_REGION}} | grep -i staging | wc -l) -gt 0 ]; then
```
Julien4218, line 97: @noahmmcgivern we should also have logic to remove the `staging: true` if the current install is not for staging

recipes/newrelic/infrastructure/amazonlinux.yml
```diff
@@ -87,25 +87,23 @@ install:
...
+              echo 'staging: true' | sudo tee -a /etc/newrelic-infra.yml > /dev/null
+            fi
           fi
 
     install_infra:
       cmds:
         - sudo curl -s -o /etc/yum.repos.d/newrelic-infra.repo https://download.newrelic.com/infrastructure_agent/linux/yum/el/6/x86_64/newrelic-infra.repo
         - sudo yum -y -q makecache --disablerepo='*' --enablerepo='newrelic-infra'
-        - sudo yum -y -q install newrelic-infra
+        - command -v newrelic-infra &>/dev/null || sudo yum -y -q install newrelic-infra
```
Julien4218, line 106: @noahmmcgivern I'm a bit curious as to why we're adding this other command before installing newrelic-infra.
Could you share details as to what this is doing and why it's needed, particularly for the scope of this MMF?

recipes/newrelic/infrastructure/logs/debian-logs.yml
```diff
@@ -98,55 +98,58 @@ install:
...
+                echo -e "  - name: configured-logs-$index\n    file: $filePath" | sudo tee -a /etc/newrelic-infra/logging.d/discovered.yml > /dev/null
+                index=$((index+1))
+              done
+            fi
 
-            for filePath in $(echo -n $discoveredLogFiles | sed "s/,/ /g"); do
-              echo -e "  - name: configured-logs-$index\n    file: $filePath" | sudo tee -a /etc/newrelic-infra/logging.d/discovered.yml > /dev/null
-              index=$((index+1))
-            done
+            # Enable infra-agent logs to New Relic
```
Julien4218, line 148: @noahmmcgivern this whole section (Enable infra-agent logs to New Relic) is now to be removed (already removed on main).
This was causing a large ingest (too verbose logging). We now rely on the default agent configuration for those values.

recipes/newrelic/infrastructure/logs/logs.yml
```diff
@@ -94,55 +94,58 @@ install:
...
+                echo -e "  - name: configured-logs-$index\n    file: $filePath" | sudo tee -a /etc/newrelic-infra/logging.d/discovered.yml > /dev/null
+                index=$((index+1))
+              done
+            fi
 
-            for filePath in $(echo -n $discoveredLogFiles | sed "s/,/ /g"); do
-              echo -e "  - name: configured-logs-$index\n    file: $filePath" | sudo tee -a /etc/newrelic-infra/logging.d/discovered.yml > /dev/null
-              index=$((index+1))
-            done
+            # Enable infra-agent logs to New Relic
```
Julien4218, line 144: @noahmmcgivern similar comment, to be removed

recipes/newrelic/infrastructure/logs/windows-logs.yml
```diff
@@ -37,9 +37,14 @@ install:
 
     setup:
       cmds:
-        - powershell -command '$logs = "C:\\Program Files\\New Relic\\newrelic-infra\\logging.d\\logs.yml"; if (Test-Path $logs) { Remove-Item $logs }'
-        - powershell -command 'echo "{{.LOGS_CONTENT}}" >> "C:\\Program Files\\New Relic\\newrelic-infra\\logging.d\\logs.yml"'
-        - powershell -command 'Restart-Service -Name newrelic-infra'
+        - |
+          powershell -command '
+          $logs = "C:\\Program Files\\New Relic\\newrelic-infra\\logging.d\\logs.yml"
+          if (-not (Test-Path $logs)) {
```
Julien4218, line 43: @noahmmcgivern why do we now have a "-not" in the condition? We didn't have this originally.
The logs.yml file is generated by this recipe, so it should be ok to remove with every install.

recipes/newrelic/infrastructure/ohi/apache/debian.yml
```diff
@@ -79,21 +79,17 @@ install:
...
         - |
           sudo apt-get update -yq
         - |
-          sudo apt-get install nri-apache -y
+          sudo apt-get install nri-apache -y -qq --no-upgrade
         - |
-          if [ -f /etc/newrelic-infra/integrations.d/apache-config.yml ]; then
-            sudo rm /etc/newrelic-infra/integrations.d/apache-config.yml;
-          fi
+          if [ ! -f /etc/newrelic-infra/integrations.d/apache-config.yml ]; then
```
Julien4218, line 87: @noahmmcgivern I don't think this is right If I understand correctly.
When re-installing an OHI (apache in this case), we should re-create the apache config file because the user may be inputting new values.
Also if we install a new OHI version (we do), we ll also want to have the latest published config which may contain new elements that the new OHI runtime would be looking for.
I think the intent of the ticket was focused on infra/log, not the OHI configs.
Let's circle back with the PM to clarify.

recipes/newrelic/infrastructure/windows.yml
```diff
@@ -26,63 +26,26 @@ install:
   tasks:
     default:
       cmds:
-        - task: remove_any_previous
         - task: install
         - task: assert_agent_started
 
-    remove_any_previous:
```
Julien4218, line 33: @noahmmcgivern we need to keep this.
The Windows infra agent needs to be un-installed (if it exists) before installing a new one.

recipes/newrelic/infrastructure/windows.yml
```diff
@@ -26,63 +26,26 @@ install:
...
         - powershell -command '$LICENSE_KEY="{{.NEW_RELIC_LICENSE_KEY}}"; msiexec.exe /qn /i "$env:TEMP\newrelic-infra.msi" GENERATE_CONFIG=true LICENSE_KEY="$LICENSE_KEY" | Out-Null;'
         - powershell -command '(New-Object System.Net.WebClient).DownloadFile("https://aka.ms/vs/16/release/vc_redist.x64.exe", "$env:TEMP\vc_redist.x64.exe");'
         - powershell -command 'start-process -FilePath "$env:TEMP\vc_redist.x64.exe" -ArgumentList "/install /q /norestart" -Verb RunAs -wait | Out-Null;'
-        - powershell -command 'Add-Content -Path "C:\\Program Files\\New Relic\\newrelic-infra\\newrelic-infra.yml" -Value "enable_process_metrics`:` true" -Force;'
-        - powershell -command '$NEW_RELIC_REGION="{{.NEW_RELIC_REGION}}"; if ($NEW_RELIC_REGION -ilike "staging") { Add-Content -Path "C:\\Program Files\\New Relic\\newrelic-infra\\newrelic-infra.yml" -Value "staging`:` true" -Force; };'
+        - |
+          powershell -command '
+          $InfraConfig = "C:\\Program Files\\New Relic\\newrelic-infra\\newrelic-infra.yml"
+          if (-not (Test-Path $InfraConfig)) {
```
Julien4218, line 41: @noahmmcgivern why the not? We should always create a new file if it's not there.
I believe we want to set the "enable_process_metrics" according to what the recipe says (true), we just don't want to override other values than the ones the install sets.
Similarly to linux, we'll also want to remove the "staging" element if it's not a staging deployment

noahmmcgivern:
Closing this PR and opening another after further discussion.

PR #341, opened 2021-05-06
Adding dotnet/linux regression tests

Julien4218:


Julien4218:
@nr-ahemsath those 2 new tests are now passing.
I had to put a fix for the recipe, it seemed not to detect staging correctly.

PR #345, opened 2021-05-07
fix: mysql password containing spaces causes connection error

noahmmcgivern:
This approach allows the use of a MySQL database password containing spaces, as well as a combination of single/double quotes and other special characters.


sanderblue (approved):
Nice :)

PR #349, opened 2021-05-12
Split windows infra install into multiple task to get more info when failing

Julien4218:


Julien4218:
Tested manually, with re-install as well, on windows server 2016 and 2019, using -c recipeUrl (to branch recipe location)

sanderblue (approved):

PR #352, opened 2021-05-13
.Net linux recipe for debian, redhat and centos.

vuqtran88:
.Net linux recipe for debian, redhat and centos.


jaffinito (commented):
This is great Vu!  I am glad you were able to get both distros into a single recipe!

Julien4218:
@jaffinito this looks pretty good.
You might consider moving the is_yum_installed to its own sh/var section like this https://github.com/newrelic/open-install-library/blob/main/recipes/newrelic/infrastructure/amazonlinux2.yml#L128-L132
I believe you need to update the installTarget on the recipe so it actually installs for all linux (including RHEL), it currently is only scoped to debian/ubuntu
Do we have any automated tests? You might want to add 1 for centos, and linux2

Julien4218 (approved):

Julien4218:
@gabeobrien feel free to review/merge, I'm good with this one.

PR #353, opened 2021-05-14
feat: Nginx recipe: allow user to retry inputData and detect status_module

alvarocabanas:
Changes to The Nginx recipe to allow the user retry inputting data and to not fail in case it is NGINX Plus that has another kind of response.


recipes/newrelic/infrastructure/ohi/nginx/debian.yml
```diff
@@ -40,49 +40,95 @@ validationNrql: "SELECT count(*) from NginxSample where hostname like '{{.HOSTNA
...
+        - |
+          tries=0
+          while [ $tries -lt {{.MAX_RETRIES}} ]; do
+            printf "\nNGINX status URL (default: http://127.0.0.1/status): "
+            read -r input
+            input=${input:-http://127.0.0.1/status}
+            code=$(curl -I $input -w %{response_code} -so '/dev/null')
+            if [ $code == "200" ]; then
+              echo $input > /tmp/nginx-ohai-input
+              chmod 600 /tmp/nginx-ohai-input && sudo chown root:root /tmp/nginx-ohai-input
```
alvarocabanas, line 83: This workaround saving the input data into a temporary file has been the only way we've found with go-task to pass dynamic data between tasks, because, so far, they don't have another way as stated in :
- https://github.com/go-task/task/issues/92
- https://github.com/go-task/task/pull/236

alvarocabanas: The file is set with only root reading permissions, it's removed after use, and, if, it would be a sensitive url with user and password, at the end that information is gonna be in the config.yml file anyway.

Julien4218:
@alvarocabanas this looks quite promising, thanks for all the work.
I reviewed and gave similar feedback to Carlos for the MySql.
We need to still allow data to be inputted in the recipe through environment variables, so for example the NR_CLI_STUB_STATUS_URL. The automated tests rely on that, and customers might too when installing in non-interactive mode.

Also the writing of data to /tmp file to deal with the passing of variable between go-tasks is not ideal. It would be better in the meantime to collapse all the go-tasks tasks into 1 (yes it is a go-tasks limitation for now).

When possible, try merging the latest main into your branch, we're having too many tests detected and run for no reasons somehow.

recipes/newrelic/infrastructure/ohi/nginx/rhel.yml
```diff
@@ -43,77 +43,103 @@ validationNrql: "SELECT count(*) from NginxSample where hostname like '{{.HOSTNA
...
+        - task: restart
 
-    assert_pre_req:
+    assert_infra:
       cmds:
         - |
           IS_INFRA_INSTALLED=$(sudo ps aux | grep newrelic-infra-service | grep -v grep | wc -l)
           if [ $IS_INFRA_INSTALLED -eq 0 ] ; then
             echo "The infrastructure agent is required to install this integration, we recommend going through our guided install path for this pre-requisite which can be found here:  https://docs.newrelic.com/docs/new-relic-guided-installation-overview" >> /dev/stderr
+            sudo rm /tmp/nginx-ohai-input
```
Julien4218, line 73: I believe this is no longer needed

recipes/newrelic/infrastructure/ohi/nginx/debian.yml
```diff
@@ -35,82 +35,107 @@ logMatch:
...
         - task: restart
 
-    assert_pre_req:
+    assert_infra:
       cmds:
         - |
           IS_INFRA_INSTALLED=$(sudo ps aux | grep newrelic-infra-service | grep -v grep | wc -l)
           if [ $IS_INFRA_INSTALLED -eq 0 ] ; then
             echo "The infrastructure agent is required to install this integration, we recommend going through our guided install path for this pre-requisite which can be found here:  https://docs.newrelic.com/docs/new-relic-guided-installation-overview" >> /dev/stderr
+            sudo rm /tmp/nginx-ohai-input
```
Julien4218, line 69: I believe this is no longer needed

recipes/newrelic/infrastructure/ohi/nginx/debian.yml
```diff
@@ -35,82 +35,107 @@ logMatch:
...
+          STATUS_RESPONSE=$(curl $NR_CLI_STUB_STATUS_URL -s)
+          if [ ! $(echo $STATUS_RESPONSE | grep "Active connections:" | wc -l) -eq 0 ] ; then
+            NR_CLI_STATUS_MODULE="ngx_http_stub_status_module"
+          elif [ ! $(echo $STATUS_RESPONSE | grep '"nginx_version"' | wc -l) -eq 0 ] ; then
+            NR_CLI_STATUS_MODULE="ngx_http_status_module"
+          elif [ ! $(echo $STATUS_RESPONSE | grep '"nginx"' | wc -l) -eq 0 ] ; then
+            NR_CLI_STATUS_MODULE="ngx_http_api_module"
+          else
+            echo "NGINX status must be configured for New Relic to collect data. \n - See https://docs.newrelic.com/docs/integrations/host-integrations/host-integrations-list/nginx-monitoring-integration#config for more info." >> /dev/stderr
+            sudo rm /tmp/nginx-ohai-input
```
Julien4218, line 108: I believe this is no longer needed

recipes/newrelic/infrastructure/ohi/nginx/rhel.yml
```diff
@@ -43,77 +43,103 @@ validationNrql: "SELECT count(*) from NginxSample where hostname like '{{.HOSTNA
...
+          STATUS_RESPONSE=$(curl $NR_CLI_STUB_STATUS_URL -s)
+          if [ ! $(echo $STATUS_RESPONSE | grep "Active connections:" | wc -l) -eq 0 ] ; then
+            NR_CLI_STATUS_MODULE="ngx_http_stub_status_module"
+          elif [ ! $(echo $STATUS_RESPONSE | grep '"nginx_version"' | wc -l) -eq 0 ] ; then
+            NR_CLI_STATUS_MODULE="ngx_http_status_module"
+          elif [ ! $(echo $STATUS_RESPONSE | grep '"nginx"' | wc -l) -eq 0 ] ; then
+            NR_CLI_STATUS_MODULE="ngx_http_api_module"
+          else
+            echo "NGINX status must be configured for New Relic to collect data. \n - See https://docs.newrelic.com/docs/integrations/host-integrations/host-integrations-list/nginx-monitoring-integration#config for more info." >> /dev/stderr
+            sudo rm /tmp/nginx-ohai-input
```
Julien4218, line 112: I believe this is no longer needed

recipes/newrelic/infrastructure/ohi/nginx/suse.yml
```diff
@@ -32,81 +32,107 @@ logMatch:
...
+        - task: restart
 
-    assert_pre_req:
+    assert_infra:
       cmds:
         - |
           IS_INFRA_INSTALLED=$(sudo ps aux | grep newrelic-infra-service | grep -v grep | wc -l)
           if [ $IS_INFRA_INSTALLED -eq 0 ] ; then
             echo "The infrastructure agent is required to install this integration, we recommend going through our guided install path for this pre-requisite which can be found here:  https://docs.newrelic.com/docs/new-relic-guided-installation-overview" >> /dev/stderr
+            sudo rm /tmp/nginx-ohai-input
```
Julien4218, line 66: I believe this is no longer needed

recipes/newrelic/infrastructure/ohi/nginx/suse.yml
```diff
@@ -32,81 +32,107 @@ logMatch:
...
+          STATUS_RESPONSE=$(curl $NR_CLI_STUB_STATUS_URL -s)
+          if [ ! $(echo $STATUS_RESPONSE | grep "Active connections:" | wc -l) -eq 0 ] ; then
+            NR_CLI_STATUS_MODULE="ngx_http_stub_status_module"
+          elif [ ! $(echo $STATUS_RESPONSE | grep '"nginx_version"' | wc -l) -eq 0 ] ; then
+            NR_CLI_STATUS_MODULE="ngx_http_status_module"
+          elif [ ! $(echo $STATUS_RESPONSE | grep '"nginx"' | wc -l) -eq 0 ] ; then
+            NR_CLI_STATUS_MODULE="ngx_http_api_module"
+          else
+            echo "NGINX status must be configured for New Relic to collect data. \n - See https://docs.newrelic.com/docs/integrations/host-integrations/host-integrations-list/nginx-monitoring-integration#config for more info." >> /dev/stderr
+            sudo rm /tmp/nginx-ohai-input
```
Julien4218, line 105: I believe this is no longer needed

recipes/newrelic/infrastructure/ohi/nginx/debian.yml
```diff
@@ -35,82 +35,107 @@ logMatch:
...
-          STUB_STATUS_ENABLED=$(curl {{.NR_CLI_STUB_STATUS_URL}} -s | grep "Active connections:" | wc -l)
-          if [ $STUB_STATUS_ENABLED -eq 0 ] ; then
-            echo "NGINX HTTP status module must be configured for New Relic to collect data. \n - See https://docs.newrelic.com/docs/integrations/host-integrations/host-integrations-list/nginx-monitoring-integration#config for more info." >> /dev/stderr
+          tries=0
+
+          # Check Env Vars
+          NR_CLI_STUB_STATUS_URL="{{.NR_CLI_STUB_STATUS_URL}}"
+          if [[ "$NR_CLI_STUB_STATUS_URL" == "" ]]; then
+            while [ $tries -lt {{.MAX_RETRIES}} ]; do
+              printf "\nNGINX status URL (default: http://127.0.0.1/status): "
```
Julien4218, line 82: If the install is run with assumeYes (-y) which is this go-task variable {{.NEW_RELIC_ASSUME_YES}} , then we should not prompt for the URL and use the default value of http://127.0.0.1/status

One step further perhaps, and independently of assumeYes, maybe we could try to see if the default URL works, and only prompt if we see it doesn't work?

carlossscastro: @Julien4218 

Regarding AssumeYes, we are reading the env vars, which I assume are being injected by the test unit, and if empty we proceed with user input. Is this approach ok or should we use {{.NEW_RELIC_ASSUME_YES}}?

As for your second suggestion, that's a good idea. We'll keep that in mind for a follow up review of this recipe.
Thanks!

Julien4218: Yes, what you've done with the reading of the env vars is great, this mainly supports our test framework.
However customer would rely on the AssumeYes option to essentially say "just install with all defaults", and some customers may potentially inject the specific values through env var (just like our test) but this would be optional (maybe different value than default).
So in short, we should support both, and always prompt if there is no env var value or no default value.

carlossscastro: Understood now @Julien4218 . Thanks for the details.

I'll change that and resubmit.
Thanks!

carlossscastro:
@Julien4218 

AssumeYes support added now.

Julien4218 (approved):
This looks good @alvarocabanas and @carlossscastro, I'll move this forward.

PR #354, opened 2021-05-14
Mysql recipe improvements

carlossscastro:
Changes to The Mysql recipe:

* Added connectivity test to mysql instance

### DIFF
--- a/recipes/newrelic/apm/java/debian.yml
+++ b/recipes/newrelic/apm/java/debian.yml
@@ -31,34 +31,300 @@ install:
   version: "3"
   silent: true
 
+  vars:
+    TMP_DIR: '/tmp/newrelic-java'
+
+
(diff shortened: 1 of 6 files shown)


Julien4218:
@carlossscastro thanks for all this work, this looks like we're going in a good direction.
Could you get the latest changes on main, there seems to be a few conflicts to resolve.

I wanted to mention, we need to allow a non-interactive usage of the recipe, the automated tests rely on that, and customers are potentially using that mechanism to install as well. so all the previous NR_XXX for the inputs, could contain a valuable default value from an environment variable.

Also, the marshalling of variable into /tmp file is not ideal, particularly for username/password. We should collapse the various go-tasks tasks to avoid writing files that get read later on (limitation of go-tasks for the time been).

PR #355, opened 2021-05-14
Adds persistence to the java recipe

meiao:


recipes/newrelic/apm/java/debian.yml
```diff
@@ -31,34 +31,300 @@ install:
   version: "3"
   silent: true
 
+  vars:
+    TMP_DIR: '/tmp/newrelic-java'
```
gabeobrien, line 35: I have raised the idea of introducing some random value into the path.  We will address this in future work on the recipe.

gabeobrien (approved):
Looks like some good work here!

PR #356, opened 2021-05-14
AWS Linux 1 Logs recipe is unsupported in order to prevent failure

noahmmcgivern:


sanderblue (approved):

Julien4218:
@noahmmcgivern while looking at the data on live, it looks like there isn't a platformVersion of "1" for the older linux.
image

We might need to flip the criteria the other way:
- make specific version targeting amazon linux 2, with the correct logging recipe
- make generic version for amazon, and surface a warning that logging is not supported on this version of amazon linux

Julien4218 (changes requested):
There is no amazon linux platform family of "1" for the older amazon linux

Julien4218 (approved):

PR #357, opened 2021-05-18
Mysql improvements

carlossscastro:
Changes to The Mysql recipe:

* Added connectivity test to mysql instance

### DIFF
--- a/recipes/newrelic/infrastructure/ohi/mysql/debian.yml
+++ b/recipes/newrelic/infrastructure/ohi/mysql/debian.yml
@@ -39,37 +39,18 @@ logMatch:
 
 # NRQL the newrelic-cli will use to validate the agent/integration this recipe
 # installed is successfully sending data to New Relic
-validationNrql: "SELECT count(*) from MysqlSample where hostname like '{{.HOSTNAME}}' FACET entityGuid SINCE 10 minutes ago"
-
-# Prompts for input from the user. These variables then become
-# available to go-task in the form of {{.VAR_NAME}}
-inputVars:
-  - name: "NR_CLI_DB_USERNAME"
-    prompt: "MySQL Username"
-  - name: "NR_CLI_DB_PASSWORD"
-    prompt: "MySQL Password"
-    secret: true
-  - name: "NR_CLI_DB_HOSTNAME"
-    prompt: "MySQL Hostname (default: localhost)"
-    default: "localhost"
-  - name: "NR_CLI_DB_PORT"
-    prompt: "MySQL port (default: 3306)"
-    default: 3306
-  - name: "NR_CLI_DATABASE"
-    prompt: "MySQL Database Name"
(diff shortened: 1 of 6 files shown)


carlossscastro:
@Julien4218 Thuis new PR already merges all tasks into a single task to avoid using tmp files to hold var values.

Julien4218:
@carlossscastro the tests are currently timing out because they re waiting for an input to be entered. Similarly, we also want to support the non-interactive flow (assume yes).
Here is an example from the apm/.net linux recipe. https://github.com/newrelic/open-install-library/blob/main/recipes/newrelic/apm/dotNet/debian-systemd.yml#L71-L90
We'll want to do something similar for all the inputs (the value can be injected through an env var).
Here is an example how the tests are writing the value in env vars (so the recipe can pick those up) https://github.com/newrelic/open-install-library/blob/main/test/deploy/linux/mysql/install/debian/roles/configure/tasks/main.yml#L55-L66

Julien4218 (approved):

carlossscastro:
> @carlossscastro this looks very good to me.
> One caveats normally, if AssumeYes option is used, we should use the default value instead of prompting. I don't see this behavior happening anymore. The 2 questions for NR_CLI_DB_HOSTNAME and NR_CLI_DB_PORT should be defaulted when assume yes is specified.
> With that said, I think what you have is a good first step, and an amelioration of what we had before, so I'm happy to move this forward, and maybe keep in mind the defaulting in a subsequent iteration. I also think we shouldn't prompt for hostname and hard-wire to localhost since the assumption is that this should be local to the host anyway (we don't support distributed/remote install for now).

Hi @Julien4218 Thanks for you comments.

Regarding the points you mention:
1. HOSTNAME and PORT are set by using default values if the user just press enter on those inputs. you can see that here:
https://github.com/newrelic/open-install-library/pull/357/files#diff-69beb59d1ef06f584f89a6673b0a833bd3f1d38ec587d954cabe7bf74443acb1R92

and here:
https://github.com/newrelic/open-install-library/pull/357/files#diff-69beb59d1ef06f584f89a6673b0a833bd3f1d38ec587d954cabe7bf74443acb1R95

Those 2 lines will set the values as localhost and 3306 if the user just bypasses them pressing enter (assume default)

2. Regarding hardcoding hostname as localhost, it's true that we don't support remote install for now, however some users use the hostname to set their permissions and the integration credentials must match what they have set in their mysql environment.
Example: if I'm installing the integration on a host named `hostname1` and set my mysql permissions to allow user `newrelic` to login from `hostname1`, I specifically need to to setup the integration to connect to `hostname1`. If I try to connect using localhost, although they resolve to the same machine, mysql will only accept connection made to `hostname1`
It all depends on how the users and permissions are set on mysql.

We have seen this happening before where users had to change localhost by the "hostname" to have the integration working.

Julien4218:
Got it, thanks for the details @carlossscastro 

Side note, AssumeYes means we should not prompt at all (when default exist). It's not as important in this recipe (because several questions dont have default and must be answered).

PR #359, opened 2021-05-19
Added send_transaction, create_shared, delete_shared, detect_services, and updated restart_services

zsistla:
Added additional tasks:
* `send_transaction`
* `create_shared`
* `delete_shared`
* `detect_services`

updated `restart_services`

Original below:
Goal is to make minimal changes to the user’s environment and to reduce the amount of questions/interactions.

The php agent needs one transaction to initialize the agent then subsequent transactions are sent to NR1.  Normally, calling `php --ini` will always use the CLI `php.ini` and the associated `newrelic.ini`.

However, by calling something like `php -c /etc/php/7.2/apache2/php.ini -c /etc/php/7.2/apache2/conf.d/newrelic.ini -n --ini`, we force the command to use ONLY the ini files used by the web server (it is forced to ignore all other INIs) and can thereby verify the data flowing pipeline without other disadvantages.

This has the advantages of:
1) Verifies the data pipeline with the `php.ini` and `newrelic.ini` files that the web server will use by verifying the interoperability of those two `ini` files.
2) Doesn’t require the user to send data.
3) Doesn’t ask the user for the website URL in non-silent mode.
4) Doesn’t make us guess about URLs or detected ip/port pairs that may or may not be correct.
5) Curl doesn’t always trigger PHP scripts so we could curl to a valid page and still not get anything (for instance, Wordpress plugins aren’t triggered).
6) We don’t have to make a test.php to run on the user’s website.

To test:
1) deployed:
```
docker run -it \
-v <<local dir>>/demo-deployer/configs/:/mnt/deployer/configs/ \
-v <<local dir>>/open-install-library/test/:/mnt/deployer/test/ \
-v <<local dir>>/open-install-library/:/mnt/open-install-library \
--entrypoint ruby ghcr.io/newrelic/deployer:latest main.rb \
-c /mnt/deployer/configs/mycreds.credentials.local.json \
-d "test/manual/definitions/apm/php-agent/php-apache-wordpress-ubuntu18.json" -l debug
```

2) installed `9.17.1.301` php agent 
3) `sudo NEW_RELIC_REGION=STAGING NEW_RELIC_CLI_PRERELEASEFEATURES=ALLOW NEW_RELIC_API_KEY=<<add api key>> NEW_RELIC_ACCOUNT_ID=<<add account id>> /usr/local/bin/newrelic install -c <<add path to this recipe>>`

One caveat, it assumes the installation step uses the following temporary directory:
`TMP_INSTALL_DIR="/tmp/php-agent-install”` which is then deleted as the last step of the recipe.
@kneitinger , please let me know if this can be integrated into your install script logic.

Note, while the `send_transaction` section is intended to be run immediately after the installation and that task will only generate web server `ini` related transactions, if the agent is already installed and the recipe is run again on the same machine, it will currently generate some additional cli transactions as well just by nature of the other tasks that are run.

### DIFF
--- a/recipes/newrelic/apm/php-agent/php.yml
+++ b/recipes/newrelic/apm/php-agent/php.yml
@@ -219,9 +219,11 @@ install:
           # Web server was detected. Send transaction associated with the webserver php.ini
           # and the newrelic.ini file associated with that web server.
           #
-             sudo -u ${nonpriv_user} php -c "${WEB_PHP_INI_DIR}/php.ini" -c "${WEB_NR_INI_DIR}/newrelic.ini" -n --ini &>/dev/null
-             sudo -u ${nonpriv_user} php -c "${WEB_PHP_INI_DIR}/php.ini" -c "${WEB_NR_INI_DIR}/newrelic.ini" -n --ini &>/dev/null
-             sudo -u ${nonpriv_user} php -c "${WEB_PHP_INI_DIR}/php.ini" -c "${WEB_NR_INI_DIR}/newrelic.ini" -n --ini &>/dev/null
+          for i in {1..10}; do
+            sleep 1
+            sudo -u $nonpriv_user php -c "${WEB_PHP_INI_DIR}/php.ini" -c "${WEB_NR_INI_DIR}/newrelic.ini" -n --ini &>/dev/null
+            #sudo -u www-data php -c /etc/php/7.2/fpm/php.ini -c /etc/php/7.2/fpm/conf.d/newrelic.ini -n --ini &>/dev/null
+          done
           else
           #
           # No web server detected, pure PHP CLI.
--- a/recipes/newrelic/apm/php-agent/php.yml
+++ b/recipes/newrelic/apm/php-agent/php.yml
@@ -77,8 +77,20 @@ install:
           do
             priv_user=$(ps auxf | egrep $PROCESS | grep -v grep | head -n1 | awk '{print $1}')
             full_process_name=$(ps auxf | egrep $PROCESS | grep -v grep | head -n1 | awk '{print $11}')
-            process_name=$(echo $full_process_name | sed -n 's/.*\/\(.*\)$/\1/p')
-
+            fpm_nginx_process_name=$(echo "${full_process_name}" | sed -n 's/\(.*\):$/\1/p')
+            if [ -n "${fpm_nginx_process_name}" ]; then
+              fpm_process_loc=$(ps auxf | egrep $PROCESS | grep -v grep | head -n1 | awk '{print $14}')
+              fpm_ver=$(echo "${fpm_process_loc}" | sed -n 's/.*\/php\/\([578].[0-9]\)\/.*/\1/p')
+              if [ -n "${fpm_ver}" ]; then
+                # It's fpm.
+                process_name="php${fpm_ver}-fpm"
+              else
+                # It's nginx.
+                process_name=$fpmnginx_process_name
+              fi
+            else
+              process_name=$(echo $full_process_name | sed -n 's/.*\/\(.*\)$/\1/p')
+            fi
             nonpriv_user=
 
             if [ -n "${priv_user}" ]; then
@@ -187,7 +199,7 @@ install:
             # Exit with Agent install failure code exit(16)
             exit 16
           fi
-
+   
           #
           # Get the non-privileged user that was previously stored.
           #
--- a/recipes/newrelic/apm/php-agent/php.yml
+++ b/recipes/newrelic/apm/php-agent/php.yml
@@ -131,17 +131,27 @@ install:
     restart_services:
       cmds:
         - echo "Restarting the services as needed to pick up newrelic.ini information."
+          #
+          # Case: php-fpm/apache or php-fpm/nginx, only php-fpm needs to be restarted.
+          # Case: apache only or nginx only, they respective service needs to be restarted.
+          # This means only restart the first process in processes_to_restart.txt.
+          #
         - |
           if [ "{{.NR_PHP_RESTART}}" = "y" ]  && [ -f "{{.TMP_INSTALL_DIR}}/processes_to_restart.txt" ]; then
             cd {{.TMP_INSTALL_DIR}}
-            while read -r process user_name; do
-              echo "Restarting $process as privileged user $user_name"
-              sudo -u $user_name service $process restart
-              echo "Sleeping for 10 seconds to give process time to recover."
-              sleep 10
-            done < "{{.TMP_INSTALL_DIR}}/processes_to_restart.txt"
+            read -r process user_name < "{{.TMP_INSTALL_DIR}}/processes_to_restart.txt"
+            echo "Restarting $process as privileged user $user_name"
+            sudo -u $user_name service $process restart
+            echo "Sleeping for 10 seconds to give process time to recover."
+            sleep 10
           else
-            echo "You will need to restart your PHP application in order for instrumentation to be enabled."
+            if [ -f "{{.TMP_INSTALL_DIR}}/processes_to_restart.txt" ]; then
+              cd {{.TMP_INSTALL_DIR}}
+              read -r process user_name < "{{.TMP_INSTALL_DIR}}/processes_to_restart.txt"
+              echo "Please restart  $process as privileged user $user_name for instrumentation to be enabled."
+            else
+              echo "You will need to restart your PHP web server in order for web instrumentation to be enabled."
+            fi
           fi
 
     send_transaction:
--- a/recipes/newrelic/apm/php-agent/php.yml
+++ b/recipes/newrelic/apm/php-agent/php.yml
@@ -91,12 +91,12 @@ install:
               #
               echo "$process_name $priv_user" >> {{.TMP_INSTALL_DIR}}/processes_to_restart.txt
               #
-              # Output all processes to be restarted to a file with the following forat:
-              # process_name privileged_user
-              # When initiating the transaction, it will be ONLY be used witt
+              # Save all non-privileged users associated with detected processes.
+              # When initiating the transaction, it will ONLY be used with
               # the first associated non-privileged user because due the reasoning above
               # that the installation logic adheres to, between FPM and Apache,
-              # FPM will be the one to use.
+              # FPM will be the one to use and `newrelic-install.sh` only
+              # installs the `newrelic.ini` in one web location.
               #
               echo "$nonpriv_user" >> {{.TMP_INSTALL_DIR}}/nonpriv_users.txt
             fi
--- a/recipes/newrelic/apm/php-agent/php.yml
+++ b/recipes/newrelic/apm/php-agent/php.yml
@@ -71,7 +71,7 @@ install:
           #
           declare -a guided_support=("(fpm)" "(httpd|apache2|apache)" "(nginx)")
 
-          current_user=$(whoami)
+          current_user=$SUDO_USER
 
           for PROCESS in ${guided_support[@]}
           do
@@ -177,7 +177,7 @@ install:
           if [ -f {{.TMP_INSTALL_DIR}}/nonpriv_users.txt ]; then
           nonpriv_user=$(head -1 {{.TMP_INSTALL_DIR}}/nonpriv_users.txt)
           else
-            nonpriv_user=$(whoami)
+            nonpriv_user=$SUDO_USER
           fi
 
           #
@@ -197,9 +197,10 @@ install:
           #
           # No web server detected, pure PHP CLI.
           #
-             php —-ini
-             php —-ini
-             php —-ini
+            echo "Running PHP-CLI with user $SUDO_USER"
+            sudo -u $SUDO_USER php —-ini
+            sudo -u $SUDO_USER php —-ini
+            sudo -u $SUDO_USER php —-ini
           fi
         - echo "Done sending transactions."
 
--- a/recipes/newrelic/apm/php-agent/php.yml
+++ b/recipes/newrelic/apm/php-agent/php.yml
@@ -41,11 +41,12 @@ install:
       cmds:
         - task: create_shared
         - task: detect_services
-              #        - task: install_packages
-              #- task: config_newrelic_ini
-              #- task: finalize_agent_config
+        - task: install_packages
+        - task: config_newrelic_ini
+        - task: finalize_agent_config
         - task: restart_services
         - task: send_transaction
+        - task: delete_shared
 
     create_shared:
       cmds:
@@ -61,8 +62,6 @@ install:
           rm -f {{.TMP_INSTALL_DIR}}/nonpriv_users.txt
           rm -f {{.TMP_INSTALL_DIR}}/processes_to_restart.txt
 
-          SUPPORTED_PROCESS=0
-          PROCESS_TERMS=1
           #
           # For guided installs, we currently support `php-fpm`, `apache`, and `nginx`.
           # There is a reason for the order.  According our install script:
--- a/recipes/newrelic/apm/php-agent/php.yml
+++ b/recipes/newrelic/apm/php-agent/php.yml
@@ -16,6 +16,8 @@ keywords:
 processMatch:
   - /php-fpm/
 
+validationNrql: "SELECT count(*) from Transaction WHERE host like '{{.HOSTNAME}}%' facet entityGuid since 10 minutes ago"
+
 inputVars:
   - name: "NR_PHP_APPLICATION"
     prompt: "What is the name of your PHP application?"
@@ -37,6 +39,7 @@ install:
         - task: config_newrelic_ini
         - task: finalize_agent_config
         - task: restart_services
+        - task: send_transaction
 
     install_packages:
       cmds:
@@ -48,7 +51,6 @@ install:
           fi
           DEBIAN_FRONTEND=noninteractive sudo apt update
           DEBIAN_FRONTEND=noninteractive apt -y install newrelic-php5
-
     config_newrelic_ini:
       cmds:
         - echo "TODO: configure newrelic.ini"
@@ -67,3 +69,45 @@ install:
           else
             echo "You will need to restart your PHP application in order for instrumentation to be enabled."
           fi
+    send_transaction:
+      cmds:
+        - echo "Send queryable transactions."
+        - |
+          TMP_INSTALL_DIR="/tmp/php-agent-install"
+          cd $TMP_INSTALL_DIR
+          #
+          # Expand the log files from the most recent installation attempt,
+          # and get the name of the most recent log file.
+          #
+          tar -xvf $(ls -t nrinstall*.tar | head -1)
+          NR_INSTALL_LOG=$(ls -t nrinstall*.log | head -1)
+          #
+          # Get the web server PHP INI directories that the most recent installation attempt used.
+          #
+          WEB_PHP_INI_DIR=$(sed -n 's/.*pi_inidir_dso=\(.*\)\/conf.d/\1/p' $NR_INSTALL_LOG)
+          WEB_NR_INI_DIR=$(sed -n 's/.*pi_inidir_dso=\(.*\/\)/\1/p' $NR_INSTALL_LOG)
+          if [ -z "${WEB_PHP_INI_DIR}" ]; then
+              WEB_PHP_INI_DIR=$WEB_NR_INI_DIR
+          fi
+          #
+          # The first transaction initializes the app.
+          # The second transaction is for luck.
+          # The third time’s the charm.
+          #
+          if [ -n "${WEB_PHP_INI_DIR}" ]; then
+          #
+          # Web server was detected. Send transaction associated with the webserver php.ini
+          # and the newrelic.ini file associated with that web server.
+          #
+             php -c "${WEB_PHP_INI_DIR}/php.ini" -c "${WEB_NR_INI_DIR}/newrelic.ini" -n --ini &>/dev/null
+             php -c "${WEB_PHP_INI_DIR}/php.ini" -c "${WEB_NR_INI_DIR}/newrelic.ini" -n --ini &>/dev/null
+             php -c "${WEB_PHP_INI_DIR}/php.ini" -c "${WEB_NR_INI_DIR}/newrelic.ini" -n --ini &>/dev/null
+          else
+          #
+          # No web server detected, pure PHP CLI.
+          #
+             php —-ini
+             php —-ini
+             php —-ini
+          fi
+        - echo "Done sending transactions."
--- a/recipes/newrelic/apm/php-agent/php.yml
+++ b/recipes/newrelic/apm/php-agent/php.yml
@@ -51,6 +51,7 @@ install:
           fi
           DEBIAN_FRONTEND=noninteractive sudo apt update
           DEBIAN_FRONTEND=noninteractive apt -y install newrelic-php5
+
     config_newrelic_ini:
       cmds:
         - echo "TODO: configure newrelic.ini"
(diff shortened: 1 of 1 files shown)


joshuabenuck:
There is a risk with this approach that we will start the daemon using a different user than the one used by the web server which could result in permissions problems.

recipes/newrelic/apm/php-agent/php.yml
```diff
@@ -67,3 +70,47 @@ install:
           else
             echo "You will need to restart your PHP application in order for instrumentation to be enabled."
           fi
+
+    send_transaction:
+      cmds:
+        - echo "Send queryable transactions."
+        - |
+          TMP_INSTALL_DIR="/tmp/php-agent-install"
```
joshuabenuck, line 78:
```suggestion
          TMP_INSTALL_DIR=$(mktemp -d -t php-agent-install-XXXXX)
```

recipes/newrelic/apm/php-agent/php.yml
```diff
@@ -67,3 +70,47 @@ install:
...
+             php -c "${WEB_PHP_INI_DIR}/php.ini" -c "${WEB_NR_INI_DIR}/newrelic.ini" -n --ini &>/dev/null
+             php -c "${WEB_PHP_INI_DIR}/php.ini" -c "${WEB_NR_INI_DIR}/newrelic.ini" -n --ini &>/dev/null
+             php -c "${WEB_PHP_INI_DIR}/php.ini" -c "${WEB_NR_INI_DIR}/newrelic.ini" -n --ini &>/dev/null
+          else
+          #
+          # No web server detected, this was a pure PHP CLI php agent installation.
+          #
+             php —-ini
+             php —-ini
+             php —-ini
```
joshuabenuck, line 114: I expect we will want to know how many of our virtuoso based installs go through this path vs the web path. How do other Virtuoso based agent installs track such things? A supportability metric is what comes to mind, but I don't think we have an exposed API to send one. Creating a custom PHP script for each path here would put the data in the customer's account which doesn't help us look for trends across the install base. Are we only able to track failure causes via return codes currently?

This might be something we need to sort out later. I just wanted to start the conversation.

recipes/newrelic/apm/php-agent/php.yml
zsistla: This is assuming the tmp install directory was already created and used in prior steps (it assumes the install ran install from that directory).  Should not be created here, but ideally created in prior step and referenced here.

recipes/newrelic/apm/php-agent/php.yml
```diff
@@ -67,3 +70,47 @@ install:
...
+        - echo "Send queryable transactions."
+        - |
+          TMP_INSTALL_DIR="/tmp/php-agent-install"
+          cd $TMP_INSTALL_DIR
+          #
+          # Expand the log files from the most recent installation attempt,
+          # and get the name of the most recent log file.
+          #
+          tar -xvf $(ls -t nrinstall*.tar | head -1)
+          NR_INSTALL_LOG=$(ls -t nrinstall*.log | head -1)
```
joshuabenuck, line 85: We should guard against a missing tar file or log file. Unlikely to happen, but may result in us following the non-web path when there is actually an installation problem.

zsistla:
> There is a risk with this approach that we will start the daemon using a different user than the one used by the web server which could result in permissions problems.

@joshuabenuck excellent point.  I think I can `ps` the processes and infer the user then run the command as that user.  Can I assume that the webserver process (apache vs fpm/nginx) has already been detected and propogated somewhere (via text in the tmp directory)?

recipes/newrelic/apm/php-agent/php.yml
joshuabenuck: Using a randomized temp directory name helps guard against multiple installs conflicting with one another and makes certain attacks slightly more difficult to execute.

In order for us to be able to do this, we will need to be able to either access env vars set by previous steps or will need to be able to share shell vars across steps. Is a fresh shell env used for each step or do they all share a common one?

joshuabenuck:
> > There is a risk with this approach that we will start the daemon using a different user than the one used by the web server which could result in permissions problems.
> 
> @joshuabenuck excellent point. I think I can `ps` the processes and infer the user then run the command as that user. Can I assume that the webserver process (apache vs fpm/nginx) has already been detected and propogated somewhere (via text in the tmp directory)?

That sounds like a reasonable assumption. For our first round of recipes we are targeting fpm based installs independent of the web server used (apache and nginx are the ones we are validating against, but it is possible caddy or some other web server could be present). To start, the recipe can just assume that php-fpm is the process to look for until we get to the point where that varies.

recipes/newrelic/apm/php-agent/php.yml
zsistla: Good point about the possibility for attack against known directory names.

I wasn't sure about fresh shell vs shared, but since I've seen other agents use text to share info so I'd assumed fresh shell.

Although the directory will be created/destroyed each run, because of how the agent would be downloaded/installed, multiple install attempts could theoretically exist in the same dir (i.e., the log file from installation attempts is timestamped) and why the script looks for the most recent log file not just a log file.  

I think it is a fresh shell so it'd be hard to pass the var name between tasks.  Also, given it's created/torn down with each run, slightly less worried about probability of attack.

zsistla: Added blocks for now that will create/delete similar to how dotnet does [creates](https://github.com/newrelic/open-install-library/blob/e7347282f607e079f528d8199b678f29dbeb4fcb/recipes/newrelic/apm/dotNet/linux-systemd.yml#L129) / [deletes](https://github.com/newrelic/open-install-library/blob/e7347282f607e079f528d8199b678f29dbeb4fcb/recipes/newrelic/apm/dotNet/linux-systemd.yml#L336).  Exact naming scheme can still be under discussion.

joshuabenuck: Looks like there's a way to declare variables in recipes. Here's an example from a Java recipe Gabe found: https://github.com/newrelic/open-install-library/blob/0f621414cda830c4eea15752e4ad65847df82897/recipes/newrelic/apm/java/linux.yml#L32

zsistla: Cool, thanks!  I'll get that integrated.

recipes/newrelic/apm/php-agent/php.yml
```diff
@@ -29,20 +29,80 @@ inputVars:
...
+      cmds:
+        - echo "Detect services."
+        - |
+          cd {{.TMP_INSTALL_DIR}}
+          rm -f {{.TMP_INSTALL_DIR}}/nonpriv_users.txt
+          rm -f {{.TMP_INSTALL_DIR}}/processes_to_restart.txt
+
+          #
+          # For guided installs, we currently support `php-fpm`, `apache`, and `nginx`.
+          # There is a reason for the order.  According our install script:
```
joshuabenuck, line 67: If "install script" is a reference to the deploy config, the decision made there was merely a convenience since we have control of that environment. It isn't something we should rely on or mention here.

recipes/newrelic/apm/php-agent/php.yml
```diff
@@ -29,20 +29,80 @@ inputVars:
...
+          # If both Apache and FPM are installed, we want FPM to win:
+          # while there are many ways to end up with libapache2-mod-php5 installed,
+          # it's unlikely php5-fpm will be installed unless FPM is actually in use.
```
joshuabenuck, line 70: If both FPM and Apache are present, FPM is likely in use (or at least intended to be in use), but any case where it ends up not being used is a potential failed installed. In my testing of the deploy configs, I ran into several cases where I had PHP-FPM installed, but had not configured Apache properly to actually use it. I recommend that we either restart both or figure out whether Apache is configured to use FPM (e.g. parse configs, run a phpinfo to look for "Server API", or grep our logs for the sapi in use - this last one requires that we run a non-php-cli process).

recipes/newrelic/apm/php-agent/php.yml
```diff
@@ -71,10 +124,16 @@ install:
 
     restart_services:
       cmds:
+        - echo "Restarting the services as needed to pick up newrelic.ini information."
         - |
-          if [ -z "{{.NR_PHP_RESTART}}" ]; then
-            echo "Restarting php-fpm"
-            sudo service php-fpm restart
+          if [ "{{.NR_PHP_RESTART}}" = "y" ]  && [ -f "{{.TMP_INSTALL_DIR}}/processes_to_restart.txt" ]; then
```
joshuabenuck, line 129: If non-silent installs, when will the user be prompted about the restart? (I ask because I want to us to include the process names that will be restarted in the associated prompt).

recipes/newrelic/apm/php-agent/php.yml
```diff
@@ -110,12 +190,12 @@ install:
           # Web server was detected. Send transaction associated with the webserver php.ini
           # and the newrelic.ini file associated with that web server.
           #
-             php -c "${WEB_PHP_INI_DIR}/php.ini" -c "${WEB_NR_INI_DIR}/newrelic.ini" -n --ini &>/dev/null
-             php -c "${WEB_PHP_INI_DIR}/php.ini" -c "${WEB_NR_INI_DIR}/newrelic.ini" -n --ini &>/dev/null
-             php -c "${WEB_PHP_INI_DIR}/php.ini" -c "${WEB_NR_INI_DIR}/newrelic.ini" -n --ini &>/dev/null
+             sudo -u ${nonpriv_user} php -c "${WEB_PHP_INI_DIR}/php.ini" -c "${WEB_NR_INI_DIR}/newrelic.ini" -n --ini &>/dev/null
+             sudo -u ${nonpriv_user} php -c "${WEB_PHP_INI_DIR}/php.ini" -c "${WEB_NR_INI_DIR}/newrelic.ini" -n --ini &>/dev/null
+             sudo -u ${nonpriv_user} php -c "${WEB_PHP_INI_DIR}/php.ini" -c "${WEB_NR_INI_DIR}/newrelic.ini" -n --ini &>/dev/null
```
joshuabenuck, line 195: In the FPM + Apache scenario where Apache isn't configured to use FPM, this may run these processes as the wrong user. Restarting both processes, as a way of avoiding the need to detect if FPM is really in use, fails in this scenario as we have to choose a user to use here which could introduce permissions problems if we choose incorrectly. Switching to a URL based approach would avoid this complexity at the cost of another user prompt.

recipes/newrelic/apm/php-agent/php.yml
```diff
@@ -110,12 +190,12 @@ install:
...
           else
           #
-          # No web server detected, this was a pure PHP CLI php agent installation.
+          # No web server detected, pure PHP CLI.
```
joshuabenuck, line 198: In this scenario, we also risk a permissions problem if we do not know which user runs the PHP CLI app.

recipes/newrelic/apm/php-agent/php.yml
zsistla: Currently the logic is detect fpm, apache, nginx in that order.
it will restart all that it detects.
The order really only matters in that only the first detected non-privileged user is used to run the php --ini at the end.

recipes/newrelic/apm/php-agent/php.yml
```diff
@@ -29,20 +29,80 @@ inputVars:
...
+    detect_services:
+      cmds:
+        - echo "Detect services."
+        - |
+          cd {{.TMP_INSTALL_DIR}}
+          rm -f {{.TMP_INSTALL_DIR}}/nonpriv_users.txt
+          rm -f {{.TMP_INSTALL_DIR}}/processes_to_restart.txt
+
+          #
+          # For guided installs, we currently support `php-fpm`, `apache`, and `nginx`.
```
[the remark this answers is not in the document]
zsistla: I can take it out, added it for completeness as it doesn't do anything if it's not detected, but when we start with nginx, we are ready for it.

recipes/newrelic/apm/php-agent/php.yml
zsistla: The warning will occur [here](https://github.com/newrelic/open-install-library/blob/4bb4b4e41336a0b50f4e96c0be908cae20a29d8c/recipes/newrelic/apm/php-agent/php.yml#L138).  It could also use a big warning like [this](https://github.com/newrelic/open-install-library/blob/e7347282f607e079f528d8199b678f29dbeb4fcb/recipes/newrelic/apm/dotNet/linux-systemd.yml#L58) before we restart (in non-silent mode).

Todos for the recipe that:
1) Silent mode needs to be incorporated like [this](https://github.com/newrelic/open-install-library/blob/e7347282f607e079f528d8199b678f29dbeb4fcb/recipes/newrelic/apm/dotNet/linux-systemd.yml#L71).  This can either be it's own ticket so it is consistent across the tasks, or I can work on it here as the blueprint to be used in other tasks.
2) Add error exit codes.  This can either be it's own ticket so it is consistent across the tasks, or I can work on it here as the blueprint to be used in other tasks.

recipes/newrelic/apm/php-agent/php.yml
zsistla: Few things about this:
When the processes are restarted, they restart the daemon so it is already running when this call is made so the non-privileged user used here does not affect the daemon.

I considered looping through the non-privileged users and doing the php call that way; however, the `newrelic-install.sh` script only installs `newrelic.ini` in one directory anyway.  If both fpm and apache exist on the system, the install script will only put the `newrelic.ini` file in one location.  So this is calling only the once with the priority non-privileged user because that is literally the only non-cli place our install will put the newrelic.ini file.

URL-based adds complexity and decreases reliability.  URL based approach is not just complexity of another user prompt (although everytime a user inputs data is a chance for wrong data to be entered) and the complexity of trying to guess what it should be in case of silent mode (which complexity increases once we move to dockerized support).  It is also the complexity of triggering a PHP page load.  What if they url they give/we guess doesn't have php to trigger the agent application?  Wordpress is notorious for not loading plugins with curl (they have their own functions to test).  I experimented a lot with this could not get the wordpress site to generate php traffic ( in order to trigger the initializing transaction )when using curl.  So to make it work reliably, we'd have to generate a test page (like one with phpinfo) and put it as a best guess in their webserver assuming we can locate it and have read access to it.

recipes/newrelic/apm/php-agent/php.yml
joshuabenuck: TIL. I didn't know we had any such detection logic in `newrelic-install.sh`. I don't think the reference needs to be removed.

recipes/newrelic/apm/php-agent/php.yml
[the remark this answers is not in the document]
joshuabenuck: If we are restarting all the processes we find, leaving it in may actually increase install success rates as this is the key (only?) thing needed to handle non-fpm nginx scenarios.

recipes/newrelic/apm/php-agent/php.yml
joshuabenuck: Separate tickets sounds like a better option to me.

recipes/newrelic/apm/php-agent/php.yml
zsistla: Possibly, but with the CLI this seems unlikely as they should be logged on as the user that will use the app(otherwise, if they aren't, there's no way for us to guess this).  If they were installing manually, they'd do these same steps.
They would download the agent, untar, run install, then run `php --info` (or another php cli command) from command prompt to verify.

recipes/newrelic/apm/php-agent/php.yml
joshuabenuck: If the process gets restarted as a privileged user and this uses a non-privileged user, don't we run the risk that the PHP cli process will be unable to communicate with the daemon?

recipes/newrelic/apm/php-agent/php.yml
zsistla: I also put `nginx` because I thought that was the focus of the current [MMF](https://github.com/newrelic/newrelic-php-agent/issues/93).  Is that incorrect?

```
For users of Debian or RHEL derived systems, running PHP under apache or fpm/nginx, the New Relic Open Installer library can be used to easily install and configure the PHP Agent on their systems.

This epic targets the initial buildout of supporting this, with a focus on FPM+nginx under Debian. Additional target environments will be supported in the subsequent epic.
```

recipes/newrelic/apm/php-agent/php.yml
```diff
@@ -27,16 +29,79 @@ inputVars:
...
+          # it's unlikely php5-fpm will be installed unless FPM is actually in use.
+          #
+          declare -a guided_support=("(fpm)" "(httpd|apache2|apache)" "(nginx)")
+
+          current_user=$(whoami)
+
+          for PROCESS in ${guided_support[@]}
+          do
+            nonpriv_user=$(ps -ef | egrep $PROCESS | grep -v grep | grep -v root | head -n1 | awk '{print $1}')
+            priv_user=$(ps -ef | egrep $PROCESS | grep -v grep | head -n1 | awk '{print $1}')
```
joshuabenuck, line 79: What guarantees that this entry will be a privileged user? Are there any privileged users other than root?

recipes/newrelic/apm/php-agent/php.yml
```diff
@@ -27,16 +29,79 @@ inputVars:
...
+              #
+              # Output all processes to be restarted to a file with the following format:
+              # process_name privileged_user
+              # When the process is restarted, it will be used with the associated
+              # privileged user.
+              #
+              echo "$process_name $priv_user" >> {{.TMP_INSTALL_DIR}}/processes_to_restart.txt
+              #
+              # Output all processes to be restarted to a file with the following forat:
+              # process_name privileged_user
```
joshuabenuck, line 95: Delete? Duplicate of lines 87-88.

recipes/newrelic/apm/php-agent/php.yml
```diff
@@ -27,16 +29,79 @@ inputVars:
...
+              # When initiating the transaction, it will be ONLY be used witt
```
joshuabenuck, line 96:
```suggestion
              # When initiating the transaction, it will ONLY be used with
```

recipes/newrelic/apm/php-agent/php.yml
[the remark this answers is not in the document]
zsistla: But wait, why don't we support nginx with php-fpm?
I thought that was the [MMF](https://github.com/newrelic/newrelic-php-agent/issues/93)?
Based on the text from the MMF, I thought nginx was the focus:
```
For users of Debian or RHEL derived systems, running PHP under apache or fpm/nginx, the New Relic Open Installer library can be used to easily install and configure the PHP Agent on their systems.

This epic targets the initial buildout of supporting this, with a focus on FPM+nginx under Debian. Additional target environments will be supported in the subsequent epic.
```
Is this incorrect?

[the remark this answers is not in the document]
zsistla: Either way, this recipe handles the combo of fpm/apache/nginx.

recipes/newrelic/apm/php-agent/php.yml
joshuabenuck: Aren't all recipes run as root?

recipes/newrelic/apm/php-agent/php.yml
zsistla: It is the user that starts the process.

Although I know for apache, it should always be root.
Apache is always started as "root" and it then uses "setuid" to spawn children that do the actually processing of requests for the apache user.

The service starts as root in order to do things like bind to reserved ports (e.g. 80 and 443). Then it starts whatever the configured number of processes are, to do the web-server work, and any other tasks, as the defined users. That way requests are being handled by non-privileged processes. You will notice the the parent ID (PPID) is the same for all of the other processes. That idea with be the PID for that one process running as root.

recipes/newrelic/apm/php-agent/php.yml
joshuabenuck: Recipes run as root by default. Gabe provided this snippet for how to run a command as the user who invoked sudo to run the recipe.

https://github.com/newrelic/open-install-library/blob/main/recipes/newrelic/apm/node/linux.yml#L49

recipes/newrelic/apm/php-agent/php.yml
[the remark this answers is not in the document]
joshuabenuck: We do support that case. When FPM is present, restarting it is sufficient to load the agent. nginx doesn't need to be restarted in that case. If there is a way to run a PHP site with nginx without FPM, restarting nginx is likely to be needed.

recipes/newrelic/apm/php-agent/php.yml
zsistla: Oh!  Thanks for pointing that out, I hadn't realized, but since it's run with obviously it would be.  Thanks for the link, I'll get that incorporated.

recipes/newrelic/apm/php-agent/php.yml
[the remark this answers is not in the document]
zsistla: In that case, I think we only need to restart the first process we detect.
We detect in this order: fpm, apache, nginx
If we detect fpm, we restart that and move on even if we also detect apache and/or nginx.
Otherwise, if fpm doesn't exist we will detect apache or nginx and restart that.

recipes/newrelic/apm/php-agent/php.yml
joshuabenuck: The started PHP process needs to be able to access the created socket. If the socket was created by the PHP process running as root, it might not be readable by a PHP process running as ubuntu (depending on the umask setting). I thought this was the core of the permissions issues caused by running PHP processes as different users.

recipes/newrelic/apm/php-agent/php.yml
joshuabenuck: Do parent processes always show up before child processes in the output of `ps`? (In my experience, they tend to, but I'm not sure it is a guarantee). If not, the user that started the parent process may not be the user identified by this pipeline.

recipes/newrelic/apm/php-agent/php.yml
joshuabenuck: The started PHP process needs to be able to access the created socket. If the socket was created by the PHP process running as root, it might not be readable by a PHP process running as ubuntu (depending on the umask setting). I thought this was the core of the permissions issues caused by running PHP processes as different users.

recipes/newrelic/apm/php-agent/php.yml
joshuabenuck: Does `ps` guarantee child process show up after parent processes? (They tend to, in my experience, I just don't know that is guaranteed). If not, you may end up with the less privileged user from a child process entry.

recipes/newrelic/apm/php-agent/php.yml
zsistla: I'm experimenting with this a bit, expect to have more details tomorrow.

recipes/newrelic/apm/php-agent/php.yml
zsistla: Good question.  Using `f` modifier in `ps -ef` I assumed would always list the parent first, but I've changed the flags to a way that guarantees the parent comes first.

recipes/newrelic/apm/php-agent/php.yml
zsistla: Permissions issues are not related to anything different we should do with the virtuoso install as they are more systemic and should be fixed at a higher level.

See related [issue](https://github.com/newrelic/newrelic-php-agent/issues/168) for details.

joshuabenuck (approved):

Fahmy-Mohammed (approved):

PR #361, opened 2021-05-19
Add apache fpm deploy config

joshuabenuck:
Since I neglected to open a PR for some of my recent changes, here is a summary of the approach being used.

Both the deploy config (json file) and the ansible play (yaml file) are kept as minimal as possible.

The deploy config's job is to create the EC2 instance of the needed OS and to specify the ansible play that needs to run to install the needed services (wordpress in this case).

The ansible play's job is to install the prerequisite OS packages, copy over the template files, and run a single shell script (install-wordpress.sh). The presence or absence of OS packages or specific binaries installed by those packages are used by the install script to determine which steps to perform on the target EC2 instance.

The install script is based on our wordpress setup script in our test-apps repo. It downloads wordpress using the wp-cli, sets up the site, and configures the web server based on which one has been installed.

This PR adds a new deploy config for an Apache FPM Wordpress setup. The install script has been updated to detect the presence of FPM in an Apache setup. In this scenario, an additional apache module is enabled and we use a slight variation of the previous wordpress site config that sends PHP files through a fast cgi proxy. (I was unable to use a single apache site config for both apache setups as it doesn't look like ansible templates allow setting arbitrary jinja template properties).

### DIFF
--- a/test/deploy/linux/php/templates/install-wordpress.sh
+++ b/test/deploy/linux/php/templates/install-wordpress.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 # Install wp-cli
 curl -O https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli.phar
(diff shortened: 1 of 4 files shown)


test/deploy/linux/php/templates/install-wordpress.sh
```diff
@@ -56,12 +56,20 @@ if [ $(which nginx) ]; then
 fi
 
 if [ $(which apache2) ]; then
-	if [ ! -f /etc/apache2/sites-available/wordpress-site ]; then
+	SITE="wordpress"
+	if [[ $(dpkg -l php-fpm) ]]; then
```
kneitinger, line 60: Is this intentionally Debian only?

The file name to me feels generically Linux, but `dpkg` is strictly deb right? Other than potentially this, the PR looks great though!

joshuabenuck: Good eye! The original intention was for it to be OS agnostic, but `php-fpm` doesn't get install anywhere on the path so running `which php-fpm` wasn't sufficient to determine if it is installed. `dpkg` is absolutely Debian specific. We will either need to do some OS detection here to specialize the command used or find a reliable OS agnostic way to detect whether `php-fpm` is installed.

I'm inclined to leave this as-is and address this once we formally add support for RedHat.

zsistla (approved):
Thanks for getting the environments up and stable.

PR #362, opened 2021-05-20
fix: use sudo when modifying the agent config

jlegoff:
Some customers receive an error because sed doesn't have permission to
modify the infra agent configuration. Add sudo in the command to fix it.


sanderblue (approved):

Julien4218:
@jlegoff a few tests failed, but it's not uncommon for the PR tests to fail when a lot of them are run at the same time.
I'll watch the next non-regression test job, which is more stable. If/when it passes I'll go ahead deploy this change

PR #365, opened 2021-05-21
Postgres improvements

carlossscastro:
Changes to Postgres recipe:

* Added connectivity test to postgres instance
* SSL inputs will only be requested if USE_SSL is true


Julien4218 (approved):
This is looking good @carlossscastro. I'll deploy shortly.

PR #366, opened 2021-05-21
Install gnupg in ubuntu and debian

jlegoff:
I noticed that on some versions of ubuntu, gpg is not installed, which results in an error when adding the gpg key for the infra agent.

We're currently installing gpg on some versions of debian. I modified the recipe to install only when it's not already installed, which seems more robust. Let me know what you think.

### DIFF
--- a/recipes/newrelic/infrastructure/debian.yml
+++ b/recipes/newrelic/infrastructure/debian.yml
@@ -34,7 +34,7 @@ install:
         - task: assert_pre_req
         - task: setup_license
         - task: update_apt
-        - task: add_gnupg2_if_required
+        - task: install_gnupg
         - task: add_gpg_key
         - task: add_nr_source
         - task: update_apt_nr_source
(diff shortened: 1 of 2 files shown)


recipes/newrelic/infrastructure/debian.yml
```diff
@@ -145,15 +145,15 @@ install:
       # apt will return an error if fails to update any of its sources. Ignore these errors and let the "install_infra" task fail.
       ignore_error: true
 
-    add_gnupg2_if_required:
+    install_gnupg:
       cmds:
         - |
-          if [ $(({{.DEBIAN_VERSION}})) -ge 10 ]; then
+          if [ {{.HAS_GPG}} -eq 0 ] ; then
```
sanderblue, line 151: Should we also keep the Debian version check? (`if [ $(({{.DEBIAN_VERSION}})) -ge 10 ]`) I'm a little unfamiliar with this specific task in the recipe, but at least wanted to ask 🙂

jlegoff: if gpg is not installed, it's going to fail whatever the version. I tried that on debian jessie, and it worked. To be honest, I'm not sure why we have this version check in the first place. Perhaps @Julien4218 knows

Julien4218: super

Julien4218 (approved):

Julien4218 (changes requested):
@jlegoff I just noticed some of the tests are failing, let s hold a bit until we find out why and confirm if it's really failing with this change

Julien4218 (approved):
nevermind it's some of the OHI tests, those can fail on PR test sometimes, approving.

PR #369, opened 2021-05-21
fix(ohi): remove redirect URL filter for due to avoid redirect to blank page

sanderblue:
Most entities seem to have a bit of a delay before they show up in the UI, so instead of redirecting users to a blank page by filtering to a specific type of entity, it's better to have them redirected to the main explore page so they at least see their Infra host entity at the very least.


gabeobrien (approved):
+1

PR #372, opened 2021-05-24
Redis recipe improvements

carlossscastro:
Changes to Redis recipe:

* Added connectivity test to Redis instance
* Password requested only if Redis authentication is set
* Decoupled infra agent presence testing from apache tests to fail before user input

### DIFF
--- a/test/deploy/linux/redis/install/debian/roles/configure/tasks/main.yml
+++ b/test/deploy/linux/redis/install/debian/roles/configure/tasks/main.yml
@@ -26,7 +26,7 @@
 
 - block:
   - name: Export PASSWORD
-    shell: "echo export NR_CLI_DB_PASSWORD=Virtuoso4all! >> ~/.bashrc"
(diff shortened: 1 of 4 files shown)


test/deploy/linux/redis/install/debian/roles/configure/tasks/main.yml
```diff
@@ -26,7 +26,7 @@
 
 - block:
   - name: Export PASSWORD
-    shell: "echo export NR_CLI_DB_PASSWORD=Virtuoso4all! >> ~/.bashrc"
+    shell: "echo export NR_CLI_PASSWORD=Virtuoso4all! >> ~/.bashrc"
```
Julien4218, line 29: @carlossscastro I don't want to bother with the semantic, I'll merge, but it might be good to be consistent with the other recipes since we use NR_CLI_DB_PASSWORD in other recipes. I agree it would be subjective to consider redis as a DB though.

Julien4218 (approved):
This redis enhancements looks good @carlossscastro, I'm going to merge and deploy.

PR #373, opened 2021-05-24
Check for PHP pre-reqs

joshuabenuck:
Note: This is based on Amber's PR (#359) and will need to be rebased once that is merged.

This PR adds an `assert_pre_req` step that checks the following:
* The recipe is being run via sudo.
* The target host has systemd installed.
* grep, egrep, sed, awk, and curl are installed.

The bulk of the code for this step was taken from the [.NET recipe](https://github.com/newrelic/open-install-library/blob/e7347282f607e079f528d8199b678f29dbeb4fcb/recipes/newrelic/apm/dotNet/linux-systemd.yml#L102).

### DIFF
--- a/recipes/newrelic/apm/php-agent/php.yml
+++ b/recipes/newrelic/apm/php-agent/php.yml
@@ -39,6 +39,7 @@ install:
   tasks:
     default:
       cmds:
+        - task: assert_pre_req
         - task: create_shared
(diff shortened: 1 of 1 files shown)


zsistla (approved):
Looks good!

Fahmy-Mohammed (approved):

PR #376, opened 2021-05-24
chore(infra): reduce validation SINCE to '5 minutes ago'

sanderblue:
Since we cut off recipe/resource validation (`validationNrql`) after 5 minutes, our validation NRQL should only query for data within a 5-minute window to ensure we're not querying extraneous events that can cause additional stress for larger accounts. This hopefully will help alleviate some of the error scenarios where customers have seen this error message: `Query blocked due to resource exhaustion by this query group.`

### DIFF
--- a/recipes/newrelic/infrastructure/amazonlinux.yml
+++ b/recipes/newrelic/infrastructure/amazonlinux.yml
@@ -18,7 +18,7 @@ keywords:
 
 processMatch: []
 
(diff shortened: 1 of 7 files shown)


jbeveland27 (dismissed):
I like these updates 🐑 🇮🇹 (after the tests pass)

Julien4218 (approved):

PR #378, opened 2021-05-25
Detect unsupported windows version and re-route the user to manual install

Julien4218:
https://newrelic.atlassian.net/browse/VIRTUOSO-959

### DIFF
--- a/recipes/newrelic/apm/dotNet/windows-iis-unsupported.yml
+++ b/recipes/newrelic/apm/dotNet/windows-iis-unsupported.yml
@@ -1,0 +1,50 @@
+# Visit our schema definition for additional information on this file format
+# https://github.com/newrelic/open-install-library/blob/main/docs/recipe-spec/recipe-spec.md#schema-definition
+
+name: dotnet-agent-installer
+displayName: .NET Agent Installer
+description: New Relic install recipe for .NET application on Windows
+repository: https://github.com/newrelic/newrelic-dotnet-agent
+
+installTargets:
+  - type: application
+    os: windows
+    platformVersion: "6.2.9200 Build 9200"
+  - type: application
+    os: windows
+    platformVersion: "6.3.9600 Build 9600"
+
+keywords:
+  - Apm
+  - .NET
+  - dotnet
+  - aspnet
+  - core
+
+processMatch:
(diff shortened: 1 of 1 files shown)


Julien4218:
Testing on a windows 2012 R2 server by running the recipe using Github URL (no service call to match with target host though).

[windows2012r2.txt](https://github.com/newrelic/open-install-library/files/6542720/windows2012r2.txt)

Julien4218:
@jaffinito I took a stab at implementing the recipe to specifically handle the Windows 2012 server.
I've fetched the specific value to look for from our event data. I also noticed we should cover 2012 R2 and 2012 as well.
Let me know your feedback.
We can do another round of testing with the service and target matching on staging once it's merged.

jaffinito (dismissed):
Versions match up to the 2012 family.
I do have one comment, "Installation of the .NET Agent is not available for this version of windows" reads like this version of Windows is not support at all for the .NET Agent, despite there being a link to more options.  Maybe "Installation of the .NET Agent is not available for this version of windows using this procedure"?

angelatan2:
This is probably a little late.  On the comment above stated by Jacob - I think an active voice is better.  "Please use standard install method for Window 2012 R2 version".  Point them to the link to get them there.

Julien4218:
> Versions match up to the 2012 family.
> I do have one comment, "Installation of the .NET Agent is not available for this version of windows" reads like this version of Windows is not support at all for the .NET Agent, despite there being a link to more options. Maybe "Installation of the .NET Agent is not available for this version of windows using this procedure"?

Good catch. I've updated the message. Thanks @jaffinito

jaffinito:
@angelatan2 I really like that wording!  Thoughts @Julien4218 ?

Julien4218:
> This is probably a little late. On the comment above stated by Jacob - I think an active voice is better. "Please use standard install method for Window 2012 R2 version". Point them to the link to get them there.

Yes @angelatan2 we are giving them the URL to follow for continuing the installation. The full message would look like this:

Installation of the .NET Agent is not available for this version of windows using this tool
Please see our other installation options:
https://docs.newrelic.com/docs/agents/net-agent/installation/install-net-agent-windows/.

PR #379, opened 2021-05-25
Update Node Recipe and UX

carlo-808:
* Enabled Node recipe to be included in Guided Installations
* Updated the node recipe to align the ux with .NET including verification step


Julien4218 (dismissed):
This looks good @carlo-808 
Side note, the failing tests have nothing to do with the PR, they're detected because of an issue in the git action noticing file differences with main (usually those are fixed if latest main is merged in the branch)

Julien4218 (approved):

PR #381, opened 2021-05-26
[Postgres] Fix test operator errors

carlossscastro:
Fix for test operator errors due to random mysql reponses to curl when inputting hostname and port

```
20:8: not a valid test operator: packets
20:8: not a valid test operator: of
24:10: not a valid test operator: packets
24:10: not a valid test operator: of
```


Julien4218 (approved):
Good find and fix @carlossscastro

PR #382, opened 2021-05-26
PHP RedHat deploy configs

joshuabenuck:
This PR creates PHP deploy configs for Amazon Linux 2 (a RedHat derivative). It mirrors what was done for Debian by creating a Wordpress install for Apache, Apache + FPM, and nginx + FPM.

### DIFF
--- a/test/deploy/linux/php/apache-fpm-wordpress/redhat/roles/configure/tasks/main.yml
+++ b/test/deploy/linux/php/apache-fpm-wordpress/redhat/roles/configure/tasks/main.yml
@@ -1,0 +1,28 @@
+---
+- debug:
+    msg: Install PHP
+
+- name: install PHP
+  shell: "amazon-linux-extras install -y lamp-mariadb10.2-php7.2 php7.2"
+  become: yes
+
+- name: install other pre-reqs
+  yum:
+    pkg:
+      - httpd
+      - mariadb
+      - mariadb-server
+    state: present
+  become: yes
+
+- template:
+    src: "{{ item }}"
+    dest: ~/templates/
+    mode: 0755
+  with_fileglob:
+    - ../../../../../templates/*
+  become: yes
+
+- shell:
+    cmd: ~/templates/install-wordpress.sh
+  become: yes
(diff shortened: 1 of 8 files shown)


zsistla:
Thanks for putting these up!

Was able to deploy :)

Questions:

1) when I ran `newrelic-install.sh`, it showed slightly different behavior from the previous jsons.
With install, the debian jsons only ask for the license key.

In `test/manual/definitions/apm/php-agent/php-nginx-fpm-wordpress-linux2.json`, the install additionally asked for:

```
New Relic PHP Agent Installation (interactive mode)
===================================================

   Enter New Relic license key (or leave blank): 07a2ad66c637a29c3982469a3fe8d1982d002c4a

Below is a list of the directories in which we found a copy of PHP.
Please select the directory or directories for which you wish to install
New Relic. You can select either a single directory or
multiple directories by separating each choice with either a space or a
comma. To select all of the directories shown, please enter the special
keyword 'all' (without the quotes).

   1)   /bin
   2)   /usr/bin

   0)   Exit

   Selection (1-2, 0 to exit or all):
```

Probably a valid use case that we'll need to ensure we handle for chatty and silent installs.

2) For `test/manual/definitions/apm/php-agent/php-nginx-fpm-wordpress-linux2.json`, the website is running, but ps and top aren't showing the processes?
```
ps aux
  PID TTY      STAT   TIME COMMAND
 5034 pts/0    Ss     0:00 -bash
 5080 pts/0    R+     0:00 ps aux
```

joshuabenuck:
@zsistla 
1. On this image `/bin` is a symlink to `/usr/bin`. I am not sure why. I agree that we'll want to make sure to handle this case (as well as situations where there are really two versions of PHP installed). Thanks for calling it out!
2. If you run ps as root you'll see the nginx processes. Perhaps there's an SELinux config or other security setting on Linux 2 instances that controls whether non-root users can see the processes for all users? Either way, since the recipes are run via sudo, I don't think we need to do anything different than we already are. Let me know if I'm missing anything.
```
sudo ps aux|grep nginx
root      4481  0.0  0.2 125884  2236 ?        Ss   14:00   0:00 nginx: master process /usr/sbin/nginx
nginx     4482  0.0  0.4 126340  4004 ?        S    14:00   0:00 nginx: worker process
nginx     4483  0.0  0.4 126340  4004 ?        S    14:00   0:00 nginx: worker process
ec2-user  4554  0.0  0.0 119416   868 pts/0    S+   14:01   0:00 grep --color=auto nginx
```

zsistla:
That makes sense, thanks @joshuabenuck.

zsistla (approved):

PR #384, opened 2021-05-27
feat: Nginx Check if user has stub_status enabled and change error codes

alvarocabanas:
- Add check to control if user has "with-http_stub_status_module" enabled, as it's required by NGINX in order to be monitored.
- If "with-http_stub_status_module" is detected but the url failed, inform the user that also the status endpoint should be set up and point him to our documentation in order to correctly set it up.


carlossscastro (approved):
LGTM

Julien4218 (approved):

PR #386, opened 2021-05-27
Improving Tomcat autodetection for CLI

meiao:
Instead of simply matching for `catalina`, will match for the whole class that starts Tomcat.
This can prevent mismatching in case they are running another Java process which has `catalina` as part of a parameter.


gabeobrien (approved):
+1

PR #387, opened 2021-05-31
feat: infrastructure agent status check after installation, just Linux

varas:
Once the infra agent gets installed, it'll request the status and report an error with appropriate error code and message in case something is not ok.

Description about the infrastructure agent status endpoint at: https://github.com/newrelic/infrastructure-agent/pull/475 (still not released!).

---
https://newrelic.atlassian.net/browse/VIRTUOSO-983

Solves https://github.com/newrelic/infrastructure-agent/issues/478


jlegoff:
Looks good to me, although it introduces an additional risk if the 8003 port is already taken. What do you think @Julien4218 ?

Julien4218 (changes requested):
I don't think we can assume port 8003 is free for us to use.
I assume we would have to document this, and handle the case when the port is not available on the host. Should we have logic to try another port, what do we do in the end when we can't find a port (fail the install, or the health check and skip over?)

Julien4218 (changes requested):
@varas we've talked about this work with @jlegoff 
In the short term, let's use a less likely to be used port number, a 5 digit one for instance
I've added a new MMF to take advantage of this work, so that we can fetch the entityGuid using the local endpoint in https://newrelic.atlassian.net/browse/VIRTUOSO-997

Julien4218 (approved):

PR #388, opened 2021-06-01
NGINX recipe Autodetect status url and NGINX PLUS

alvarocabanas:
- Autodetect status url if present and also check if NGINX PLUS or NGINX


sanderblue (approved):

Julien4218 (approved):
@alvarocabanas this looks like a good improvement, so I'll deploy it.
I believe we've talked about trying out the status URL we think is right, and if we see it works, using it without prompting the user.
Is this something you'll look into doing next?

PR #392, opened 2021-06-01
Dotnet linux preinstall

jaffinito:
Adds a preinstall script to detect if a recipe should be executed.  This is based on our existing introspector task in the recipe and uses the .NET introspector to gather a list of process that are running .NET Core  or .NET 5+.  All steps that can be sent to dev/null are to prevent any extra output.

Workflow:
1. Attempts to delete the folder we use to temporarily hold the introspector - `/tmp/dotnet-introspector` (same one from the introspector task)
2. Downloads the Linux x64 introspector zip to `/tmp`
3. Creates the `/tmp/dotnet-introspector` directory for the extracted introspector
4. Extracts the intospector using zcat (no dependencies)
5. Removes the zip file from `/tmp`
6. Uses chmod to set the `/tmp/dotnet-introspector` to 777 to allow the tool to run properly
7. Runs the instrospector, converts the json list of PIDs into a simple list, counts the number of lines, and puts that value in a variable
8. Checks if the value from step 7 not equal to 0, if it is then we found .NET process and exit with a 0.   If we did get back 0 then we did not find any running .NET process exit with a 3 - the error for for "No processes found to monitor".

### DIFF
--- a/recipes/newrelic/apm/dotNet/linux-systemd.yml
+++ b/recipes/newrelic/apm/dotNet/linux-systemd.yml
@@ -21,21 +21,19 @@ processMatch: []
 preInstall:
   requireAtDiscovery:
     - |
-      rm -rf /tmp/dotnet-introspector >/dev/null
-      curl -s https://virtuoso-testing.s3-us-west-2.amazonaws.com/dotnet-is-linux-x64.gz -o /tmp/dotnet-is-linux-x64.gz >/dev/null
-      mkdir -p /tmp/dotnet-introspector >/dev/null
-      zcat /tmp/dotnet-is-linux-x64.gz > /tmp/dotnet-introspector/nri-lsi-dotnet
-      rm -f /tmp/dotnet-is-linux-x64.gz >/dev/null
-      chmod -R 777 /tmp/dotnet-introspector >/dev/null
-      processCount=$(/tmp/dotnet-introspector/nri-lsi-dotnet -c | tr -d '[]' | tr ',' '\n' | wc -w)
+      TMP_DIR=$(mktemp -dq /tmp/newrelic.XXXXXX)
+      curl -s https://virtuoso-testing.s3-us-west-2.amazonaws.com/dotnet-is-linux-x64.gz -o ${TMP_DIR}/dotnet-is-linux-x64.gz >/dev/null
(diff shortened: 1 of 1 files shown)


Julien4218 (approved):
This is looking good @jaffinito

recipes/newrelic/apm/dotNet/linux-systemd.yml
```diff
@@ -18,6 +18,26 @@ keywords:
...
 
+preInstall:
+  requireAtDiscovery:
+    - |
+      rm -rf /tmp/dotnet-introspector >/dev/null
+      curl -s https://virtuoso-testing.s3-us-west-2.amazonaws.com/dotnet-is-linux-x64.gz -o /tmp/dotnet-is-linux-x64.gz >/dev/null
+      mkdir -p /tmp/dotnet-introspector >/dev/null
+      zcat /tmp/dotnet-is-linux-x64.gz > /tmp/dotnet-introspector/nri-lsi-dotnet
+      rm -f /tmp/dotnet-is-linux-x64.gz >/dev/null
+      chmod -R 777 /tmp/dotnet-introspector >/dev/null
```
gabeobrien, line 29: Can we not 777?  If you need to make sure it is executable maybe just a 'chmod u+x', that will give the current user execution permission for the file.

recipes/newrelic/apm/dotNet/linux-systemd.yml
```diff
@@ -18,6 +18,26 @@ keywords:
 
 processMatch: []
 
+preInstall:
+  requireAtDiscovery:
+    - |
+      rm -rf /tmp/dotnet-introspector >/dev/null
```
gabeobrien, line 24: One this we have been trying to get better at is store files in non deterministic location.  See the java recipe for an example:
https://github.com/newrelic/open-install-library/blob/main/recipes/newrelic/apm/java/linux.yml#L33

recipes/newrelic/apm/dotNet/linux-systemd.yml
jaffinito: Updated with the suggested change.

Julien4218:
@gabeobrien could you confirm if this PR is good to go? I'd like to get it merged and deployed, so it will be ready when ED ships.
Thank You

Julien4218 (approved):

PR #394, opened 2021-06-02
Add Debian+PHP FPM recipe

kneitinger:
Adds recipe (minus validation steps/proper service restart logic) for Debian systems using PHP FPM.

There are many benefits, both in this recipe as well as future ones, to using our tarball's `newrelic-install` command to execute the initial installation.  Most helpfully, the logs of that command have shown to be invaluable in extracting valuable information about aspects of the system, such as PHP versions, and INI conf.d directories. The references to `/tmp/nrinstall*`files is that log.

While this install method is the most platform agnostic and informative, we do still want to provide the user with a method to update their agent as we release, so after this config, the APT repository is added and the agent is semi-reinstalled.

After this, there are 2 New Relic ini files in the respective conf.d directories (`/etc/php/7.2/{fpm,cli}/conf.d/`), one called `newrelic.ini` created by `newrelic-install` and a `20-newrelic.ini` created by the Debian specific tooling. I feel that ultimately this recipe should `mv` the `newrelic.ini` files to the respective `20-newrelic.ini` files so that there is only one (and more specifically, the one `apt` will try to recreate if it isn't present on subsequent upgrades), but as it currently stands, the non `20-` is the one that is populated with the correct values, and it has a higher precedence than the `20-newrelic.ini` file so all is well.

This has only rudimentary service restart handling, and the logic to validate successful transactions has not been imported into this yet.

One other eventual modification is that the agent version is hardcoded (in a semi-convenient variable). The eventual approach we'll want prior to release is to parse the listing of `https://download.newrelic.com/php_agent/release/` and find the whole filename of the `-linux.tar.gz` archive (which currently cannot be statically determined).

### DIFF
--- a/recipes/newrelic/apm/php/debian-fpm.yml
+++ b/recipes/newrelic/apm/php/debian-fpm.yml
@@ -1,0 +1,192 @@
+# Visit our schema definition for additional information on this file format
+# https://github.com/newrelic/open-install-library/blob/main/docs/recipe-spec/recipe-spec.md#schema-definition
+
+name: php-agent-debian-fpm-installer
+displayName: PHP Agent Installer for Debian + FPM
+description: New Relic install recipe for instrumenting PHP applications on Debian systems with FPM
+repository: https://github.com/newrelic/newrelic-php-agent
+
+installTargets:
+  - type: application
+    os: linux
+    platform: "debian"
+  - type: application
+    os: linux
+    platform: "ubuntu"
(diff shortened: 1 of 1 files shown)


joshuabenuck (approved):

joshuabenuck:
Merged this as-is. We will split up the work to incorporate the relevant bits from the other recipe file as separate PRs.

PR #396, opened 2021-06-02
PHP: Merge changes from other recipe

joshuabenuck:
This PR merges most of the needed changes from the other recipe. I have not tested this yet. I'm putting it up for review so others can us it as a base in order to reduce merge conflicts.

### DIFF
--- a/recipes/newrelic/apm/php/debian-fpm.yml
+++ b/recipes/newrelic/apm/php/debian-fpm.yml
@@ -199,7 +199,7 @@ install:
         - |
           cd {{.TMP_INSTALL_DIR}}
           # Remove old log file to ensure we don't accidentally use it later
-          rm -f /tmp/nrinstall*.log
+          rm -f /tmp/nrinstall*
           AGENT="newrelic-php5-{{.AGENT_VERSION}}-linux"
           AGENT_TARBALL="$AGENT.tar.gz"
           curl -s https://download.newrelic.com/php_agent/release/$AGENT_TARBALL -o $AGENT_TARBALL
--- a/recipes/newrelic/apm/php/debian-fpm.yml
+++ b/recipes/newrelic/apm/php/debian-fpm.yml
@@ -115,14 +115,21 @@ install:
 
     create_shared:
       cmds:
-        - rm -rf {{.TMP_INSTALL_DIR}}
-        - mkdir -p {{.TMP_INSTALL_DIR}}
-        - echo "Created temporary directory {{.TMP_INSTALL_DIR}} for installation."
+        - |
+          WHITE='\033[0;97m'
+          GRAY='\033[0;240m'
+          NOCOLOR='\033[0m'
+          rm -rf {{.TMP_INSTALL_DIR}}
+          mkdir -p {{.TMP_INSTALL_DIR}}
+          echo -e "${WHITE}Created temporary directory {{.TMP_INSTALL_DIR}} for installation.${NOCOLOR}"
 
     detect_services:
       cmds:
-        - echo "Detect services."
         - |
+          WHITE='\033[0;97m'
+          GRAY='\033[0;240m'
+          NOCOLOR='\033[0m'
+          echo -e "${WHITE}Detect services.${NOCOLOR}"
           cd {{.TMP_INSTALL_DIR}}
           rm -f {{.TMP_INSTALL_DIR}}/nonpriv_users.txt
           rm -f {{.TMP_INSTALL_DIR}}/processes_to_restart.txt
@@ -195,8 +202,10 @@ install:
 
     install_tarball:
       cmds:
-        - echo "Downloading newest PHP Agent Release"
         - |
+          WHITE='\033[0;97m'
+          GRAY='\033[0;240m'
+          echo -e "${WHITE}Downloading newest PHP Agent Release${GRAY}"
           cd {{.TMP_INSTALL_DIR}}
           # Remove old log file to ensure we don't accidentally use it later
           rm -f /tmp/nrinstall*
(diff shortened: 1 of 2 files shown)


recipes/newrelic/apm/php/debian-fpm.yml
```diff
@@ -21,25 +21,40 @@ keywords:
...
     default:
       cmds:
         - task: verify_continue
         - task: assert_pre_req
+        - task: create_shared
+        - task: detect_services
         - task: add_gnupg2_curl_if_required
         - task: install_tarball
         - task: configure
         - task: start
```
zsistla, line 52: This task (and it's block [below](https://github.com/newrelic/open-install-library/blob/c600586e11ef2a2a12539a2cedb3756c266278d8/recipes/newrelic/apm/php/debian-fpm.yml#L261)) can be removed since the `restart services` task handles that.

zsistla: [Should](https://github.com/newrelic/open-install-library/blob/c600586e11ef2a2a12539a2cedb3756c266278d8/recipes/newrelic/apm/php/debian-fpm.yml#L201) navigate to `TMP_INSTALL_DIR` to install the tarball in.

zsistla: Since this section is expanding the tar logfile here, we don't need to do it [here](https://github.com/newrelic/open-install-library/pull/396/files#L299-L322), but the [logical](https://github.com/newrelic/open-install-library/pull/396/files#L299-L322) checks for the files should be moved up.  Additionally, since the first task already expanded the logfile, it can copy the expanded pertinent log file to a [global name](https://github.com/newrelic/open-install-library/pull/396/files#L39) that the second [task](https://github.com/newrelic/open-install-library/pull/396/files#L299-L322) can simply use.

zsistla: It might make sense to just do a `rm nrinstall*.log` before t[his line](https://github.com/newrelic/open-install-library/blob/c600586e11ef2a2a12539a2cedb3756c266278d8/recipes/newrelic/apm/php/debian-fpm.yml#L211) to ensure that [here](https://github.com/newrelic/open-install-library/blob/c600586e11ef2a2a12539a2cedb3756c266278d8/recipes/newrelic/apm/php/debian-fpm.yml#L230) we don't accidentally use an older file if one exists.

zsistla: [This](https://github.com/newrelic/open-install-library/blob/c600586e11ef2a2a12539a2cedb3756c266278d8/recipes/newrelic/apm/php/debian-fpm.yml#L218) was nicely done!! The modularization gives us so much more flexibility for adapting to different Linux flavors.

joshuabenuck: Changes made / commentary:
* `start` has been removed in favor of `restart_services`
* There's no way to change where the log tarball is stored so these references to `/tmp` remain.
* Everything else should be using `{{.TMP_INSTALL_DIR}}`
* It does not look like global vars can reference previously defined global vars so I had to define `$NR_INSTALL_LOG` as an env var at the top of each place it is needed. We now create a symlink to a static name which at least avoids the need to run any command to get the log file name.
* The log file is now extracted in `{{.TMP_INSTALL_DIR}}` so I added in the removal of all agent install artifacts instead.
* The log file expansion logic has been deduped and now lives completely within `configure`. This has the added benefit of failing the install earlier if the tarball agent install fails.

zsistla: I thought doing the following:
```
cd /{{.TMP_INSTALL_DIR}}
wget http://download.newrelic.com/php_agent/archive/9.17.1.301/newrelic-php5-9.17.1.301-linux.tar.gz
tar xzf newrelic-php5-9.17.1.301-linux.tar.gz
sudo NR_INSTALL_USE_CP_NOT_LN=1 ./newrelic-php5-9.17.1.301-linux/newrelic-install install
```
installed it in the calling directory.
Does it always install in `/tmp` instead?

recipes/newrelic/apm/php/debian-fpm.yml
```diff
@@ -116,6 +197,9 @@ install:
       cmds:
         - echo "Downloading newest PHP Agent Release"
         - |
+          cd {{.TMP_INSTALL_DIR}}
+          # Remove old log file to ensure we don't accidentally use it later
+          rm -f /tmp/nrinstall*
```
zsistla, line 202: This needs to have errors piped to null in case the file doesn't exist.

recipes/newrelic/apm/php/debian-fpm.yml
```diff
@@ -124,69 +208,150 @@ install:
...
         - rm -rf /tmp/nrinstall*
         - rm -rf newrelic-php5-*-linux
         - rm -rf newrelic-php5-*-linux.tar.gz
         - echo "Done cleaning up"
+
+    delete_shared:
+      ignore_error: true
+      label: "Cleaning up"
+      cmds:
+        - rm -rf {{.TMP_INSTALL_DIR}}
```
zsistla, line 356: Same comment for these `rm`s.  If we pipe to `2>/dev/null` then it won't error out on non-existent files.

recipes/newrelic/apm/php/debian-fpm.yml
joshuabenuck: The logfile is [always put](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/newrelic-install.sh#L88-L92) in `/tmp` regardless of the working directory used to launch the installer.

recipes/newrelic/apm/php/debian-fpm.yml
```diff
@@ -99,6 +113,80 @@ install:
...
+          rm -rf {{.TMP_INSTALL_DIR}}
+          mkdir -p {{.TMP_INSTALL_DIR}}
+          echo -e "${WHITE}Created temporary directory {{.TMP_INSTALL_DIR}} for installation.${NOCOLOR}"
+
+    detect_services:
+      cmds:
+        - |
+          WHITE='\033[0;97m'
+          GRAY='\033[38;5;240m'
+          NOCOLOR='\033[0m'
```
zsistla, line 131: A couple more `rm`s to pipe to null if the file is nonexistent.

zsistla (approved):
Looks good, thanks for merging these two files and for adding color!

PR #398, opened 2021-06-04
PHP: Fix prompt on first run, improve temp file clean up

joshuabenuck:
This PR does the following:
* Prevents apt from prompting the user for license key and application name the first time the recipe is run on a system. (Fix here is to pass `DEBIAN_FRONTEND` after the call to sudo instead of before).
* Merge clean up tasks and use a wildcard to delete the temp directory in order to clean up from previous failed install attempts. (This allows us to keep around files from failed installs to aid in troubleshooting).
* Directly use recipe variables where ever possible (instead of using them to set env vars). This is less complex and has the benefit of being resilient to changes in when and where the recipe runner chooses to use a clean shell environment.

### DIFF
--- a/recipes/newrelic/apm/php/debian-fpm.yml
+++ b/recipes/newrelic/apm/php/debian-fpm.yml
@@ -53,7 +53,6 @@ install:
         - task: restart_services
         - task: send_transaction
         - task: cleanup_temp_files
-        - task: delete_shared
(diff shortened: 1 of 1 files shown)


zsistla (approved):
This looks great! Approved pending anything coming up in the validation run.

PR #400, opened 2021-06-05
Dynamically determine most recent PHP agent version

kneitinger:
`curl`s the download site listing to determine the current most recent release of the PHP Agent at recipe run time.

In lieu of error checking the `curl` -> `grep` -> `head` pipeline, I opted for the simpler logic of checking the variable's emptiness.  Testing shows the same failure conditions as the following script:

```
#!/usr/bin/env bash

URL="https://download.newrelic.com/php_agent/release/"
WRONG_URL="https://download.newrelic.com/php_agent/releas/"

PATTERN='[1-9][0-9]\?\(\.[0-9]\+\)\{3\}'
WRONG_PATTERN='[1-9][A-Z]\+\(\.[0-9]\+\)\{3\}'


echo "Happy path"
if (set -o pipefail; curl -s "$URL" | grep --only-match "$PATTERN" | head -n1 ); then
    echo "pass"
else
    echo "fail"
fi

echo "Bad Curl"
if (set -o pipefail; curl -s "$WRONG_URL" | grep --only-match "$PATTERN" | head -n1); then
    echo "pass"
else
    echo "fail"
fi

echo "Bad pattern"
if ( set -o pipefail; curl -s "$URL" | grep --only-match "$WRONG_PATTERN" | head -n1 ); then
    echo "pass"
else
    echo "fail"
fi

echo "Bad everything"
if ( set -o pipefail; curl -s "$WRONG_URL" | grep --only-match "$WRONG_PATTERN" | head -n1 ); then
    echo "pass"
else
    echo "fail"
fi
```

```
$ ./version_handling.sh
Happy path
9.17.1.301
pass
Bad Curl
fail
Bad pattern
fail
Bad everything
fail
```

### DIFF
--- a/recipes/newrelic/apm/php/debian-fpm.yml
+++ b/recipes/newrelic/apm/php/debian-fpm.yml
@@ -204,16 +204,28 @@ install:
           cd {{.TMP_INSTALL_DIR}}
           # Remove old log tarballs to ensure we don't accidentally use it later
           rm -f /tmp/nrinstall* 2>/dev/null
-          AGENT="newrelic-php5-{{.AGENT_VERSION}}-linux"
+
+          # Versions of the form MAJ.MIN.PATCH.BUILD
+          VERSION_REGEX='[1-9][0-9]\?\(\.[0-9]\+\)\{3\}'
+          # PHP Agent current release tarball listing
+          RELEASE_URL='https://download.newrelic.com/php_agent/release/'
+          AGENT_VERSION="$(curl -s "$RELEASE_URL" | grep --only-match "$VERSION_REGEX" | head -n1)"
+          if [ -z "${AGENT_VERSION}" ]; then
+            echo -e "{{.RED}}Unable to determine current PHP Agent version from New Relic's downloads site.{{.NOCOLOR}}"
+            # Exit with Agent install failure code exit(16)
+            exit 16
+          else
+            echo "Found agent version: $AGENT_VERSION"
+          fi
+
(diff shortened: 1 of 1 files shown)


zsistla (approved):

recipes/newrelic/apm/php/debian-fpm.yml
```diff
@@ -204,16 +204,28 @@ install:
...
+          VERSION_REGEX='[1-9][0-9]\?\(\.[0-9]\+\)\{3\}'
+          # PHP Agent current release tarball listing
+          RELEASE_URL='https://download.newrelic.com/php_agent/release/'
+          AGENT_VERSION="$(curl -s "$RELEASE_URL" | grep --only-match "$VERSION_REGEX" | head -n1)"
+          if [ -z "${AGENT_VERSION}" ]; then
+            echo -e "{{.RED}}Unable to determine current PHP Agent version from New Relic's downloads site.{{.NOCOLOR}}"
+            # Exit with Agent install failure code exit(16)
+            exit 16
+          else
+            echo "Found agent version: $AGENT_VERSION"
```
joshuabenuck, line 218: Consider outputting this message in white and the agent version in cyan. Which version of the agent we are installing feels like something worth highlighting.

If you do implement this, you'll need to add this var:
```
CYAN: '\033[0;36`
```

joshuabenuck (approved):

Fahmy-Mohammed (approved):

zsistla (approved):

Fahmy-Mohammed (approved):

PR #412, opened 2021-06-12
Fix for when the recipe name contains a keyword we search for

zsistla:
The `fpm` grep was getting triggered by the process that called a recipe containing the keyword `fpm`.

Fixed this in two ways:
1) Narrowed the keyword we search for to `php-fpm`.
2) Added `grep` clause to ignore items containing `yml` so that if a future recipe name contains the keywords we are looking for, we won't grab the line containing the call to run that recipe.
This could have been an issue with apache as well if for instance there were a recipe named `debian-apache2.yml`.

### DIFF
--- a/recipes/newrelic/apm/php/debian-fpm.yml
+++ b/recipes/newrelic/apm/php/debian-fpm.yml
@@ -152,11 +152,11 @@ install:
           current_user=$SUDO_USER
           for PROCESS in ${guided_support[@]}
           do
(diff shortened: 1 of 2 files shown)


joshuabenuck (approved):
Great fix! I like the filter on `yml` (we probably don't even need to change the process name filter with it in place). I agree that this is likely not the last time this particular type of issue will be encountered.

Fahmy-Mohammed (approved):

PR #413, opened 2021-06-14
feat(o11y-packs): add observabilityPacks definitions to OHIs

jbeveland27:


kidk:
This looks great @jbeveland27 

I do have 1 concern. It looks like you're using the pack name, which might change quickly. I'm wondering if it's better to use the pack id or directory name?

jbeveland27:
Thanks @kidk. It's a good callout about the name vs an ID. The way this has worked in the recipes so far is we use `name` as an identifier, and `displayName` was meant to be for visual purposes. We could borrow that idea in the packs library as well, or go the identifier route, whatever makes sense. 

For now, the API is reliant on `name` as input to the `observabilityPackSearch` endpoint in NerdGraph. So we need to keep using that until we agree on what identifier should be used across the recipes/packs/service (@jpvajda also has a bullet point to discuss this during our weekly sync today).

jbeveland27:
Failures are unrelated to this change. Merging.

PR #414, opened 2021-06-15
Improvements on Java's Tomcat recipe

meiao:
This PR has improvements on the checking of the prerequisites and a small change in case catalina.sh does not have run permission.
But the bulk of it is support to some RHEL derivatives whose Tomcat packages are modified from the original version.


gabeobrien (approved):
+1

PR #415, opened 2021-06-15
feat: add infra agent status check for windows

jlegoff:


recipes/newrelic/infrastructure/windows.yml
```diff
@@ -122,6 +135,23 @@ install:
           # Ensure agent has enough time to start
           powershell -command 'Start-Sleep -s 10; $output = Get-Service "newrelic-infra"; if ( -not ($output -like "*newrelic-infra*") ) { Write-Error "The infrastructure agent has not started after installing. Please try again later, or see our documentation for installing manually https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/install-configure/install-new-relic."; Exit 31; }'
 
+    assert_agent_status_ok:
+      cmds:
+        - |
+          powershell -command '
+            Start-Sleep -s 5;
```
Julien4218, line 142: @jlegoff 
It looks like we already have a 10s sleep in the start_infra step, perhaps we can remove this extra sleep?

Julien4218 (approved):
@jlegoff I've removed the extra 5s sleep, since there was already a 10s before that to ensure the agent is up.
I've tested on a windows server, and it worked well, re-install too.
I think this is good to go, but holding until we've completed our research for linux.

varas (approved):

PR #416, opened 2021-06-15
Remove infra agent cache folders

Julien4218:
https://newrelic.atlassian.net/browse/VIRTUOSO-1018

### DIFF
--- a/recipes/newrelic/infrastructure/amazonlinux.yml
+++ b/recipes/newrelic/infrastructure/amazonlinux.yml
@@ -28,6 +28,7 @@ install:
     default:
       cmds:
         - task: assert_pre_req
+        - task: cleanup
         - task: setup_license
         - task: install_infra
         - task: restart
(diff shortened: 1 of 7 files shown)


Julien4218:
@jlegoff in my testing for windows, I notice the un-install removes all the files for the infra cache folder documented at https://github.com/newrelic/infrastructure-agent/blob/master/pkg/config/config.go#L285
"C:\Program Files\New Relic\newrelic-infra" (by the way the doc is missing a space between New and Relic, the real path has a space.

Given the un-install removes everything *except* the newrelic-infra.yml, which we manage through the recipe, I think we might want to not delete the folder because that would remove any client config change in the file beside the field we carefully manage in the recipe.

Please confirm, it's easy enough to remove the cleanup task for windows. Alternatively, I can remove all items except the "newrelic-infra.yml".
Side note, linux doesn't have this issue because the newrelic-infra.yml is in a different folder than /var/db/newrelic-infra

Julien4218:
@jlegoff following our conversations, I've removed the windows cleanup and will rely on the un-install feature to clean the directory.

jlegoff (approved):

PR #417, opened 2021-06-15
Fix python regex to show recommendet nerdlet

Julien4218:
Verified the regex is now working properly by running `newrelic install --trace`, there is no longer an error when evaluating python regex.


gabeobrien (approved):
+1

PR #418, opened 2021-06-16
feat: set infra up with proxy if provided

ctrombley:


recipes/newrelic/infrastructure/amazonlinux2.yml
```diff
@@ -111,6 +112,17 @@ install:
           sudo echo 'status_server_enabled: true' >> /etc/newrelic-infra.yml
           sudo echo 'status_server_port: 18003' >> /etc/newrelic-infra.yml
 
+    setup_proxy:
```
Julien4218, line 115: Look the way it's been done so far in https://github.com/newrelic/open-install-library/blob/main/recipes/newrelic/infrastructure/amazonlinux2.yml#L93-L112
Typically we remove any previous value, then add the new value.
One thing to consider is re-deploy. When redeploying we need to take out the value from the config, hence it's easier to remove it all the time and just append the new value if needed to be

recipes/newrelic/infrastructure/amazonlinux2.yml
```diff
@@ -111,6 +112,17 @@ install:
...
+      cmds:
+        - |
+          if [ ! -z "$HTTPS_PROXY" ]; then
+            if grep -q "proxy:" /etc/newrelic-infra.yml; then
+              sudo sed -i "s/^\(proxy: \).*/\1{{.HTTPS_PROXY}}/" /etc/newrelic-infra.yml
+            else
+              sudo echo 'proxy: {{.HTTPS_PROXY}}' >> /etc/newrelic-infra.yml
```
Julien4218, line 122: do we need HTTP_PROXY as well?

ctrombley: We do not

recipes/newrelic/infrastructure/amazonlinux2.yml
ctrombley: Explain this more?  Why do we need to take out the value from the config?  In which specific situation does this solution break down?

Julien4218: when the next run doesnt use proxy...

ctrombley: So installing without a proxy after already choosing to install with a proxy... I think I understand.

Julien4218 (commented):
@ctrombley this looks good. Feel free to do the other os/platforms

Julien4218 (approved):

PR #421, opened 2021-06-16
Retry infa validation with health endpoint

Julien4218:


Julien4218:
@jlegoff and @varas I've added this change to retry the validation with the infra healthpoint, hoping that perhaps a retry would eventually succeeds if the source of the issue relates to some network latency with our backend.
Let me know what you think.

jlegoff (approved):

PR #422, opened 2021-06-16
Java Tomcat - improving process detection

meiao:
Keeping both the processMatch and the requireAtDiscovery because the code for the latter can take over 20 seconds.


gabeobrien (approved):
+1

PR #427, opened 2021-06-21
mssql - add retries and more error cases

carlossscastro:
Improvements to mssql install script

* Added retries (max retries 3)
* Added extra checkings:
  * username/password are not valid/cannot login - exit code 3
  * username/password are empty - exit code 4
  * hostname is not valid - exit code 6 
  * hostname/port are not responding (timeout) - exit code 7
  * port is not a valid integer - exit code 8


Julien4218 (approved):
I've reviewed and tested and this looks good.
I'll merge and deploy shortly.
Thanks @carlossscastro

PR #429, opened 2021-06-23
Add a few checks for unsupported cases

brushknight:
- mark docker unsupported 131
- check apt-get is stable
- re-add license key in the config

### DIFF
--- a/recipes/newrelic/infrastructure/amazonlinux.yml
+++ b/recipes/newrelic/infrastructure/amazonlinux.yml
@@ -78,7 +78,7 @@ install:
           IS_DOCKER_CONTAINER=$(sudo grep 'docker\|lxc' /proc/1/cgroup | wc -l)
           if [ $IS_DOCKER_CONTAINER -gt 0 ] ; then
             echo "We’ve detected that you are installing our infrastructure agent inside a docker container. This agent is not designed to be installed within a container, but rather on the host it is running on or as a separate container. For additional information visit: https://docs.newrelic.com/docs/infrastructure/install-infrastructure-agent/linux-installation/docker-container-infrastructure-monitoring/" >&2
-            exit 18
+            exit 131
           fi
         - |
           IS_WSL_CONTAINER=$(sudo grep -i 'Microsoft' /proc/version | wc -l)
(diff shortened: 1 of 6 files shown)


recipes/newrelic/infrastructure/amazonlinux.yml
```diff
@@ -100,18 +100,16 @@ install:
...
             sudo touch /etc/newrelic-infra.yml
-            sudo cat > /etc/newrelic-infra.yml <<EOT
-          license_key: {{.NEW_RELIC_LICENSE_KEY}}
-          EOT
           fi
         - |
           if [ $(echo {{.NEW_RELIC_REGION}} | grep -i staging | wc -l) -gt 0 ]; then
             sudo echo 'staging: true' >> /etc/newrelic-infra.yml
           fi
+          sudo sed -i "1s/^/license_key: {{.NEW_RELIC_LICENSE_KEY}}\n/" /etc/newrelic-infra.yml
```
jlegoff, line 112: why do you use sed here instead of echoing the line?

brushknight: I just wanted to keep the same structure with license on top of config, it can be echo to the end as well

recipes/newrelic/infrastructure/debian.yml
```diff
@@ -114,6 +114,12 @@ install:
               exit 131
             fi
           fi
+        - |
+          APT_GET_STATUS=$(sudo bash -c 'apt-get update -yq > /dev/null 2>&1; echo $?')
+          if [ $APT_GET_STATUS -gt 0 ] ; then
+            echo "apt check exited with $APT_GET_STATUS code. $(sudo apt-get update -yq 2>&1 | grep "Err:")" >&2
```
jlegoff, line 120: it's bugging me that we're running apt-get update twice. It shouldn't be necessary.

brushknight: in fact it will be 3 times, as one exists in the actual install step. We need to do it somehow else probably

recipes/newrelic/infrastructure/amazonlinux.yml
Julien4218: Could we keep it consistent with the other echo below, it's a bit puzzling for a reader to understand why licenseKey is handled differently.

recipes/newrelic/infrastructure/debian.yml
Julien4218: @brushknight a change was deployed yesterday to ignore errors due to invalid package https://github.com/newrelic/open-install-library/blob/main/recipes/newrelic/infrastructure/debian.yml#L167
Could you mirror the same `-o Acquire::Check-Valid-Until=false` to skip those potential errors?

Julien4218:
@brushknight , or @cristianciutea , could you re-create this PR on the open-install-library repository, so that the end to end tests can run with the repo's secret and give us confidence we'll have no regression?

jlegoff:
this PR was recreated here: https://github.com/newrelic/open-install-library/pull/449

PR #430, opened 2021-06-23
PHP Virtuoso: Handle multiple PHP versions existing on the system.

zsistla:
1) Moved the `WEB_NR_INI_DIR` assignment down in case `WEB_NR_INI_DIR` gets reassigned.
2) Handle multiple `php` installations being detected.  In the case of multiple detections, `newrelic-install.sh` will default to `all` to install in all known locations.  This will result in multiple directory locations being recorded in the log file, so `sed` will pick up multiple instances.
3) The recipe sends php data via all known WEB php.ini/newrelic.ini combinations.  Note: because they all have the same app name, the post install logic will succeed if any of the php versions was able to successfully send data.
4) Modified the install logic which was only ever handling at most 1 php version.:w
5) The pure PHP CLI clause doesn’t loop through since it uses the default PHP that the user has set.
6) Changed the filenames to be less specific.

Note:  It runs on all redhat deploys and all but one debian deploy.  It will not run on test/manual/definitions/apm/php-agent/php-apache-wordpress-ubuntu18.json but neither will any other recipe.
Even the bare bones [recipe](https://github.com/newrelic/open-install-library/blob/df946260442f4ae0dd0afa526fe4b0cb6cb92a00/recipes/newrelic/php-agent/php.yml) will not run.


joshuabenuck:
In our RedHat configs, don't the /bin and /usr/bin php point to the same binaries via a symlink? Is there any harm in not de-duplicating the list? We'll run the installer twice against the same install and will end up sending two passes of data to the same install.

joshuabenuck:
`php-apache-wordpress-*` isn't a config we have to support this release. We need use enhanced process detection to correctly identify applicable systems so it is okay to leave it not working. (I was able to get it to work at one point... not sure what changed or if my test environment was corrupted).

recipes/newrelic/apm/php/redhat.yml
```diff
@@ -66,7 +66,7 @@ install:
           if [ "{{.NR_PHP_RESTART}}" = "y" ]; then
             echo -e "{{.YELLOW}}* Automatic process restart was selected.  This installation WILL restart the php-fpm or nginx process depending on the application architecture detected.{{.NOCOLOR}}"
           else
-            echo -e "{{.YELLOW}}* Manual process restart was selected.  This installation WILL NOT automatically restart the php-fpm, apache, or nginx process depending on the application architecture detected.  You will be prompted to restart manually in order to complete a successful installation of the PHP agent.{{.NOCOLOR}}"
+            echo -e "{{.YELLOW}}* Manual process restart was selected.  This installation WILL NOT automatically restart the php-fpm or nginx process depending on the application architecture detected.  You will be prompted to restart manually in order to complete a successful installation of the PHP agent.{{.NOCOLOR}}"
```
joshuabenuck, line 69: Was removing `apache` from the list intentional?

joshuabenuck: Looks like `apache` is also missing from the list in the debian recipe.

recipes/newrelic/apm/php/redhat.yml
```diff
@@ -376,7 +382,13 @@ install:
             #
             for i in {1..10}; do
               sleep 1
-              sudo -u $nonpriv_user php -c "${WEB_PHP_INI_DIR}/php.ini" -c "${WEB_NR_INI_DIR}/newrelic.ini" -n --ini &>/dev/null
+              #
+              #    Loop through all the detected PHP directories where newrelic.ini files
+              #    were placed.
+              #
+              for ((i=0; i<${#WEB_PHP_INI_DIR[@]}; i++)); do
+                sudo -u $nonpriv_user php -c "${WEB_PHP_INI_DIR[$i]}/php.ini" -c "${WEB_NR_INI_DIR[$i]}/newrelic.ini" -n --ini &>/dev/null
```
joshuabenuck, line 390: Doesn't each loop run the same PHP version / binary? If someone has PHP 5.6 and 8.0 installed on the system, wouldn't this run one of them with the ini file for the other version? That ini file will point to a different `newrelic.so` and fail to load. I think we need to somehow figure out which PHP binary to use with each set of ini files.

zsistla: It loops through all the detected php dirs that the install installed in.  When the `newrelic-install.sh` is in silent mode, it installs automatically in all detected PHP directories.  So the loop goes through all the specific PHP directories that have their own distinct `php.ini` and `newrelic.ini` files and doesn't intermix the different versions  of inis.  

Yes, that's a good point about the binary.  I'm pretty sure the log file has this info, so we should be able to use it to pull the correct PHP binary.

recipes/newrelic/apm/php/redhat.yml
zsistla: Not sure how it got removed.  I'll readd it.

zsistla:
> In our RedHat configs, don't the /bin and /usr/bin php point to the same binaries via a symlink? Is there any harm in not de-duplicating the list? We'll run the installer twice against the same install and will end up sending two passes of data to the same install.

The recipe doesn't run the installer twice.  `newrelic-install.sh` detects it as two different versions so in silent mode, it automatically does what it does to install in all the directories it detects.  Once that exits and back in the recipe the `configure` task will loop through and add the name and collector as needed.  After it modified the ini file the first time, the  appname/collector `sed` doesn't match anymore so it doesn't do anything.

In the `send_transaction` task, it will loop through duplicates as if they are unique.  I'd vote for not de-duplicating since there's no harm in this, leaving it doesn't risk breaking anything, this particular case might not be common, the logic to de-duplicate would introduce more complexity, and the data sent is minimal.

Do you have a strong opinion one way or the other on this?

joshuabenuck:
> In the `send_transaction` task, it will loop through duplicates as if they are unique. I'd vote for not de-duplicating since there's no harm in this, leaving it doesn't risk breaking anything, this particular case might not be common, the logic to de-duplicate would introduce more complexity, and the data sent is minimal.
> But I could add some de-duplicating logic. Do you have a strong opinion one way or the other on this?

I agree with you that leaving the duplication is the better option.

recipes/newrelic/apm/php/debian.yml
```diff
@@ -400,7 +400,24 @@ install:
               #    were placed.
               #
               for ((i=0; i<${#WEB_PHP_INI_DIR[@]}; i++)); do
-                sudo -u $nonpriv_user php -c "${WEB_PHP_INI_DIR[$i]}/php.ini" -c "${WEB_NR_INI_DIR[$i]}/newrelic.ini" -n --ini &>/dev/null
+                #
+                # Get the PHP binary associated with a specific PHP/newrelic ini file.
+                #
+                #
+                # When there are multiple PHPs detected, the log file creates blocks
+                # of information witha specific format.
```
joshuabenuck, line 408:
```suggestion
                # of information with a specific format.
```

recipes/newrelic/apm/php/redhat.yml
```diff
@@ -387,7 +387,24 @@ install:
               #    were placed.
               #
               for ((i=0; i<${#WEB_PHP_INI_DIR[@]}; i++)); do
-                sudo -u $nonpriv_user php -c "${WEB_PHP_INI_DIR[$i]}/php.ini" -c "${WEB_NR_INI_DIR[$i]}/newrelic.ini" -n --ini &>/dev/null
+                #
+                # Get the PHP binary associated with a specific PHP/newrelic ini file.
+                #
+                #
+                # When there are multiple PHPs detected, the log file creates blocks
+                # of information witha specific format.
```
joshuabenuck, line 395:
```suggestion
                # of information with a specific format.
```

joshuabenuck (approved):

Fahmy-Mohammed (approved):

PR #433, opened 2021-06-25
Migrate windows from processMatch to preinstall scripts and fixes some other issues

jaffinito:
Migrates both windows-iis-unsupported.yml and windows-iis.yml to use a preInstall script.
Uses the existing assert_pre_req task script to perform the check.
Removed the assert_pre_req task entirely since we don't need to check for IIS and we want to install even if ASP.NET is not installed to support .NET Core/.NET 5+

### DIFF
--- a/recipes/newrelic/apm/dotNet/windows-iis-unsupported.yml
+++ b/recipes/newrelic/apm/dotNet/windows-iis-unsupported.yml
@@ -21,8 +21,20 @@ keywords:
   - aspnet
   - core
 
-processMatch:
-  - w3wp.exe
+processMatch: []
+
+preInstall:
+  requireAtDiscovery: |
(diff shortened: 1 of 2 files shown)


Julien4218 (commented):
@jaffinito this looks like a nice improvement.
Question, do we also need to check for the aspNet45Enabled in the script pre-check? It seems we wouldn't want to try to install without that?

jaffinito:
Since .NET Core doesn’t need the ASP.NET feature, if we just install based on IIS we get Core as well as Framework.   This will increase the number of successful installs.

Julien4218 (approved):

Julien4218:
@jaffinito this looks good, I'll get it deployed shortly.

PR #434, opened 2021-06-28
fix/infra: don't remove the entire db directory

jlegoff:
Removing /var/db/newrelic-infra also removes the installed integrations
and makes logging fails on reinstall. Only removing the data subdirectory
should be enough.


rubenruizdegauna (approved):
LGTM

Julien4218 (approved):

PR #443, opened 2021-06-30
feat: added windows permissions check

cristianciutea:
This PR will check if the correct permissions are provided.


Julien4218:
@cristianciutea can you re-create this PR from the open-install-library repo. We need the PR to be from a branch on this repo, so the end-to-end tests can run and give us feedback that we don't have any regression.

PR #445, opened 2021-06-30
Revert "fix: disable the status check in the infra recipe"

jlegoff:
Reverts newrelic/open-install-library#440 - the problem should be fixed in the new release of the agent


Julien4218 (approved):
@jlegoff changes look good, re-running the test to make sure windows is still working with the latest infra (failure could have been a fluke).

PR #454, opened 2021-07-01
PHP: fix dup inis and other changes

joshuabenuck:
This PR makes the following changes:
* Add logic to move the ini files created by the tarball installer to the location that will be used by the package installer.
* Display the detected agent version in gray instead of red. It isn't an error condition and we display the version we are using in brighter colors a little later on so it is a detail that the user can safely ignore in most cases.
* Set the color to gray when detecting services. This reduces the attention drawn to any shell output. (Aside: I'm seeing a usage message from grep displayed at this step. Not sure why, but it does not appear to be related to my changes).

**Design considerations:**
There's a question about what to do if the target ini directory doesn't exist. We could fail the install right there, but since there may be more than one PHP install, there's a chance another will succeed where this fails.

The approach chosen is to just warn if the target destination does not exist. This essentially results in the same behavior as before these changes were made. It results in multiple ini files which may cause problems later down the road.

The alternative is to fail the first time we are unable to find a target ini dir, but this may leave the system in a partially installed state with some ini files in the correct location and some in the SAPI specific directories. We could write clean up logic to attempt to remove them all, but that adds additional complexity to the recipe.

Should the ini file move succeed, there is a chance that the module does not get enabled. The changes purposefully do not manually call `phpenmod` to enable the module as there is logic in the installer to do that. This avoids the complication of having to determine which version of `phpenmod`, if any, needs to be called and also reduces the duplication between the recipe and the agent installers.

Note that the recipe has been failing more often than not for me today, but it doesn't seem to be related to my changes as it seems to pass if I run it enough. I also noticed that the recipe installer no longer seems to check the connection to New Relic before installing the recipe so I'm not sure if the problem is on the NR side or not.

### DIFF
--- a/recipes/newrelic/apm/php/debian.yml
+++ b/recipes/newrelic/apm/php/debian.yml
@@ -40,7 +40,7 @@ install:
       sh: mktemp -d /tmp/newrelic-php-agent.XXXXXX
     WHITE: '\033[0;97m'
     RED: '\033[0;31m'
-    GRAY: '\033[38;5;248m'
+    GRAY: '\033[38;5;240m'
     CYAN: '\033[0;36m'
     NOCOLOR: '\033[0m'
     YELLOW: '\033[0;33m'
@@ -135,7 +135,7 @@ install:
     detect_services:
       cmds:
         - |
-          echo -e "{{.ARROW}}Detecting services{{.NOCOLOR}}"
+          echo -e "{{.ARROW}}Detecting services{{.GRAY}}"
           cd {{.TMP_INSTALL_DIR}}
           rm -f {{.TMP_INSTALL_DIR}}/nonpriv_users.txt 2>/dev/null
           rm -f {{.TMP_INSTALL_DIR}}/processes_to_restart.txt 2>/dev/null
@@ -226,7 +226,7 @@ install:
           # for treating directory as an HTML directory listing
           RELEASE_URL='https://download.newrelic.com/php_agent/release/'
           AGENT_VERSION="$(curl -s "$RELEASE_URL" | grep --only-match "$VERSION_REGEX" | head -n1)"
-          echo -e "{{.RED}}AGENT VERSION: $AGENT_VERSION{{.NOCOLOR}}"
+          echo -e "{{.GRAY}}AGENT VERSION: $AGENT_VERSION{{.NOCOLOR}}"
           if [ -z "${AGENT_VERSION}" ]; then
             echo -e "{{.RED}}Unable to determine current PHP Agent version from New Relic's downloads site.{{.NOCOLOR}}"
             # Exit with Agent install failure code exit(16)
@@ -305,6 +305,19 @@ install:
             if [ "{{.NEW_RELIC_REGION}}" = "STAGING" ]; then
               sed -i 's/;newrelic.daemon.collector_host = ""/newrelic.daemon.collector_host = "staging-collector.newrelic.com"/' $ini_full_name
             fi
+            target_ini_dir="/etc/php5/conf.d/"
+            if [ -x /usr/sbin/phpquery ]; then
+              target_ini_dir="${ini}/../../mods-available/"
+            elif [ -x /usr/sbin/php5endmod ]; then
+              target_ini_dir="/etc/php5/mods-available/"
+            fi
+            if [ ! -d $target_ini_dir ]; then
+              echo -e "{{.RED}}Warning: Unable to move ${ini_full_name} to ${target_ini_dir}{{.GRAY}}"
+              continue
+            fi
+            # Move newrelic.ini created by tarball to the name that the installer will use.
+            mv ${ini_full_name} ${target_ini_dir}
+            # We rely on the package installer to enable the newrelic module.
           done;
 
     restart_services:


recipes/newrelic/apm/php/debian.yml
```diff
@@ -305,6 +305,19 @@ install:
...
+              target_ini_dir="${ini}/../../mods-available/"
+            elif [ -x /usr/sbin/php5endmod ]; then
+              target_ini_dir="/etc/php5/mods-available/"
+            fi
+            if [ ! -d $target_ini_dir ]; then
+              echo -e "{{.RED}}Warning: Unable to move ${ini_full_name} to ${target_ini_dir}{{.GRAY}}"
+              continue
+            fi
+            # Move newrelic.ini created by tarball to the name that the installer will use.
+            mv ${ini_full_name} ${target_ini_dir}
```
zsistla, line 319: This is why you will see the verification logic fail.  We use the ini_full_name as to link the php dir, the ini dir, and the bin dir https://github.com/newrelic/open-install-library/pull/454/files#diff-c69192e44c92e64181623439ca11082a8eed8f52672cd8ad803f4932789b0364L410-L420.  

Once you have moved it, it won't exist for us to use. 

I think if that logic were moved up to this location and those three values for php dir, target ini dir, and bin dir for each detected php were output to a text file, we could just use the text file in send_transactions instead of scraping the logfile again.

joshuabenuck: Great catch! That explains the behavior I was seeing with the inconsistent recipe installation success.... If we are going to fix the multiple ini file issue, we won't be able to scrape the ini file location from the log file after this step. If we were to output it to a text file (or shell script that we could source), we'd need to update the values after this move.

I cannot shake the feeling that we are layering complexity on top of complexity here. It makes me think we need to revisit some of our decisions.

I know we are running the tarball installer because it produces an easily parseable log that lets us know which ini files to use when sending transaction. We are running the package installer afterwards to make it so the installation is easily upgradeable by the user (and because the packages don't have a nice log file to parse).

These two seem to be at the core of the complexities we are fighting here. If the packages gave us machine parseable results, we could just install them. If the newrelic-install.sh script put the ini files in the same place as the packages, we wouldn't need to mess with moving ini files around.

We have tried up until this point to release the first pass of Virtuoso recipes without requiring any agent changes. I think both this issue and the permissions issue point to the need to do otherwise. I think we should fix the inconsistencies between newrelic-install.sh and the packages. Make it so they do the same things and then switch to only using the packages in the recipes so we only have to run an install / configuration process once in these recipes.

Additionally, given the current focus of the Virtuoso team, I think we should defer fixing either of these issues until after the next bug fix release of the agent (where we can hopefully include enough of the needed changes to make the next attempt at resolving these issues straightforward and maintainable).

Agree? Disagree? Thoughts?

zsistla: Using the tarball and apt-get together is a fairly innocuous combination.  I've used them frequently together in test apps and haven't noticed the issues. 

Even with the duplicate ini issue, the recipes still work because of the INI order of precedence.  The ini issue and the permissions issue are a bit subtle. 

After chasing the permissions issue, I've started to suspect the permissions issue is somewhat related to both the installs being called from within a sudo-called script or possibly related to both scripts `chmod`ing the same directory since the issues are slightly different that what I observed by calling the install scripts manually.

I think your ini changes look good, and it'd be fairly straightforward to finish off the scraping and outputting relevant file paths at the same time and output the info to a file for later use. It would reduce the current complexity by putting the scraping in one spot instead of scraping the same information multiple times.  I'm familiar with that area of the recipe and have a good idea how to do that portion if you want me to work on it.

My concerns with waiting for the fix in the install files:
1) No guarantee this will be fixed in the bug fixes as changing our install scripts in a way that would require a significant verification effort and could push our Virtuoso effort out to an unknown date.  We don't know how many bugs/how much time will be allocated.  Assuming it goes into the bug fix effort, this also makes all of our Virtuoso work dependent on the next release of the agent which currently doesn't have an estimated date.  The recipe wouldn't work end-to-end until after the agent was released and up on the download site meaning our recipe dev efforts would be hobbled a bit and the virtuoso team's efforts would be completely stymied as they wouldn't be able to test/verify our recipe until the agent was released.
2) Would require additional effort for both recipes which would then diverge even more instead of having a larger common recipe code base which is more easily maintainable.
3) With our main Virtuoso contact leaving Friday, if we put Virtuoso into pending, it might require more effort on both teams' parts for the 2nd goaround.  It seems the most pragmatic choice to deliver working recipes ASAP before either team moves on to other tasks.
4) Realistically, there is no clear picture of how/when Virtuoso would fit in the roadmap and would require some spin up time with possible new team members.
5) Even without the ini/permissions fixes, the recipes themselves still install/send data successfully; it is the debug aspect that they currently impact.  There is a way to resolve the ini `mv` impact on verification issue (which also simplifies the current recipes), and the permission issue seems close to  being resolved.  

Given the above, I'd advocate to finish the current incarnation.

joshuabenuck: The issue with multiple ini files will not be experienced until the package makes a change to an ini setting that gets applied on an upgrade. The package will update the ini file that has a lower precedence, but if the setting modified is one that has a value in ini file with higher precedence, it have any effect (and may take a while to track down if it is raised in a support case).

I'm definitely not as familiar with that portion of the script. If you want to work on it, I have no objections.

Comments on your concerns:
1. We have the ability to influence what is included and the time allocated for the bug fix release. If we feel this is the only maintainable way to create the PHP recipes, we shouldn't shy away from advocating for it. The next agent bug fix release is currently scheduled to start right after we wrap up the recipe work. The Virtuoso team has already been pulled off to work on other priorities. Even if we do have this completed tomorrow, it has no estimated release date (see the Q&A in the latest instrumentation all hands). The Virtuoso team is working on reducing Infra install error rates and then they will be reassessing their roadmap. The window to get these recipes released soon has already closed. (We could, however, get them merged to main soon by wrapping up the current approach, but with no clear release date, I have trouble seeing the value).
2. If we defer completion of the PHP recipes, I think the goal should be to make whatever changes are needed to unify the two install paths and minimize any divergence. It will definitely require additional effort on both recipes, but the result would (hopefully) be more maintainable and simpler.
3. Even should the recipes be merged tomorrow, we have no ETA on when they will be released. That is currently unknown and unscheduled.
4. I completely agree that there is no clear picture on when Virtuoso will fit into the roadmap after this. Whether we finish out the current approach or revisit the approach after the next bug fix release, we do not know when these recipes will be released.
5. The debug aspect severely limits the supportability of the resulting installs and would never go away on its own without either running a version of the recipe that can patch up the problems, manual intervention, or adding something to the packages. This could follow us for quite some time should it get used by a large number of users.

If there was still a window to release these recipes and get customer feedback in short order, I'd be more inclined to say that we should finish the current implementation.

How much does the timeline of the recipe release to customers impact whether you feel we should finish out the current approach?

zsistla: My proposal is really to do both.

I feel we should finish the current incarnation.
If we address these two issues, the recipe should be in a stable state.
I'll move the scraping logic and add the permissions fix by end of Friday or Tuesday (since Monday is off).  At that point, we have working recipes, and can neatly close off that work.

During the next bug fix MMF, we can address how the two installs have slowly been diverging and synch them up again with all the best parts retained as well as add any tweaks that could be useful for the recipes. 

After the agent release we can optimize/restructure as needed for the recipes. Having working recipes then gives less pressure on any subsequent restructure/optimization work since we already have a working recipe in reserve that can be used.

I feel it makes sense to take the additional day or two now to get to at least stable, working recipes that can be merged (and that we can have in reserve in case the schedule becomes problematic later on).  Then we can schedule a time to restructure with the updated/streamlined install scripts.

zsistla:
```
The issue with multiple ini files will not be experienced until the package makes a change to an ini setting that gets applied on an upgrade. The package will update the ini file that has a lower precedence, but if the setting modified is one that has a value in ini file with higher precedence, it have any effect (and may take a while to track down if it is raised in a support case).
```
Regarding ini files, if the installs detect a newrelic.ini file exists, they don't modify it or change it in any way.  However, the logic can be improved since `newrelic-install.sh` only looks for the existence of `newrelic.ini` instead of `*newrelic.ini`; if that were fixed, I think the duplicate ini files wouldn't exist anymore and we'd be able to remove that logic.  I believe the packages handle this more gracefully.

PR #458, opened 2021-07-07
PHP Virtuoso: Permissions and Validation fixes

zsistla:
1) calling Debian installer package followed by `newrelic-install` is the more common use case.
2) Call `newrelic-install` with the `NR_INSTALL_USE_CP_NOT_LN` [flag](https://docs.newrelic.com/docs/agents/php-agent/advanced-installation/silent-mode-install-script-advanced/); otherwise the installer was creating a symbolic link that was then copied into a directory referenced by another installer link which was causing issues like:
	* seemingly permissions issues since the log files weren’t being written correctly
     * the `php-cli` could never work due to the issue: `PHP Warning: PHP Startup: Unable to load dynamic library ‘newrelic.so’ `
3) While the web services correctly logged debug messages to agent/daemon logs, if they started first, cli processes wouldn’t write to the logs. Change permissions to allow the cli process to modify these files.
4) Moved the log scraping into a file so we don’t have to keep scraping it for Debian.
5) If we are in staging, enabled debug log files by default.
6) Add `clean_slate` to terminate any running daemons and remove the log directory to reduce the chance of permissions issues.

### DIFF
--- a/recipes/newrelic/apm/php/debian.yml
+++ b/recipes/newrelic/apm/php/debian.yml
@@ -54,9 +54,10 @@ install:
         - task: create_shared
         - task: detect_services
         - task: add_gnupg2_curl_if_required
+        - task: install_deb
         - task: install_tarball
         - task: configure
-        - task: install_deb
+        #- task: install_deb
         - task: restart_services
         - task: send_transaction
         - task: cleanup_temp_files
@@ -253,7 +254,10 @@ install:
           echo 'deb http://apt.newrelic.com/debian/ newrelic non-free' | sudo tee /etc/apt/sources.list.d/newrelic.list
           curl -s https://download.newrelic.com/548C16BF.gpg | sudo apt-key add -
           sudo apt-get update
-          sudo DEBIAN_FRONTEND=noninteractive apt-get install newrelic-php5 -y -qq
+          sudo DEBIAN_FRONTEND=noninteractive apt-get install newrelic-php5 newrelic-php5-common newrelic-daemon -y -qq
+          #sudo DEBIAN_FRONTEND=noninteractive apt-get install newrelic-php5 -y -qq
+          sudo killall newrelic-daemon
+          rm -rf /var/log/newrelic
 
     configure:
       cmds:
@@ -304,6 +308,9 @@ install:
             sed -i "s/newrelic.appname = \"PHP Application\"/newrelic.appname = \"{{.NR_PHP_APPLICATION}}\"/" $ini_full_name
             if [ "{{.NEW_RELIC_REGION}}" = "STAGING" ]; then
               sed -i 's/;newrelic.daemon.collector_host = ""/newrelic.daemon.collector_host = "staging-collector.newrelic.com"/' $ini_full_name
+              sed -i 's/;newrelic.loglevel = "info"/newrelic.loglevel = "verbosedebug"/' $ini_full_name
+              sed -i 's/;newrelic.daemon.loglevel = "info"/newrelic.daemon.loglevel = "debug"/' $ini_full_name
+                            #sed -i 's/;newrelic.daemon.loglevel = "info"/newrelic.daemon.loglevel = "verbosedebug"/' $ini_full_name
             fi
             target_ini_dir="/etc/php5/conf.d/"
             if [ -x /usr/sbin/phpquery ]; then
--- a/recipes/newrelic/apm/php/debian.yml
+++ b/recipes/newrelic/apm/php/debian.yml
@@ -52,12 +52,12 @@ install:
         - task: verify_continue
         - task: assert_pre_req
         - task: create_shared
+        - task: clean_slate
         - task: detect_services
         - task: add_gnupg2_curl_if_required
         - task: install_deb
         - task: install_tarball
         - task: configure
-        #- task: install_deb
         - task: restart_services
         - task: send_transaction
         - task: cleanup_temp_files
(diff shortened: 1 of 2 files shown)


recipes/newrelic/apm/php/debian.yml
```diff
@@ -296,30 +309,85 @@ install:
...
+              #
+              # We rely on the package installer to enable the newrelic module.
+              #
+
+              #
+              # Three important pieces of information: the php binary location, the php ini
+              # file location, the actual (not symlink) New Relic ini file location.
+              # If we have web directories, copy the three key bits of info to the web file;
+              # otherwise copy to the cli file.  We will save the php directory, the target
+              # directory where the ini file is, and the php bin directory.
```
joshuabenuck, line 380: The either / or nature of this check feels a bit off to me. According to [this comment](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/newrelic-install.sh#L928-L940) in `newrelic-install.sh`, when there are separate PHPs for web and cli, they often (but not always) point to a separate scan directory, but end up sharing their config by storing them in a common directory and then symlinking to them.

The debian package follows this convention by detecting a common conf directory, putting the ini file there, and then calling `phpenmod` to create the symlink to each of the sapi directories present on the system.

Favoring the web version of PHP over the cli version when sending sample transactions makes sense. If that is all we are doing here, then a comment update is all that is needed. If, however, we really are only putting the ini file in the web or the cli conf directory, then I think we need to change the recipe to match the installer.

recipes/newrelic/apm/php/debian.yml
```diff
@@ -296,30 +309,85 @@ install:
...
+              #
+              if [ -z "${CLI_DIRS}" ]; then
```
joshuabenuck, line 382: I only see this being set after this if clause and it is only referenced in these two places. Is there something else that sets or references this not in the recipe?

zsistla: The first set of dirs the loop goes through is web, after it is done with that, the next set of dirs will be CLI so it is marked.

recipes/newrelic/apm/php/debian.yml
```diff
@@ -372,21 +440,8 @@ install:
...
-          # Fallback to cli ini dir if there is no web specific ini dir
-          if [ -z "${WEB_PHP_INI_DIR}" ]; then
-            WEB_PHP_INI_DIR=($(sed -n 's/.*final pi_inidir_cli=\(.*\)\/conf.d/\1/p' $NR_INSTALL_LOG))
-            WEB_NR_INI_DIR=($(sed -n 's/.*final pi_inidir_cli=\(.*\/\)/\1/p' $NR_INSTALL_LOG))
-            if [ -z "${WEB_PHP_INI_DIR}" ]; then
-              WEB_PHP_INI_DIR=$WEB_NR_INI_DIR
-            fi
-          fi
 
+          sudo -u $SUDO_USER sudo chmod 666 /var/log/newrelic/*
```
joshuabenuck, line 444: This makes me uncomfortable. A malicious user on the system could use this to fill the log volume before performing an attack they don't want to be logged in the system's logs.

Alternatives to consider are (in no particular order):
1. Configure newrelic.ini to use a different filename to avoid the need for a world writable file.
2. `chown` the file to the primary user we expect to be writing to the log file (potentially emitting a warning about the others being unable to write to it)
3. Detect the incompatibility and fail the installation with a unique error code. If this ends up being common, address it then.
4. Double check with the security team and ship this as-is, if they are okay with the risk.

recipes/newrelic/apm/php/debian.yml
zsistla: The comment needs to be clarified.

recipes/newrelic/apm/php/debian.yml
joshuabenuck: Could there be multiple web ini dirs if there are multiple versions of PHP installed on the system? If so, do we risk setting the cli flag too soon?

recipes/newrelic/apm/php/debian.yml
zsistla: Those are valid concerns.  The change here is only to allow php-cli to send debug msg to the logs when web services are installed; in the case of php-cli only installation, it is not an issue, so not a reason to fail the recipe. 

Some options:
1) If we installed for web, make a note that for the php-cli to send log messages the user would need to change the permissions.  There would be no need for us to change any permissions as all the web users are able to write to the file.
2) investigate behavior of chowning to sudo_user.
3) Double check with the security team and ship this as-is, if they are okay with the risk.

joshuabenuck: I like option one. I'd also be okay with not logging that message until this scenario proves to be problematic. Logging a message for a scenario that ends up not being an issue in practice would result in more confusion than help.

recipes/newrelic/apm/php/debian.yml
zsistla: All the web dirs are bundled together in an array, all the cli dirs are bundled together in an array.
After we are done with the web dirs array, we set the flag and move on to handling the cli dirs array.

recipes/newrelic/apm/php/debian.yml
zsistla: I'm okay also not logging the message.
I investigate #2, but I'll just remove the chmod now.

recipes/newrelic/apm/php/debian.yml
joshuabenuck: I see that now. Thank you for so patiently pointing out the parts of this code I missed!

joshuabenuck (approved):

Fahmy-Mohammed (approved):

PR #469, opened 2021-07-22
Fix the dotnet agent install validate step

Julien4218:
Resolves newrelic/newrelic-dotnet-agent#658


jaffinito (approved):
That looks good to me.  Double quotes should work better.

PR #470, opened 2021-07-22
Retry starting IIS if first attempt fails in .net windows iis recipe

JcolemanNR:
There have been error reports of IIS failing to start during the guided install process. We used to get similar reports for stopping IIS, but were able to resolve most of them by adding retry logic. This PR adds similar retry logic to the starting of IIS.

This should resolve: https://github.com/newrelic/newrelic-dotnet-agent/issues/656

Tested on Server 2019 and did not see any failures. The behavior of IIS failing to start is intermittent so is difficult to test for explicitly.

### DIFF
--- a/recipes/newrelic/apm/dotNet/windows-iis.yml
+++ b/recipes/newrelic/apm/dotNet/windows-iis.yml
@@ -120,11 +120,12 @@ install:
             catch { return $false }
           }
           
-          $DidIisStop = (Stop-IIS)       
-          if (-not $DidIisStop) {
-              # Try to stop IIS a second time.
-              # Based on past experience, the second attempt almost always works.
-              Stop-IIS | Out-Null
+          for ($tries = 0; $tries -lt 30; $tries++) {
+            if ((Stop-IIS)) {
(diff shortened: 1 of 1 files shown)


Julien4218 (dismissed):
@JcolemanNR this is a very good idea.
You might consider retrying a few times, maybe just 3 at most for this, but also put a little delay to give time to the OS to resolve any potential contentions.
Here is an example when we start the infra agent, to wait and check a few times.
https://github.com/newrelic/open-install-library/blob/main/recipes/newrelic/infrastructure/windows.yml#L209-L233

jaffinito (dismissed):
Nice!

JcolemanNR:
@Julien4218 Thanks for the feedback. I changed the IIS stop/start to retry 30 times with 2 second intervals. I don't think there is any harm in this... it should drastically reduce the chance of failures for customers unless something is TOTALLY wrong

JcolemanNR:
Tested and working on Server 2019

Julien4218 (approved):

PR #471, opened 2021-07-22
Disable support for Server 2008/2008 R2 in guided install

JcolemanNR:
Resolves: https://github.com/newrelic/newrelic-dotnet-agent/issues/657

There have been customer complaints/questions about whether we support guided install on Server 2008/2008 R2. We already have checks to flag 2012/2012 R2 as unsupported, but we didn't have similar rules for 2008/2008 R2.

It would be great if there was a way to programmatically set a minimum supported version and reject everything under that, but I am unfamiliar with if that is supported in our recipe YMLs. If there is an easy way to do this then let me know.

I am spinning up some server 2008 instances to test and will only merge if everything works properly.

I also added some helpful messages when IIS is not installed.

### DIFF
--- a/recipes/newrelic/apm/dotNet/windows-iis-unsupported.yml
+++ b/recipes/newrelic/apm/dotNet/windows-iis-unsupported.yml
@@ -29,6 +29,8 @@ preInstall:
     $iisEnabled = Get-WindowsOptionalFeature -Online -FeatureName IIS-WebServer | select State
     if ($iisEnabled.State -ne "Enabled") {
       # IIS is not installed
+      Write-Host "Required dependency (IIS) is not installed. To enable run:"
+      Write-Host "Enable-WindowsOptionalFeature -Online -FeatureName IIS-WebServer"
       exit 3
     }
 
--- a/recipes/newrelic/apm/dotNet/windows-iis-unsupported.yml
+++ b/recipes/newrelic/apm/dotNet/windows-iis-unsupported.yml
@@ -39,8 +39,6 @@ preInstall:
     $iisEnabled = Get-WindowsOptionalFeature -Online -FeatureName IIS-WebServer | select State
     if ($iisEnabled.State -ne "Enabled") {
       # IIS is not installed
-      Write-Host "Required dependency (IIS) is not installed. To enable, run:"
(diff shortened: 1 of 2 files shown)


recipes/newrelic/apm/dotNet/windows-iis.yml
```diff
@@ -25,6 +25,8 @@ preInstall:
     $iisEnabled = Get-WindowsOptionalFeature -Online -FeatureName IIS-WebServer | select State
     if ($iisEnabled.State -ne "Enabled") {
       # IIS is not installed
+      Write-Host "Required dependency (IIS) is not installed. To enable, run:"
```
jaffinito, line 28: I don't know if customers can actually see this message.
@Julien4218 Can they see this?

recipes/newrelic/apm/dotNet/windows-iis-unsupported.yml
```diff
@@ -7,13 +7,23 @@ description: New Relic install recipe for .NET application on Windows
 repository: https://github.com/newrelic/newrelic-dotnet-agent
 
 installTargets:
+  # Server 2008
```
jaffinito, line 10: Awesome!  I love the comments, makes it much more readable.

jaffinito (approved):
Other than my question, this looks great!

recipes/newrelic/apm/dotNet/windows-iis.yml
Julien4218: @jaffinito no this message would not be seen by anyone. Script in the "preInstall" are only used for evaluating whether the recipe should be run, so it's a yes (0 exit code) or false (non 0 exit code) answer.
Any output is discarded.

Julien4218 (dismissed):

recipes/newrelic/apm/dotNet/windows-iis.yml
JcolemanNR: I'll pull these logging changes out

Julien4218 (approved):

Julien4218:
@JcolemanNR I'm merging this. It will get deployed Monday with other changes.

JcolemanNR:
Sounds good! Thanks!

PR #472, opened 2021-07-22
tests: validate output for successful runs

ctrombley:


Julien4218 (approved):
Yep this is good @ctrombley 
For the single recipe like the infra ones, we could potentially also latch on the "Installing infrastructure-agent-installer...success"
There shouldn't be a difference, I'm not sure which one is better.
Feel free to update more tests.

Julien4218 (approved):
looks great, thanks @ctrombley

PR #473, opened 2021-07-27
feat(infrastructure): new `validation` schema, deprecate old `validationNrql` schema

sanderblue:


noahmmcgivern (approved):

docs/recipe-spec/recipe-spec.md
```diff
@@ -118,10 +118,17 @@ inputVars: list, optional
...
 
+# DEPRECATED! Use `validation` parameters instead.
 # NRQL the newrelic-cli will use to validate the agent/integration this recipe
 # installed is successfully sending data to New Relic
 validationNrql: string, optional
 
+# Validation options for an agent/integration to use to validate the recipe
+# was installed successfully and is sending data to New Relic
+validation: object, optional
+  agentUrl: string, optional # the localhost URL the newrelic-cli will use to validate an agent is successfully sending data to New Relic (currently only applies to the infra agent)
```
Julien4218, line 129: can we drop the terminology of "agent", this would make it more future proof in case down the road we're validating on any endpoint type.
cc @noahmmcgivern  @sanderblue

sanderblue: How does `validationUrl` work for you? We're open to that and other suggestions 🙂

Julien4218: I would like `validationUrl` if it wasn't under a `validation` name already. It would be nice to find something consistent with the other child/sibling for `nrql`

sanderblue: Yep, that's what we're saying. We'll make it a top-level sibling

e.g.
```yaml
validationNrql: "SELECT * FROM thing"

validationUrl: "http://imaurl.com"
```

Julien4218: ok

Julien4218 (approved):

PR #480, opened 2021-08-05
fix(redis): use TCP instead of unix sockets

ctrombley:


Julien4218:
this looks good to me, let's ship it.
cc @ctrombley

Julien4218 (approved):

PR #486, opened 2021-08-20
feat: publish artifacts to S3

ctrombley:


Julien4218:
I've updated the value for AWS_DEFAULT_REGION to mirror the one used when publishing with the newrelic-cli (east 1).

PR #496, opened 2021-08-31
php prereq error messages changed to red for easier visibility

ZNeumann:


ZNeumann:
https://github.com/newrelic/newrelic-php-agent/issues/202

zsistla (approved):

recipes/newrelic/apm/php/debian.yml
```diff
@@ -101,14 +101,14 @@ install:
...
+            echo -e "{{.RED}}This installation recipe for the New Relic PHP agent must be run under sudo.{{.NOCOLOR}}" >> /dev/stderr
             exit 7
           fi
         - |
           IS_SYSTEMCTL_INSTALLED=$(which systemctl | wc -l)
           if [ "$IS_SYSTEMCTL_INSTALLED" -eq 0 ]
           then
-            echo "This installation recipe for the New Relic PHP agent only supports services managed by 'systemd'." >> /dev/stderr
+            echo -e "{{.RED}}This installation recipe for the New Relic PHP agent only supports services managed by 'systemd'.{{.NOCOLOR}}" >> /dev/stderr
             exit 20
```
Julien4218, line 112: @ZNeumann could you update this exit code to 131 (unsupported). This is not really an error, it's just we currently don't support non-systemd installations through this mechanism. You might also mirror this change for the RHEL recipe.

zsistla: Hi @julien4218, 

I'm not sure I understand the recommendation of changing from error code 20 to 131.
```
20 |   | For linux, systemctl is not available on the host (either nor present or not running)
-- | -- | --
```

It seems using the error code specifically for that dependency would help us to help customers debug better vs the generic 131 and is similar to what [dotnet](https://github.com/newrelic/open-install-library/blob/main/recipes/newrelic/apm/dotNet/linux-systemd.yml#L114) and [java](https://github.com/newrelic/open-install-library/blob/main/recipes/newrelic/apm/java/linux.yml#L93) are doing in providing specific codes that highlight the dependency needed by the recipe to install.

Julien4218: Hello @zsistla , yes that documentation would need to be updated too.
Actually even .net, and other APM, we're now shifting the direction to handle all the unsupported cases through the NR1 UI, leading the user to our documentation which supports all the various scenarios.
The Guided install would only be used for what we can automate with high confidence. We could track a failure exit code, however this is more of an unsupported case here, and ideally we should identify this in the preInstall script before the recipe even gets to the install point.
For the preInstall, the execution result is binary, it's either good to execute the recipe (0 exit code), or not ok to execute the recipe (non 0 exit code).
Here is an extensive example for java https://github.com/newrelic/open-install-library/blob/main/recipes/newrelic/apm/java/linux.yml#L21-L45

zsistla: Got it, thanks for the example and info about the changing paradigm!
Should we move all of our pre-reqs to a similar ```preInstall:
  requireAtDiscovery:```
block?

Julien4218: @zsistla yes, it makes sense to move all the pre-req within the preInstall check, for the same reason.

mfulb: @Julien4218 We currently test if the script is being run as a sudo user - can we include a test like that in the requireAtDiscovery?  I'm curious about the context the preInstall commands are run under compared to when the recipe is executed.

Julien4218: @mfulb it should be the same, the cli doesn't do things differently when running preInstall versus executing the recipe, beside just looking for a 0/non-0 exit code and been non-interactive.

mfulb: @Julien4218 I did some quick tests with the preInstall and it seems to be working.  However when it returns non-zero the newrelic cli exits with a generic message about not being able to install the PHP agent and suggesting a support page for help.
Can we do anything to return a string which would give the user an idea why the script could not run due to a test in requiredAtDiscovery failing?  Currently we have a 'assert_pre_reqs' task that does the same sorts of tests and will echo the cause of the error.  I think it would improve the user experience to have this kind of feedback on a failure if possible.

Julien4218: Good thinking @mfulb .
To answer your question, no there isn't a way to do anything in the preInstall beside returning an exit code.
With that said, I agree with you and I think perhaps for this specific error (user not running sudo), it might make sense to have this implemented in the recipe, provide a meaningful message (run as sudo) and exit the recipe.
I would suggest using the exit code 131 (unsupported), since we don't support installing without sudo priviledges.

mfulb: Thanks @Julien4218  - just want to clarify.  When you mean "implemented in the recipe" do you mean as a task in the install section like we have it now (as done in this existing PR), or in the preInstall section where we output a message as well as emit an error code?  

BTW I tried using echo in the preInstall:requiredAtDiscovery section and the text is output as expected but I was running the newrelic cli in a shell - I'm still a little unclear what a user doing the guided install would see in this situation.

Julien4218: @mfulb "implemented in the recipe", I meant to say implemented in the "install/tasks" section, not the preInstall script part.

Yes, echo in the preInstall would print out, however we don't want to do that because those echos could be run on some host that don't have any PHP running if somehow the processMatch filter is removed.
As a general rule, preInstall are only meant to be non-interactive, to evaluate if the recipe should execute. This gives us more control to identify the proper recipes beyond process matching.

mfulb: @Julien4218 Thank you for the clarification.  

If I follow correctly then the checks in requireAtDiscovery should be restricted to a non-interactive check to see if PHP exists in a state which can be instrumented by the agent if installed, and then the other checks related to if the environment is correct for running the recipe (run with elevated permissions, grep is installed, etc) would be best placed in the the "install/tasks" section.

zsistla: I think all those checks should go in the requireAtDiscovery.  That way we only run the recipe if there is a high confidence of it succeeding.  If it doesn't succeed, we can redirect them to a page that says x,y,z is required, please check these dependencies before trying again because sometimes they may be missing more than one dependency so it's good to redirect them to the list of requirements.  We want to separate system dependency related issues from recipe installation related issues.

mfulb: I think this PR can proceed - the work on adding a preInstall section will be discussed further elsewhere.

PR #498, opened 2021-09-01
Auto create SQL user required for MSSQL OHI

Julien4218:


Julien4218:
Hello @carlossscastro I've continued the OHI improvement to auto-create the required SQL user.
This is the PR for Microsoft SqlServer. When you have time, if you could review it, and potentially test it too, that would be great.
As far as I can see on an instance I've spawned, I can connect to a default instance using the port, and a named instance (without the port, since it's different).

Julien4218:
Hello @carlossscastro, I believe I'm done with this work.
The recipe change is now creating the SQL user automatically for the user, with a strongly random generated password.
The recipe is idempotent, so re-running it would essentially re-install with a new user credential cleaning any previous user with that same name.
I've tested on a host while creating a second named instance in addition of the default one, and I can see both instances reporting to NR after installing.
Let me know your feedback.
https://staging.onenr.io/04ERP5qdGjW

carlossscastro (approved):
Looks very good.
We will need to adjust the format of the config file it creates to adhere to v4 configuration.
Let me know if you need help with that

PR #502, opened 2021-09-07
Remove the NGINX Plus condition since we do support it on the guided install.

alvarocabanas:
- Remove the NGINX Plus condition since we do support it on the guided install.

### DIFF
--- a/recipes/newrelic/infrastructure/ohi/nginx/debian.yml
+++ b/recipes/newrelic/infrastructure/ohi/nginx/debian.yml
@@ -54,6 +54,10 @@ preInstall:
   requireAtDiscovery: |
     # extract NGINX status url
     stub_location=$(sudo nginx -T 2>&1 | grep -ozP "(?s:.*\s)\Klocation(?s).*stub_status" | grep -aoP "\/([^\s]+)")
+    if [ "$stub_location" == "" ]; then
+      exit 1
+    endif
+
     code=$(curl -I "http://127.0.0.1:80${stub_location}" -w %{response_code} -so '/dev/null')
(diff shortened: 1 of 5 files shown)


recipes/newrelic/infrastructure/ohi/nginx/debian.yml
```diff
@@ -52,18 +52,13 @@ preInstall:
...
-    # way harder to detect in NGINX PLUS
-    if [ $(nginx -V 2>&1 | grep -o nginx-plus | wc -l) -eq 0 ] ; then
-      # extract NGINX status url
-      stub_location=$(sudo nginx -T 2>&1 | grep -ozP "(?s:.*\s)\Klocation(?s).*stub_status" | grep -aoP "\/([^\s]+)")
-      code=$(curl -I "http://127.0.0.1:80${stub_location}" -w %{response_code} -so '/dev/null')
-      if [ $code == "200" ]; then
-        exit 0
-      fi
+    # extract NGINX status url
+    stub_location=$(sudo nginx -T 2>&1 | grep -ozP "(?s:.*\s)\Klocation(?s).*stub_status" | grep -aoP "\/([^\s]+)")
```
Julien4218, line 56: @alvarocabanas should we check the stub_location variable is not empty so that we don't do a curl that would timeout?

alvarocabanas: oks, I can add that chec

alvarocabanas: done

Julien4218 (approved):

PR #503, opened 2021-09-07
guarantee sources.list.d exists before using it

ZNeumann:
https://github.com/newrelic/newrelic-php-agent/issues/213

Install fails if /etc/apt/sources.list.d/ does not exist. This should exist by default if the user has apt installed. So it could not exist on a debian system for 2 reasons:

1) The user removed the directory.
This is the case addressed by this PR. The user might have wished to remove 3rd party pkg repos at some point and wiped the whole directory.

2) The system doesn't have apt.
I don't know of any situations where a debian system wouldn't have apt installed. Is this a check that should be added to prereqs?

### DIFF
--- a/recipes/newrelic/apm/php/debian.yml
+++ b/recipes/newrelic/apm/php/debian.yml
@@ -274,6 +274,10 @@ install:
       cmds:
         - |
           echo -e "{{.ARROW}}Installing New Relic PHP Agent package{{.GRAY}}"
+          if [ ! -d "/etc/apt/sources.list.d/" ]; then
+            sudo mkdir -p /etc/apt/sources.list.d
+            sudo chmod 755 /etc/apt/sources.list.d
+          fi
           echo 'deb http://apt.newrelic.com/debian/ newrelic non-free' | sudo tee /etc/apt/sources.list.d/newrelic.list
           curl -s https://download.newrelic.com/548C16BF.gpg | sudo apt-key add -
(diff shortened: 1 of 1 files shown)


mfulb:
This looks reasonable.  The directory you are creating is intended to hold files for 3rd party repos so it seems OK to create it if missing.   I guess the question would be if the recipe encounters something one would reasonably expect to be present on a platform if we should try to mend it or just report it as an error and have the system administrator look into it further.

zsistla:
@mfulb brings up an interesting point about mending the system which we should probably avoid doing.

You mentioned it happens for two reasons:
```
The user removed the directory.
This is the case addressed by this PR. The user might have wished to remove 3rd party pkg repos at some point and wiped the who directory.

The system doesn't have apt.
I don't know of any situations where a debian system wouldn't have apt installed. Is this a check that should be added to prereqs?
```
In case 1, if the user has purposely removed it, I'm not sure we should put it back (especially in a recipe where they might not realize it is back). 
in case 2, we should definitely check if apt-get (and yum for redhat) is installed in the prereqs.

Would checking for the existence of the file (and exiting with 131 unsupported) in prereqs also verify if apt-get is installed? That might get both edge cases in one fix.

mfulb:
I tend to agree about case 1.
Case 2 fits into the PR I'm putting together with preInstall so how about I add that to what I'm doing?

Julien4218 (approved):

PR #505, opened 2021-09-09
change check for sources.list.d to pre_req

ZNeumann:
an addendum to https://github.com/newrelic/open-install-library/pull/503

### DIFF
--- a/recipes/newrelic/apm/php/debian.yml
+++ b/recipes/newrelic/apm/php/debian.yml
@@ -105,6 +105,11 @@ install:
             exit 7
           fi
         - |
+          if [ ! -d "/etc/apt/sources.list.d/" ]; then
+            echo -e "{{.RED}}Ensure /etc/apt/sources.list.d/ exists.{{.NOCOLOR}}" >> /dev/stderr
+            exit 131
+          fi
+        - |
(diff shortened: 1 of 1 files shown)


mfulb (approved):

recipes/newrelic/apm/php/debian.yml
```diff
@@ -104,6 +104,11 @@ install:
             echo -e "{{.RED}}This installation recipe for the New Relic PHP agent must be run under sudo.{{.NOCOLOR}}" >> /dev/stderr
             exit 7
           fi
+        - |
+          if [ ! -d "/etc/apt/sources.list.d/" ]; then
+            echo -e "{{.RED}}Ensure /etc/apt/sources.list.d/ exists.{{.NOCOLOR}}" >> /dev/stderr
```
zsistla, line 109: This error would also occur if `apt-get` isn't installed, right?  I think we need to ensure the `apt-get is installed` check happens prior to this.

zsistla (approved):

mfulb:
I will add checks for the appropriate package manager utilities in the PR I'm working on.

mfulb:
After some discussion it was decided I will integrate these changes into a PR I am about to submit.
We can close this PR without merging once that PR has gone through.

PR #507, opened 2021-09-13
Php agent preinstall

mfulb:
This PR implements the preInstall:requireAtDiscovery method for the PHP agent recipes for RHEL and "debian-like" Linux distributions.  It adds numerous tests in the preInstall to verify the environment has a high probability of supporting the PHP agent installation.  Any test failures are documented by output which will appear in the newrelic-cli.log log file and also on the console if the '--debug' command line option is passed to the newrelic CLI.

This PR addresses the following issues:
- newrelic/newrelic-php-agent#208
- newrelic/newrelic-php-agent#212
- newrelic/newrelic-php-agent#213
- newrelic/newrelic-php-agent#222
- newrelic/newrelic-php-agent#227

### DIFF
--- a/recipes/newrelic/apm/php/debian.yml
+++ b/recipes/newrelic/apm/php/debian.yml
@@ -20,7 +20,43 @@ keywords:
   - fpm
 
 processMatch:
+  # several possible configurations we can match on:
+  # - older apache and no php-fpm (process named httpd)
+  # - newer apache and no php-fpm (process named apache2)
+  # - nginx and php-fpm (process named php-fpm)
+  # in the preInstall we will check if php is installed and further filter down the possible scenarios
   - php-fpm
+  - apache2
+  - httpd
+
+preInstall:
+  requireAtDiscovery: |
+      # verify running with sudo
+      if [ -z "$SUDO_USER" ]; then
+        exit 3
+      fi
+
+      # verify systemctl is installed
+      IS_SYSTEMCTL_INSTALLED=$(which systemctl | wc -l)
+      if [ "$IS_SYSTEMCTL_INSTALLED" -eq 0 ]; then
+        exit 3
+      fi
+
+      # Map of tool names to the associated error code
+      required_tools_and_error_codes="grep:10 sed:11 awk:12 egrep:10 curl:22 php:22"
+
+      for tuple in $required_tools_and_error_codes; do
+        tool=$(echo ${tuple} |cut -d':' -f1)
+        code=$(echo ${tuple} |cut -d':' -f2)
+
+        IS_TOOL_INSTALLED=$(which ${tool} | wc -l)
+        if [ "$IS_TOOL_INSTALLED" -eq 0 ]; then
+          exit 3
+        fi
+      done
+
+      # everything looks good!
+      exit 0
 
 validationNrql: "SELECT count(*) from Transaction WHERE host like '{{.HOSTNAME}}%' facet entityGuid since 10 minutes ago"
 
@@ -113,7 +149,7 @@ install:
           fi
         - |
           # Map of tool names to the associated error code
-          required_tools_and_error_codes="grep:10 sed:11 awk:12 egrep:10 curl:22"
+          required_tools_and_error_codes="grep:10 sed:11 awk:12 egrep:10 curl:22 php:22"
 
           for tuple in $required_tools_and_error_codes; do
             tool=$(echo ${tuple} |cut -d':' -f1)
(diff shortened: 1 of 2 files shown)


recipes/newrelic/apm/php/debian.yml
```diff
@@ -20,7 +20,59 @@ keywords:
   - fpm
 
 processMatch:
+  # several possible configurations we can match on:
+  # - older apache and no php-fpm (process named httpd)
+  # - newer apache and no php-fpm (process named apache2)
+  # - nginx and php-fpm (process named php-fpm)
+  # in the preInstall we will check if php is installed and further filter down the possible scenarios
   - php-fpm
+  - apache2
```
Julien4218, line 29: I'm a bit concerned about latching on apache2 and httpd, because those processes don't necessarily means PHP is on the host.
The process detection might/will be used soon to recommend other installations to the user, and this could lead to misunderstanding if we tell the user to go install PHP (manually) when they have nothing to do with PHP.
Could we remove apache2/httpd for the time been? Or do we know a sizeable amount of customers are running apache?

mfulb: @Julien4218,  I'll take it out for now while we discuss what might be the best approach.  I'd really like to get the rest of the PR out so we can see how big an impact it has on failure rates.

Julien4218 (approved):

recipes/newrelic/apm/php/debian.yml
zsistla: Hi @Julien4218 and @mfulb ,
Here are some PHP apache/nginx/php-fpm [numbers](https://one.newrelic.com/launcher/dashboards.launcher?platform[timeRange][duration]=1800000&platform[$isFallbackTimeRange]=true&pane=eyJuZXJkbGV0SWQiOiJkYXNoYm9hcmRzLmhvbWUiLCJlbnRpdHlEb21haW4iOiJWSVoiLCJlbnRpdHlUeXBlIjoiREFTSEJPQVJEIn0=&overlay=eyJuZXJkbGV0SWQiOiJkYXRhLWV4cGxvcmF0aW9uLnF1ZXJ5LWJ1aWxkZXIiLCJpbml0aWFsQWN0aXZlSW50ZXJmYWNlIjoibnJxbEVkaXRvciIsImluaXRpYWxBY2NvdW50SWQiOjEsImluaXRpYWxOcnFsVmFsdWUiOiIiLCJpbml0aWFsUXVlcmllcyI6W3siYWNjb3VudElkIjoxLCJucnFsIjoiU0VMRUNUIENvdW50KCopIGZyb20gUGhwTWV0YWRhdGFTdW1tYXJ5IGZhY2V0IGNhc2VzICggd2hlcmUgb3MudmVyc2lvbiBsaWtlICclZnBtJScgYXMgJ1BIUC1GUE0nLCB3aGVyZSBvcy52ZXJzaW9uIGxpa2UgJyVuZ2lueCUnIGFzICdOR0lOWCcsIHdoZXJlIG9zLnZlcnNpb24gbGlrZSAnJWFwYWNoZSUnIGFzICdBcGFjaGUnKSBzaW5jZSAxIHdlZWsgYWdvIn1dLCJpbml0aWFsQ2hhcnRTZXR0aW5ncyI6eyJjaGFydFR5cGUiOiJDSEFSVF9UQUJMRSIsImxpbWl0IjpudWxsLCJ5U2NhbGUiOnsic3RhdGljIjpmYWxzZSwiZG9tYWluIjpbbnVsbCxudWxsXX0sInlaZXJvIjp0cnVlLCJsaW5rZWRFbnRpdHlHdWlkIjpudWxsLCJsaW5rZWREYXNoYm9hcmRJZCI6bnVsbCwiZGF0YUZvcm1hdHRlcnMiOltdfX0=&state=eeaed09b-4ed8-38c1-f14c-fd7038e40e60) since 1 week ago:

PHP using apache seems to be quite popular.

Is there a way to add PHP detection to the process match?
Here's how we normally [detect](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/newrelic-install.sh#L492) it.

Julien4218: Thanks @zsistla this is quite helpful, I agree the volume on Apache is worth continuing to look at what we can do to handle it.
So the process detection is actually looking at the whole command line, it is also a regex. Is there a way we could define a regex pattern for apache to identify PHP more accurately?
It would be great if we could build a test/deploy to create this configuration, so it would help us tailor that regex and also ensure no regression in the future. Is this something the PHP team could do potentially?

PR #515, opened 2021-09-22
fix: add support for new linux: ubuntu 20.10, 20.04 hirsute, suse 15 and oracle 8

Julien4218:


recipes/newrelic/infrastructure/suse.yml
```diff
@@ -9,7 +9,10 @@ repository: https://github.com/newrelic/infrastructure-agent
 installTargets:
   - type: host
     os: linux
-    platform: "suse"
+    platform: suse
```
jbeveland27, line 12: I wonder - is this first installTarget needed? Or would `platformFamily: suse` be enough?

test/definitions/infra-agent/debians/ubuntu21x04hirsute-infra copy.json
```diff
@@ -0,0 +1,34 @@
...
+    "global_tags": {
+        "owning_team": "OpenSource",
+        "Environment": "development",
+        "Department": "Product",
+        "Product": "Virtuoso"
+    },
+
+    "resources": [{
+        "id": "host1",
+        "display_name": "Ubuntu21.04HirsuteInfraHost",
```
jbeveland27, line 11: Small nit: the name of this file is `ubuntu21x04hirsute-infra copy.json` - maybe rename it to remove the ` copy` part

jbeveland27 (approved):
Looks good overall. Just left some small comments

jbeveland27:
One note: infra-agent I see a test for Oracle 8 but the infra-agent supports Oracle 7 too. Did you want to add a test for that?

recipes/newrelic/infrastructure/suse.yml
Julien4218: I thought about that too, it looks like the older suse populate platform, and wasn't sure if all of those would populate the family as well, since no other linux should be using either with a value of suse, I thought it would be best to have a target for either?

test/definitions/infra-agent/debians/ubuntu21x04hirsute-infra copy.json
Julien4218: Good catch, TY

Julien4218:
> One note: I see a test for Oracle 8 but the infra-agent supports Oracle 7 too. Did you want to add a test for that?

I gave it some thoughts, given our low volume on Oracle Linux, I've opted to test latest only. If we see issues coming up, perhaps we'll look at adding a test?

PR #518, opened 2021-09-27
reinstated apm java recipe for rhel/debian x86 families only

mbruzina:
also updated download url to downloads.newrelic.com and uploaded artifacts for download


mbruzina:
Hold tight, re-instating tests too

Julien4218 (approved):

PR #537, opened 2021-10-08
 infra-agent: offer to install brew if no present

rubenruizdegauna:


recipes/newrelic/infrastructure/darwin.yml
```diff
@@ -63,8 +64,30 @@ install:
...
+                exit 131
+              fi
+              if [[ "$NEW_RELIC_CONTINUE" == "Y" ]]; then
+                break
+              fi
+              echo -e "Please type Y or N only."
+            done
+            echo -e "{{.ARROW}}Installing brew...{{.GRAY}}"
+            echo ""
+            BREW_INSTALL_SUCCESS=$(sudo -i -u $SUDO_USER /bin/bash -c 'curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh | bash > /dev/null 2>&1; echo $?')
```
noahmmcgivern, line 86: Looks like it works, but part of the brew installation requires manual input from the user to install xcode components and accept the license agreement, so the install script fails before this finishes. Is there a way that we can somehow wait until `which brew` returns an exit code 0 instead?

noahmmcgivern (approved):

PR #542, opened 2021-10-11
Removing standalone tomcat instrumentation

mbruzina:


recipes/newrelic/apm/java/linux.yml
```diff
@@ -49,18 +58,31 @@ preInstall:
...
+    # check if any tomcat instances are service-managed
+    nri-lsi-java -list | tr -d '[]' | tr ',' '\n' > $TMP_DIR/all_java_processes
+    for JAVA_PID in $(cat $TMP_DIR/all_java_processes)
     do
       introspectionData=$(nri-lsi-java -introspect ${JAVA_PID})
-      mainClass=$( echo "$introspectionData" | grep -oP '.*\K(?<=mainClass":").*?(?=\")' )
-      if [[ "$mainClass" == "org.apache.catalina.startup.Bootstrap" ]]; then
-        rm -rf $tmpDir
-        exit 0
+      MAIN_CLASS=$( echo "$introspectionData" | grep -oP '.*\K(?<=mainClass":").*?(?=\")' )
```
Julien4218, line 66: @mbruzina this is a gnarly regex, which I don't doubt does what we need to do. Wouldn't this work the same if we were just grepping on "catalina" or even "org.apache.catalina.startup.Bootstrap" if we wanted to be more accurate? I'm not sure we care much about parsing what's returned than making sure there is a catalina somewhere in the content?

recipes/newrelic/apm/java/linux.yml
```diff
@@ -49,18 +58,31 @@ preInstall:
...
 
-    rm -rf $tmpDir
+    if [[ -f $TMP_DIR/tomcat_standalone_process ]]; then
+      echo -e "All detected Tomcat processes are running external to systemd" >&2
+    else
+      echo -e "Tomcat processes were detected, but no running java processes found" >$2
+    fi
+
+
+    rm -rf $TMP_DIR
```
Julien4218, line 85: When we exit here, we know there is no managed tomcat, and the lsi has been installed.
In this case the recipe won't be running (what we want).
I think we should also cleanup the installed lsi and run a yum/apt-get remove nri-lsi-java to leave the host in the same state as we found it before evaluating if the recipe should run.
Here is an example https://github.com/newrelic/open-install-library/blob/feat/java-dynamic/recipes/newrelic/apm/java/linux.yml#L106-L110

Julien4218 (approved):

PR #545, opened 2021-10-13
Update linux recipe to require a .net service running under systemd

JcolemanNR:
Update the pre-install step to verify that a .net systemd service exists before proposing installation. This is a behavior change from previously only requiring a .net process to exist. We currently only support .net services running in systemd with the virtuoso installer, so it is pointless to propose installation unless one exists.

Fixes newrelic/newrelic-dotnet-agent#750

### DIFF
--- a/recipes/newrelic/apm/dotNet/linux-systemd.yml
+++ b/recipes/newrelic/apm/dotNet/linux-systemd.yml
@@ -53,7 +53,7 @@ preInstall:
     
     # .NET processes found, but none are running as systemd services :(
     rm -rf $TMP_DIR >/dev/null
-    exit 20
+    exit 132
 
 validationNrql: "SELECT count(*) FROM NrIntegrationError WHERE purpose = 'New Relic CLI configuration validation' AND hostname like '{{.HOSTNAME}}%' since 10 minutes ago"
 


recipes/newrelic/apm/dotNet/linux-systemd.yml
```diff
@@ -22,20 +22,38 @@ processMatch: []
...
+    # Get .NET introspector tooling
     TMP_DIR=$(mktemp -dq /tmp/newrelic.XXXXXX)
     curl -s https://download.newrelic.com/install/dotnet/introspector/latest/dotnet-is-linux-x64.gz -o ${TMP_DIR}/dotnet-is-linux-x64.gz >/dev/null
     zcat ${TMP_DIR}/dotnet-is-linux-x64.gz > ${TMP_DIR}/nri-lsi-dotnet
     chmod u+x $TMP_DIR/nri-lsi-dotnet >/dev/null
-    processCount=$(${TMP_DIR}/nri-lsi-dotnet -c | tr -d '[]' | tr ',' '\n' | wc -w)
-    if [[ "$processCount" -ne 0 ]]; then
-      # Found .NET processes - return 0
+    
+    # Record all .NET processes
```
jaffinito, line 31: Really like the comments, makes reading this much easier!

recipes/newrelic/apm/dotNet/linux-systemd.yml
```diff
@@ -22,20 +22,38 @@ processMatch: []
...
+              rm -rf $TMP_DIR >/dev/null
+              exit 0
+            fi
+        done
+    done
+    
+    # .NET processes found, but none are running as systemd services :(
     rm -rf $TMP_DIR >/dev/null
-    exit 3
+    exit 20
```
jaffinito, line 56: Question for everyone: Do we still want to exit with code 20 when this runs in the preinstall or code 3 or something else?

Julien4218: Excellent question!
@jaffinito @JcolemanNR 
So with the preInstall script, we can and should, now leverage a specific exit code value of 132 which signals that .net is detected, but our recipe/installer cannot handle it at this time.
This specific event, is then used in the UI to redirect the user to alternate install methodology: For apm, we'll first direct the user to our setup-nerdlet, then if it still doesn't work, we'll end them to our newrelic documentation for .net.

So In short:
- If we know the recipe can install successfully, => exit 0
- If we know there is no apm detected on this host (no .net processes perhaps), then exit code not in [0, 132]
- If we know there we can install apm on this host, but our recipe cannot do it at this time, exit code 132

Let me know if you have any more questions.

PR #548, opened 2021-10-18
Add Kubernetes recipe.

fryckbos:


recipes/newrelic/infrastructure/kubernetes.yml
```diff
@@ -0,0 +1,265 @@
...
+  version: "3"
+  silent: true
+
+  tasks:
+    default:
+      cmds:
+        - task: assert_pre_req
+        - task: assert_kubernetes
+        - task: create_namespace
+        - task: check_privileged
```
roobre, line 40: I would also add a check for `hostPath` volumes, as those are often restricted.

recipes/newrelic/infrastructure/kubernetes.yml
```diff
@@ -0,0 +1,265 @@
...
+          curl -s "${URL}" > newrelic-k8s.yml
+
+    run_kubectl:
+      cmds:
+        - |
+          kubectl apply -f https://raw.githubusercontent.com/pixie-labs/pixie/main/k8s/operator/crd/base/px.dev_viziers.yaml
+        - |
+          kubectl apply -f https://raw.githubusercontent.com/pixie-labs/pixie/main/k8s/operator/helm/crds/olm_crd.yaml
+        - |
+          kubectl apply -f newrelic-k8s.yml
```
roobre, line 256: Installing manifests directly has some UX disadvantages that we might consider:
- It requires the user to store the `newrelic-k8s.yml` file somewhere to completely uninstall the solution
- Re-running the wizard when a new version is released does not guarantee a clean upgrade path:
  - Objects that are not present in the new version will not be garbage collected
  - Depending on the changes, some objects will not fully replace anothers, potentially leading to duplicated deployments
  - Some changes might cause `kubectl` to attempt to overwrite immutable objects, resulting in hard to debug errors.

I suggest defaulting to using the helm chart instead, which does not have these problems.

If `helm` is not available, I'd suggest to _not_ remove `newrelic-k8s.yml` and instead ask the user to keep it somewhere safe, versioned if possible.

recipes/newrelic/infrastructure/kubernetes.yml
```diff
@@ -0,0 +1,265 @@
...
+
+    cleanup:
+      cmds:
+        - |
+          rm newrelic-k8s.yml
```
roobre, line 261: See above

fryckbos: I removed the rm and added a note in the postInstall.

recipes/newrelic/infrastructure/kubernetes.yml
fryckbos: I agree, I've added support for Helm.

Julien4218 (dismissed):
@fryckbos this looks pretty good. I've put a comment to eventually adjust the validation.

I think we're getting close to get this work deployed. Before we do, we need an NR1 UI work so users can be directed to our documentation when the recipe will not complete (aka fail).
This work should be deployed first, since users will be directed back to NR1, would you know if this work has already been done? If not, I can put a ticket, would you know what NR public documentation would be good to show to the users for manual installations?

recipes/newrelic/infrastructure/kubernetes.yml
```diff
@@ -0,0 +1,353 @@
...
+  - EKS
+  - GKE
+  - AKS
+
+processMatch: []
+
+successLinkConfig:
+  type: EXPLORER
+
+validationNrql: "SELECT count(*) FROM K8sClusterSample SINCE 10 minutes AGO"
```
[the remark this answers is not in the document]
fryckbos: That's a good point. I think having a local validation approach would be better, however I'm not sure how that could be implemented. For the first release is it better to omit the validationNrql?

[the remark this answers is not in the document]
Julien4218: @fryckbos I would say yes, let s remove the validationNrql.
Can you add a recipe task to do a get pods until one of our newrelic pods gets active? You'll want to have a timeout, a few minutes should do.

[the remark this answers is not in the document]
mabmierau: In case it's helpful, the guided install UI is checking for a cluster with the given name, and then it checks for visualization nodes. (Not sure if we can even query that service from the CLI?) Without that second check the cluster explorer generally isn't ready. If the UX is still the same as it was when this was implemented, we might want to make some improvements if the plan is to send them to the cluster explorer shortly after the NR pods are active.

Julien4218 (approved):

recipes/newrelic/infrastructure/kubernetes.yml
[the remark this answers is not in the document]
Julien4218: I agree, we need a follow work item to wait and check for the kubernetes data to make its way to NR so our K8 explorer shows something.
@fryckbos do you have another ticket, and or branch?

PR #551, opened 2021-10-20
fix: add arch variable for sles recipe

brushknight:


recipes/newrelic/infrastructure/suse.yml
```diff
@@ -141,9 +141,17 @@ install:
 
     install_infra:
       cmds:
+        - |
+          ARCH=$(uname -m)
```
Julien4218, line 145: @brushknight are we trying to make sure the recipe does NOT run for ARM? There is an easier way to do this by adding a kernelArch to the installTargets, here is an example for Logs https://github.com/newrelic/open-install-library/blob/main/recipes/newrelic/infrastructure/logs/linux-logs.yml#L20

PR #567, opened 2021-11-01
fix: add directory exclusions when ing tomcat.conf

mbruzina:


Julien4218 (approved):
Verified with rhl2-sysd-yum-javatomcat
[...]$ sudo find / \( -path /proc -o -path /sys -o -path /tmp \) -prune -o -name tomcat.conf -print
/etc/tomcat/tomcat.conf

PR #570, opened 2021-11-04
feat: add support on log forwarder for arm platform arch

noly:
In this PR I include a new file to cover log-fwd arm64 support:
* CentOs 8 and CentOs 7 only available for log-fwd v1.7.4
* Ubuntu 20.04, 18.04 and 16.04 (16.04 available until log-fwd v1.7.9)
* Amazon linux 2
* Debian 10, 9 and 8 (debian 8 available until log-fwd v1.7.9)

With current pipeline when customer run the installation script agent will install last version available.


Julien4218:
@noly this is great. I've made a few changes below:
- Given the only difference between the log recipe for arm and the main log recipe was just the additions of the few installTarget, I've just combined those extra target in the main linux-logs.yml (can have any number of definitions).
- While testing, I notice the log recipe wasn't matching for Debian 11. This was because the version of the host came up as "10.11", not "10".

In order to support range of versions, I thought it would be good to allow regex pattern for any of the installTarget field. So I've made this enhancement to the newrelic-cli, which is waiting on a PR review https://github.com/newrelic/newrelic-cli/pull/1176

Assuming this enhancement will be deployed, I've consolidated all the given platform installTarget into a single definition and came up with the regex pattern to allow support for the specific versions.

Please review and let me know your feedback.

noly:
Sound really good @Julien4218!!

Let me know if I can help with something, would be good to have this change deploy this week :)

Julien4218 (approved):
Validated successfully on all arm aws/linux2, debian, centos and ubuntu.
Great work, thank you @noly

PR #571, opened 2021-11-04
Adding detection for python and ruby

Julien4218:
https://newrelic.atlassian.net/browse/VIRTUOSO-1549

### DIFF
--- a/recipes/newrelic/apm/ruby/linux.yml
+++ b/recipes/newrelic/apm/ruby/linux.yml
@@ -1,0 +1,36 @@
+# Visit our schema definition for additional information on this file format
+# https://github.com/newrelic/open-install-library/blob/main/docs/recipe-spec/recipe-spec.md#schema-definition
+
+name: ruby-agent-installer
+displayName: Ruby Agent Installer
+description: New Relic install recipe for instrumenting Ruby applications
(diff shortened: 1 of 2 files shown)


recipes/newrelic/apm/ruby/linux.yml
```diff
@@ -0,0 +1,36 @@
...
+keywords:
+  - Apm
+  - ruby
+
+processMatch:
+  - ruby
+
+preInstall:
+  requireAtDiscovery: |
+    exit 132
```
jbeveland27, line 22: Just wondering - should there be more of a check here like there is for Python? Seems like we'll always report back `DETECTED` if there's a process with `ruby` in the cmdline

Julien4218: Correct, there aren't as many ruby apps running that are not created by a user. 
In python there are a lot of common tools that are there, which we don't want to signal the user they should instrument their "python" app. It's not as strong of a case for ruby.
Here is an example of python process I'm seeing on an EC2/ubuntu, whereas there are no ruby processes
/usr/bin/networkd-dispatcher
/usr/share/unattended-upgrades/unattended-upgrade-shutdown

jbeveland27 (approved):

PR #576, opened 2021-11-17
[VIRTUOSO-1600] add recipe for windows rabbitmq discovery

NRhzhao:


recipes/newrelic/infrastructure/ohi/rabbitmq/windows.yml
```diff
@@ -0,0 +1,63 @@
...
+    try{
+      if (Get-Service "rabbitmq" -ErrorAction SilentlyContinue) {
+         exit 132
+      }
+      
+      $rabbitMQMgmtUrl = "http://localhost:15672/"
+      $response=Invoke-WebRequest -UseBasicParsing -Method Get -TimeoutSec 10 -Uri $rabbitMQmgmtUrl
+      $foundRabbitMQMgmt = $response -match "rabbitmq" -ieq "True"
+      
+      if($foundRabbitMQMgmt){
```
Julien4218, line 32: @NRhzhao this is great. I think we should also check for a response code 200 in addition of checking we have something that matches the string "rabbitmq" in the response, just in case we get an error with an html content that happened to have that string in it.

Julien4218 (approved):
This looks good.

PR #577, opened 2021-11-19
Win2019 mysql recipe

NRhzhao:
- Create recipe


Julien4218 (approved):
Verified on a windows server with mysql installed.

PR #584, opened 2021-12-06
Win elasticsearch recipe

NRhzhao:


recipes/newrelic/infrastructure/ohi/elasticsearch/windows.yml
```diff
@@ -0,0 +1,58 @@
...
+# keyword convention for dealing with search terms that could land someone on this instrumentation project
+keywords:
+  - Infrastructure
+  - Integration
+  - elastic
+  - elasticsearch
+
+# CLI runs process detection; this is used to filter recipes that are appropriate for matched processes
+processMatch:
+  - elasticsearch*
```
Julien4218, line 23: @NRhzhao is there a way to tie up the regex so it matches more specifically the elasticsearch process? I assume it should with .exe at some point.

NRhzhao: @Julien4218  Updated process matching for specific executables, still works for both msi and choco install.  Also add a fallback to check elastic search port if can't find service running

Julien4218 (approved):
@NRhzhao this looks good, I've verified on the windows server provisioned with the test.

PR #588, opened 2021-12-08
feat: unmatched pattern PIDs added to ignore list

mbruzina:
Prevents us from making unnecessary java LSI calls to -introspect

### DIFF
--- a/recipes/newrelic/apm/java/linux.yml
+++ b/recipes/newrelic/apm/java/linux.yml
@@ -304,7 +304,6 @@ install:
                     else
                       echo $(build_stdout_json "Attaching agent to Tomcat process in '$dockerAppName'...complete!")
                     fi
-                    echo $JAVA_PID >> /etc/newrelic-java/install/.ignored_pids
                   else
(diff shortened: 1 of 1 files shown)


recipes/newrelic/apm/java/linux.yml
```diff
@@ -318,6 +318,8 @@ install:
                   fi
                   echo $JAVA_PID >> /etc/newrelic-java/install/.ignored_pids
                 fi
+              else
```
Julien4218, line 321: @mbruzina I notice we have that same echo pid to the ignore list on L319. Should we just move that to be done once on line 283 instead,?

mbruzina: There are actually 3 

- 307 for PIDs matching pattern running in docker
- 319 for PIDs matching pattern not running in docker
- 321 for non-matching PIDs

as I'm typing this, I realize we can remove the if/else on the pattern match and always add the PID in question to 'ignored'...I'll fix that quickly

mbruzina: @Julien4218 change is complete

Julien4218 (approved):

PR #589, opened 2021-12-11
Kafka Auto Discovery for window and linux

NRhzhao:
- for window testing, follow instruction on desktop readme
- for linux testing, start zookeeper and kafka before testing:
   - cd /opt/kafka/ && sudo bin/zookeeper-server-start.sh config/zookeeper.properties
   - cd /opt/kafka/ && sudo bin/kafka-server-start.sh config/server.properties


recipes/newrelic/infrastructure/ohi/kafka/windows.yml
```diff
@@ -0,0 +1,61 @@
...
+
+# Kafka runs under java process in windows, so we checking if kafka port 9092 is running under Java
+processMatch:
+  - java
+
+preInstall:
+  requireAtDiscovery: |
+    powershell -command '
+    try{
+      $process = Get-Process -Id (Get-NetTCPConnection -LocalPort 9092).OwningProcess | select name -ErrorAction SilentlyContinue
```
Julien4218, line 30: @NRhzhao unlike linux which has a process detect specific to kafka, for windows I don't think we should rely on java with port 9092 to determine kafka is used, this can be any other java app that happens to run on that port.
Is there something else we could latch on to ensure kafka is running?

recipes/newrelic/infrastructure/ohi/kafka/rhel.yml
```diff
@@ -0,0 +1,57 @@
+# Visit our schema definition for additional information on this file format
```
Julien4218, line 1: @NRhzhao could we may be combine debian/rhel in a single linux.yml recipe? I'm not seeing any implementation difference at this point

recipes/newrelic/infrastructure/ohi/kafka/debian.yml
```diff
@@ -0,0 +1,53 @@
...
+    setup:
+      cmds:
+        - |
+          exit 131
+
+postInstall:
+  info: |2
+      ⚙️  The kafka configuration file can be found in /etc/newrelic-infra/integrations.d/kafka-config.yml
+      Edit this file to make changes or configure advanced features for this integration. See the docs for options:
+      https://docs.newrelic.com/docs/infrastructure/host-integrations/host-integrations-list/kafka-monitoring-integration/#configg
```
Julien4218, line 53: @NRhzhao there is a typo on the url, the last config word has 2 'g'

Julien4218 (changes requested):
@NRhzhao I've noted a few thing on the PR.

recipes/newrelic/infrastructure/ohi/kafka/debian.yml
NRhzhao: @Julien4218  Sure, I wasn't sure because other recipes all have them separated.  I figure it's due to they have different package managers, so during installation, we might keep them seperate.

recipes/newrelic/infrastructure/ohi/kafka/windows.yml
NRhzhao: @Julien4218 Let me digging into this a little more.

recipes/newrelic/infrastructure/ohi/kafka/rhel.yml
NRhzhao: @Julien4218  Sure, I wasn't sure because other recipes all have them separated.  I figure it's due to them having different package managers, so, during installation, we might keep them separated.

recipes/newrelic/infrastructure/ohi/kafka/windows.yml
```diff
@@ -0,0 +1,53 @@
...
+    kernelArch: x86_64
+
+keywords:
+  - Infrastructure
+  - Integration
+  - kafka
+
+# Kafka runs under java process in windows, so we checking if kafka port 9092 is running under Java
+processMatch:
+  - kafka
```
Julien4218, line 24: @NRhzhao could we also validate on java? Maybe something like java\.exe*.kafka (double check the regex)

test/deploy/linux/kafka/install/roles/prepare/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
+---
+- debug:
+    msg: Starting Zookeper and Kafka Server
+
+- block:
+    - name: Start Zookeeper
+      shell: cd /opt/kafka/ && sudo bin/zookeeper-server-start.sh config/zookeeper.properties &
+      become: true
+      async: 5
```
Julien4218, line 9: @NRhzhao why do we have to wait up to 5s for this task to complete? Could we just run it once and wait for its completion?

test/deploy/linux/kafka/install/roles/prepare/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
+---
+- debug:
+    msg: Starting Zookeper and Kafka Server
+
+- block:
+    - name: Start Zookeeper
+      shell: cd /opt/kafka/ && sudo bin/zookeeper-server-start.sh config/zookeeper.properties &
```
Julien4218, line 7: @NRhzhao It might be good to have an assertion before this statement to check the path/executable exist, and if not throw an explicit error message (that's usually helpful to us when we see a play is not working).

test/deploy/linux/kafka/install/roles/prepare/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
+---
```
Julien4218, line 1: The /prepare step is executed very early in the deployer lifecycle. Given the statements seem to be starting something, it might be better to move this role in a /start step instead of /prepare

test/deploy/windows/kafka/install/roles/prepare/tasks/main.yml
```diff
@@ -0,0 +1,14 @@
+---
```
Julien4218, line 1: @NRhzhao same feedback, this should go in a /start step, instead of prepare

test/deploy/windows/kafka/install/roles/prepare/tasks/main.yml
```diff
@@ -0,0 +1,14 @@
...
+- block:
+    - name: Starting Zookeper
+      win_shell: |
+        c:\kafka\bin\windows\zookeeper-server-start.bat c:\kafka\config\zookeeper.properties
+      async: 5
```
Julien4218, line 6: Same feedback, why timing out at 5s, instead of letting it run until completion?

recipes/newrelic/infrastructure/ohi/kafka/windows.yml
NRhzhao: @Julien4218  I tried java*kafka, and it doesn't seem to be working.   Should I dig into this deeper?

image

test/deploy/linux/kafka/install/roles/prepare/tasks/main.yml
NRhzhao: @Julien4218 the zookeeper/kafka script would block the console because it doesn't 'complete'.  So, the ansible job never really finishes.  With Linux, I think it's okay to run the process in the background with &, but for Windows, I have to manually timeout so it would move to the 2nd part of the script.  Not sure if there is a better way of handling it.

I keep the timeout just to keep consistentcy.

test/deploy/linux/kafka/install/roles/prepare/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
...
+    msg: Starting Zookeper and Kafka Server
+
+- block:
+    - name: Start Zookeeper
+      shell: cd /opt/kafka/ && sudo bin/zookeeper-server-start.sh config/zookeeper.properties &
+      become: true
+      async: 5
+      poll: 5
+    - name: Start Kafka Server
+      shell: cd /opt/kafka/ && sudo bin/kafka-server-start.sh config/server.properties &
```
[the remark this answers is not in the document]
NRhzhao: @Julien4218 response same as above.

test/deploy/linux/kafka/install/roles/prepare/tasks/main.yml
NRhzhao: @Julien4218  will update

test/deploy/linux/kafka/install/roles/prepare/tasks/main.yml
NRhzhao: @Julien4218  will update

test/deploy/windows/kafka/install/roles/prepare/tasks/main.yml
NRhzhao: @Julien4218 will update

test/deploy/windows/kafka/install/roles/prepare/tasks/main.yml
NRhzhao: @Julien4218  Same as above, the script never "finishes", since starting a server in the shell, thus the ansible process doesn't finish.  So, I just put 5 seconds in, so it can move to the next process.

test/deploy/linux/kafka/install/roles/prepare/tasks/main.yml
Julien4218: Ok, so if this is not needed for linux, let's not have it, I have a separate comment for windows to not rely on poll/async

test/deploy/windows/kafka/install/roles/prepare/tasks/main.yml
Julien4218: For windows, see some fire and forget example with powershell, here are a few examples https://stackoverflow.com/questions/49017321/powershell-build-step-fire-and-forget

recipes/newrelic/infrastructure/ohi/kafka/linux.yml
```diff
@@ -0,0 +1,50 @@
...
+    os: linux
+    kernelArch: x86_64
+
+keywords:
+  - Infrastructure
+  - Integration
+  - kafka
+
+processMatch:
+  - kafka
```
Julien4218, line 23: @NRhzhao this is working, however I wonder if we should use the same regex pattern than we use on windows "java.*kafka" to make it a bit more specific to that java process when detecting?

test/deploy/windows/kafka/install/roles/start/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
...
+      poll: 0
+  ignore_errors: yes
+  when: zookeeper_batch_path.stat.exists and kafka_batch_path.stat.exists
+
+- block:
+    - name: end play when path not found
+      debug:
+        msg: "The zookeeper or kafka batch script doesn't exist in c:\\kafka\\bin\\windows, unable to start kafka"
+  when: not zookeeper_batch_path.stat.exists or not kafka_batch_path.stat.exist
+  ignore_errors: yes
```
Julien4218, line 32: @NRhzhao I dont think we should ignore error if we see we can't start? I would also change the ansible debug with a fail command instead as I believe we want to error the play if we couldn't start?

test/deploy/linux/kafka/install/roles/start/tasks/main.yml
```diff
@@ -0,0 +1,34 @@
...
+      async: 5
+      poll: 0
+  when: zookeeper_shell_path.stat.exists and kafka_shell_path.stat.exists
+
+- block:
+    - name: end play when path not found
+      debug:
+        msg: "The zookeeper or kafka shell script doesn't exist in /opt/kafka/, unable to start kafka"
+  when: not zookeeper_shell_path.stat.exists or not kafka_shell_path.stat.exists
+  ignore_errors: yes
```
Julien4218, line 34: @NRhzhao I dont think we should ignore error if we see we can't start? I would also change the ansible debug with a fail command instead as I believe we want to error the play if we couldn't start?

recipes/newrelic/infrastructure/ohi/kafka/linux.yml
NRhzhao: @Julien4218 Will update.

test/deploy/windows/kafka/install/roles/start/tasks/main.yml
NRhzhao: @Julien4218 Since this task only fires when the directory doesn't exist, so it shows an error when running ansible.  I cannot remove it, but not sure that's the right way of displaying.

test/deploy/linux/kafka/install/roles/start/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+    - name: Start Kafka Server
+      shell: cd /opt/kafka/ && sudo bin/kafka-server-start.sh config/server.properties &
+      become: true
+      async: 5
+      poll: 0
+  when: zookeeper_shell_path.stat.exists and kafka_shell_path.stat.exists
+
+- block:
+    - name: Cannot find zookeper, kafka
+      debug:
```
Julien4218, line 31: instead of debug we should use the `fail` command, it's the same as debug but will also throw an error.

test/deploy/windows/kafka/install/roles/start/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
...
+      poll: 0
+    - name: Starting Kafka Server
+      win_shell: c:\kafka\bin\windows\kafka-server-start.bat c:\kafka\config\server.properties
+      async: 5
+      poll: 0
+  when: zookeeper_batch_path.stat.exists and kafka_batch_path.stat.exists
+
+- block:
+    - name: Cannot find zookeper, kafka
+      debug:
```
Julien4218, line 28: instead of debug we should use the `fail` command, it's the same as debug but will also throw an error.

Julien4218 (approved):
@NRhzhao this looks good and works great on both linux and windows. Feel free to deploy when possible.

PR #590, opened 2021-12-14
Adding validation to ensure newrelic pods are starting

Julien4218:


michaelkro:
Could be a followup PR with other CLI UX tweaks, but the output looks something like

```sh
...
clusterrole.rbac.authorization.k8s.io/nri-bundle-newrelic-logging unchanged
clusterrole.rbac.authorization.k8s.io/nri-bundle-nri-prometheus unchanged
clusterrolebinding.rbac.authorization.k8s.io/nri-bundle-newrelic-infrastructure unchanged
configmap/nri-bundle-nri-prometheus-config configured
clusterrole.rbac.authorization.k8s.io/nri-bundle-nri-kube-events unchanged
Running newrelic-infrastructure status check attempt...
newrelic-infrastructure pod started
```

The status check is appended to the resource deployment logs and somewhat gets lost.

Otherwise, I ran the recipe and it worked as expected!

michaelkro (approved):

Julien4218:
@michaelkro I've pushed another commit to actually throw an error when the pod is not starting, versus continuing as if it was started.

michaelkro:
I tested the failure case by malforming one of the `grep` expression in the check (`grep -i "runing"` instead of `grep -i "running"`) and got the expected error

```sh
Running newrelic-infrastructure status check attempt...
newrelic-infrastructure pod is not starting
==> Installing Kubernetes Integration...incomplete.

  --------------------
  Installation Summary

  !  Kubernetes Integration  (incomplete)
```

michaelkro (approved):

Julien4218:
Thanks @michaelkro

PR #592, opened 2021-12-21
Fix Kubernetes pod validation.

fryckbos:
The Kubernetes validation was checking whether there was exactly 1 newrelic-infrastructure pod running. In clusters with multiple nodes, there is 1 newrelic-infrastructure pod per node. This PR fixes to check whether there are more than 0 newrelic-infrastructure pods running.


Julien4218 (approved):
Thanks @fryckbos

PR #598, opened 2021-12-28
Feat/aws integration

noahmmcgivern:


recipes/newrelic/infrastructure/cloud/aws-linux.yml
```diff
@@ -20,9 +20,92 @@ processMatch: []
...
+
+      callerIdentity=$(aws sts get-caller-identity 2>&1)
+
+      if [ "$(echo "$callerIdentity" | grep AccessDenied | wc -l)" -eq "1" ] ; then
+        exit 1
+      fi
+
+      username=$(echo "$callerIdentity" | sed -ne 's/.*"Arn": ".*\/\(.*\)".*/\1/p')
+
+      echo "username: $username"
```
Julien4218, line 48: @noahmmcgivern please remove, we should avoid echoing credential type of info now that we've completed the dev work.

recipes/newrelic/infrastructure/cloud/aws-linux.yml
```diff
@@ -34,12 +117,112 @@ install:
...
+          accountName=$(echo "$callerIdentity" | sed -ne 's/.*Account": "\(.*\)",/\1/p')
+
+          # Download and fix CloudFormation template
+          curl -sLo /tmp/MetricStreams_CloudFormation.yml {{.NEW_RELIC_DOWNLOAD_URL}}cloud_integrations/aws/cloudformation/MetricStreams_CloudFormation.yml
+          sed -ie "s/AWS::AccountId,/\'AWS::AccountId\',/g" /tmp/MetricStreams_CloudFormation.yml
+
+          printf "\nDeploying CloudFormation template, this may take a few minutes...\n"
+          deploy=$(aws cloudformation deploy --template-file /tmp/MetricStreams_CloudFormation.yml --stack-name NewRelic-Metric-Stream --parameter-overrides NewRelicLicenseKey={{.NEW_RELIC_LICENSE_KEY}} --capabilities CAPABILITY_NAMED_IAM 2>&1)
+
+          if [ "$(echo "$deploy" | grep "^An error occurred (AlreadyExistsException)*" | wc -l)" -eq "1" ] ; then
```
Julien4218, line 196: @noahmmcgivern we'll want to check if that stack name already exist in the pre-check, this will avoid doing a lot of work on subsequent run, which would be a frequent case after a new account first install.
We wouldn't want our status data to also be inflated with the subsequent runs, when we essentially NoOp.

Julien4218 (changes requested):
@noahmmcgivern we should also check for user level administrator access in the preInstall.

mbruzina:
[          curlQuery=$(echo $query | curl https://api.newrelic.com/graphql \](https://github.com/newrelic/open-install-library/blob/72224d33f25ad688ff94a526aa36cb9945aad7ab/recipes/newrelic/infrastructure/cloud/aws-linux.yml#L217-L220)

Might want to derive this URL from `.NEW_RELIC_REGION` for eu/staging-specific URLs

recipes/newrelic/infrastructure/cloud/aws-linux.yml
```diff
@@ -20,10 +20,103 @@ processMatch: []
...
+        exit 1
+      fi
+
+      function isUserAdmin {
+        username=$(echo "$1" | sed -ne 's/.*"Arn": ".*\/\(.*\)".*/\1/p')
+
+        userPolicies=$(aws iam list-user-policies --user-name $username 2>&1)
+
+        if [ "$(echo "$userPolicies" | grep AccessDenied | wc -l)" -eq 1 ] ; then
+          return 1
```
Julien4218, line 52: @noahmmcgivern could you toggle those return values to match the semantic of the function name "isUserAdmin"? Given the phrasing I would expect this function to return 1 if the user is an admin.

recipes/newrelic/infrastructure/cloud/aws-linux.yml
```diff
@@ -20,10 +20,103 @@ processMatch: []
...
+
+      if [ "$(isUserAdmin $callerIdentity)" -eq 1 ]; then
+        exit 1
+      fi
+
+      deployed=$(aws cloudformation describe-stacks --stack-name NewRelic-Metric-Stream 2>&1 | grep "^An error occurred (ValidationError)*" | wc -l)
+
+      if [ "$deployed" -eq "0" ] ; then
+        exit 1
       fi
```
Julien4218, line 119: @noahmmcgivern it might be good to give a final exit 0 and a comment to indicate success. Here is an example from another recipe:

      # everything looks good!
      exit 0

recipes/newrelic/infrastructure/cloud/aws-linux.yml
```diff
@@ -39,7 +132,112 @@ install:
...
+                              "sts:ExternalId": "{{.NEW_RELIC_ACCOUNT_ID}}"
+                          }
+                      }
+                  }
+              ]
+          }
+          EOT
+          )
+
+          createRole=$(aws iam create-role --role-name $roleName --assume-role-policy-document file://<(echo $policy) 2>&1)
```
Julien4218, line 160: @noahmmcgivern could we do a get-role before attempting to create it?

recipes/newrelic/infrastructure/cloud/aws-linux.yml
```diff
@@ -39,7 +132,112 @@ install:
...
+
+          getRole=$(aws iam get-role --role-name $roleName 2>&1)
+
+          if [ "$(echo "$getRole" | grep "^An error occurred (NoSuchEntity)*" | wc -l)" -eq 1 ] ; then
+              echo "$createRole" >&2
```
Julien4218, line 165: @noahmmcgivern it's a good idea to assert the role exists. However we'll need more user guidance here if for some reason the role can't be created. Maybe refer the user to the doc for creating the role themselves (or have someone do it for the user).

recipes/newrelic/infrastructure/cloud/aws-linux.yml
```diff
@@ -39,7 +132,112 @@ install:
...
+              echo "CloudFormation stack is already being deployed"
+              rm /tmp/MetricStreams_CloudFormation.yml
+              exit 0
+          fi
+
+          echo "CloudFormation deployment complete"
+
+          rm /tmp/MetricStreams_CloudFormation.yml
+
+          getRole=$(aws iam get-role --role-name $roleName 2>&1)
```
Julien4218, line 222: @noahmmcgivern we already did a getRole above, should we re-use the previous getRole variable instead of looking up the role again?

Julien4218 (changes requested):
@noahmmcgivern I've noted a few more things to review. Also it looks like the PR tests are not passing. You may want to pull latest main in the branch.

recipes/newrelic/infrastructure/cloud/aws-linux.yml
```diff
@@ -20,10 +20,104 @@ processMatch: []
...
+        exit 1
+      fi
+
+      function isUserAdmin {
+        username=$(echo "$1" | sed -ne 's/.*"Arn": ".*\/\(.*\)".*/\1/p')
+
+        attachedUserPolicies=$(aws iam list-attached-user-policies --user-name $username 2>&1)
+
+        if [ "$(echo "$attachedUserPolicies" | grep AccessDenied | wc -l)" -eq 1 ] ; then
+          false
```
Julien4218, line 52: @noahmmcgivern this is quite difficult to read and understand what the function isUserAdmin is doing. Could we just return a 1 to indicate the uer is an admin, and a 0 when it's not admin?

recipes/newrelic/infrastructure/cloud/aws-linux.yml
```diff
@@ -20,10 +20,104 @@ processMatch: []
...
       code=$(curl -I http://169.254.169.254/latest/meta-data/instance-id -w %{response_code} -so '/dev/null')
-      if [ $code == "200" ]; then
-        exit 132
+
+      if [ $code != "200" ]; then
+        exit 1
+      fi
+
+      if [ ! -f ~/.aws/credentials ]; then
+        exit 1
```
Julien4218, line 29: @noahmmcgivern 
After we know that this is a linux instance running on AWS (curl to IP succeeded), all the exit codes for error should be 132 to indicate a DETECTED status, so the user can see our documentation for manually installing.

recipes/newrelic/infrastructure/cloud/aws-linux.yml
Julien4218: @noahmmcgivern it would be good to have this final exit 0 to be explicit about success to run the recipe

Julien4218 (changes requested):
@noahmmcgivern it's misleading to have a function call the exit directly, it's a side effect that is typically best to avoid.
I believe we talked about it a few times, could you return a variable and handle the exit outside of the function?
cc @NRhzhao

Julien4218 (approved):

sanderblue (approved):

PR #601, opened 2022-01-03
Postgres install

NRhzhao:


recipes/newrelic/infrastructure/ohi/postgres/debian.yml
```diff
@@ -71,6 +65,22 @@ install:
             exit 1
           fi
 
+    assert_pre_req:
+      cmds:
+        - |
+          # Map of tool names to the associated error code
+          required_tools_and_error_codes="date:41 sha256sum:42 base64:43 head:44"
```
Julien4218, line 72: @NRhzhao I see we re using awk in the recipe, we should ensure that tool is installed.

recipes/newrelic/infrastructure/ohi/postgres/debian.yml
```diff
@@ -95,101 +105,67 @@ install:
...
-                printf "PEM key file (if applicable): "
-                read -r NR_CLI_CERT_KEY
-                printf "Trust certificate? (false/true) (default: false): "
-                read -r NR_CLI_TRUST_SERVER_CERTIFICATE
-                NR_CLI_TRUST_SERVER_CERTIFICATE=${NR_CLI_TRUST_SERVER_CERTIFICATE:-false}
-                if [[ "$NR_CLI_TRUST_SERVER_CERTIFICATE" == "true" ]]; then
-                  printf "Certificate Authority file (if applicable): "
-                  read -r NR_CLI_CERT_AUTH_FILE
-                fi
+          while [[ $TRIES -lt {{.MAX_RETRIES}} && $IS_DATABASE_VALID -eq "0" ]]; do
```
Julien4218, line 160: I'm not sure why we're looping here, since it will redo the exact same call to try_create_user.
Unless we're prompting for a username/password the user may have created, and we want to check we can connect, I don't think we need to loop, and perhaps we don't need a try_create_user function and could just process the creation of the user inline?

NRhzhao: @Julien4218  The loop is to re-prompt the user for postgres user password.  If they miss typing the first type when it tries to create a user, we give them a few more chances to type it in correctly.

Julien4218: @NRhzhao ho I understand now, thanks.
It might be good to add a comment to explain, I don't know that it's obvious enough for when we'll look at it down the road.
Also if you could split the part that fetches IS_DATABASE_VALID from the try_create_user function, it would be a bit more readable. It's not obvious at first glance the function has the side effect of setting the variable. You may even consider removing the function and just put it inline in the loop, since it's only called from within the loop.

Julien4218 (commented):
@NRhzhao I notice one of the test is currently failing. It seems to loop on prompting for the password.

Julien4218 (approved):
It works great for both interactive/non-interactive, prompt/no-prompt password.

PR #603, opened 2022-01-04
Redis install

NRhzhao:


Julien4218 (approved):
@NRhzhao this looks good, I've tested on debian and rhel.

PR #605, opened 2022-01-05
Fixes for installing Pixie on Minikube

fryckbos:
The PR contains 2 fixes:

1. **Fix driver matching**: the driver matching logic was wrong and worked because the driver variable contained whitespaces. The PR removes the whitespaces and does the correct matching. Tested with hyperkit driver on macos.

2. **Find minikube in default location**: due to sudo /usr/local/bin might not be in path for the sudo-ed user. The PR adds logic to find minikube in /usr/local/bin if it cannot be found using which.

### DIFF
--- a/recipes/newrelic/infrastructure/kubernetes.yml
+++ b/recipes/newrelic/infrastructure/kubernetes.yml
@@ -206,13 +206,14 @@ install:
               fi
             fi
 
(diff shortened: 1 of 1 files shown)


Julien4218 (approved):
This looks good and tested/verified on a minikube ec2 with docker driver

PR #608, opened 2022-01-06
Large refactor along with deployment markers on java PID attachment

mbruzina:


Julien4218 (approved):
@mbruzina this looks great, thanks.

PR #611, opened 2022-01-06
[Virtuoso-1714] Improve OHI Memcached installation recipe

NRhzhao:


recipes/newrelic/infrastructure/ohi/memcached/debian.yml
```diff
@@ -81,49 +66,117 @@ install:
...
+
+          # Set default values for variables
+          NR_CLI_HOST=${NR_CLI_HOST:-localhost}
+          NR_CLI_PORT=${NR_CLI_PORT:-11211}
+          MAX_RETRIES=3
+          TRIES=0 
+          CAN_CONNECT=$(curl $NR_CLI_HOST:$NR_CLI_PORT 2>&1 1>/dev/null -s -S | awk -F'[()]' '{print $2}')
+
+          while [[ $TRIES -lt $MAX_RETRIES  && ( $CAN_CONNECT == "6" || $CAN_CONNECT == "7" ) ]]; do
+            if [ $CAN_CONNECT == "6" ]; then
```
Julien4218, line 97: @NRhzhao we should default to localhost with no prompting

recipes/newrelic/infrastructure/ohi/memcached/debian.yml
```diff
@@ -81,49 +66,117 @@ install:
...
+            ((TRIES++))
+
+            CAN_CONNECT=$(curl $NR_CLI_HOST:$NR_CLI_PORT 2>&1 1>/dev/null -s -S | awk -F'[()]' '{print $2}')
+
+            if [ $CAN_CONNECT == "6" ]; then
+              printf "\n[Error]: Provided host ($NR_CLI_DB_HOSTNAME) cannot be reached.\nSee https://docs.newrelic.com/docs/integrations/host-integrations/host-integrations-list/postgresql-monitoring-integration#comp-req for more info.\n" >&2
+              if [ ! $TRIES -lt $MAX_RETRIES ]; then exit 6; fi
+              echo "Please try again"
+            elif [ $CAN_CONNECT == "7" ]; then
+              printf "\n[Error]: Connections to port ($NR_CLI_DB_PORT) are being refused.\nSee https://docs.newrelic.com/docs/integrations/host-integrations/host-integrations-list/postgresql-monitoring-integration#comp-req for more info.\n" >&2
```
Julien4218, line 115: @NRhzhao maybe we can craft a better message when we see we can't connect with the port, otherwise this sounds a bit alarming as a user, particularly since this is what users are going to see before been able to specify a port when the default port is not used (They havent had a chance to input that information yet).
Maybe something like "We can't connect to the rabbitmq on localhost using the port 11211, please enter the port number of your rabbitmq instance".

recipes/newrelic/infrastructure/ohi/memcached/debian.yml
```diff
@@ -81,49 +66,117 @@ install:
...
+            printf "Memcached Password: "
+            stty -echo
+            read -r NR_CLI_DB_PASSWORD
+            stty echo
+            ((TRIES++))
+
+            CAN_CONNECT=$(memcstat --servers="$NR_CLI_HOST" --username="$NR_CLI_DB_USERNAME" --password="$NR_CLI_DB_PASSWORD" | grep uptime | wc -l)
+            
+            if [ $CAN_CONNECT -eq 0 ] ; then
+              printf "\n\n[Error]: Memcached server is not accessible with the provided username, password.\nSee https://docs.newrelic.com/docs/integrations/host-integrations/host-integrations-list/postgresql-monitoring-integration#comp-req for more info.\n" >&3
```
Julien4218, line 143: @NRhzhao similarly here, the [Error]: is a bit scary, Maybe we could rephrase to not be as alarming.

recipes/newrelic/infrastructure/ohi/memcached/debian.yml
NRhzhao: Updated

recipes/newrelic/infrastructure/ohi/memcached/debian.yml
NRhzhao: Updated

recipes/newrelic/infrastructure/ohi/memcached/debian.yml
```diff
@@ -81,49 +66,107 @@ install:
...
+          while [[ $TRIES -lt $MAX_RETRIES  && $CAN_CONNECT == "7" ]]; do
+            printf "\nMemcached Port (default: 11211): "
+            read -r NR_CLI_DB_PORT
+            NR_CLI_DB_PORT=${NR_CLI_DB_PORT:-11211}
+            ((TRIES++))
+
+            CAN_CONNECT=$(curl $NR_CLI_HOST:$NR_CLI_PORT 2>&1 1>/dev/null -s -S | awk -F'[()]' '{print $2}')
+
+            if [ $CAN_CONNECT == "7" ]; then
+              printf "\n\nFail to connect to Memcached on localhost using port 11211.  Please enter the port number for your Memcached instance.\n" >&2
```
Julien4218, line 105: @NRhzhao sorry, but the "Fail" semantic is a bit harsh to users, again particularly if users just happen to use a different port than the standard port and were never asked to enter a port number.
Maybe we could just trim the beginning and keep the "Please enter the port number for your Memcached instance", since we know the default doesn't work, we don't have to display it and have them find out their own port number without knowing what the default is anyway.

NRhzhao: updated

recipes/newrelic/infrastructure/ohi/memcached/debian.yml
```diff
@@ -81,49 +66,107 @@ install:
...
+          NR_CLI_HOST=${NR_CLI_HOST:-localhost}
+          NR_CLI_PORT=${NR_CLI_PORT:-11211}
+          MAX_RETRIES=3
+          TRIES=0 
+          CAN_CONNECT=$(curl $NR_CLI_HOST:$NR_CLI_PORT 2>&1 1>/dev/null -s -S | awk -F'[()]' '{print $2}')
+
+          while [[ $TRIES -lt $MAX_RETRIES  && $CAN_CONNECT == "7" ]]; do
+            printf "\nPlease enter the port number for your Memcached instance: "
+            read -r NR_CLI_DB_PORT
+            NR_CLI_DB_PORT=${NR_CLI_DB_PORT:-11211}
```
Julien4218, line 99: @NRhzhao I found an issue while testing the user entering a port.
It looks like we're capturing the port in a different variable name, notice the extra _DB. I let you fix the names to have a consistent one.
Also when prompting for port here, we shouldn't default to the typical port.

Julien4218 (changes requested):
The user cannot enter a custom (correct/working) port.

recipes/newrelic/infrastructure/ohi/memcached/debian.yml
NRhzhao: fixed

Julien4218 (approved):
@NRhzhao this works beautifully now.

PR #612, opened 2022-01-07
Small fixes to Kubernetes verification

fryckbos:
On macOS, the output of the `wc -l` command contains whitespace that messes up the if check afterwards. I've added a sed to remove the whitespace, which works both linux and macOS.

On my local machine, it sometimes takes more than 2 minutes for the newrelic-infrastructure pod to become running on minikube, this PR increases the timeout to 5 minutes.


Julien4218 (approved):
Great fixes, thanks @fryckbos

PR #617, opened 2022-01-13
Set proxy configuration when defined for APM recipes

Julien4218:


JcolemanNR (approved):
Looks good, I tested locally and it works for me with a proxy

sanderblue (approved):

PR #618, opened 2022-01-14
Add proxy configuration for apm/php recipe

Julien4218:


zsistla (dismissed):
Looks straightforward.  I know @ZNeumann might have more context/expertise about the sanitizing of the data to prevent malicious input being added.

Julien4218:
@zsistla after testing with and without proxy, I found the sed needed to be updated to properly update the value in the config file. Feel free to re-review.

mbruzina (approved):

sanderblue (approved):

zsistla:
Although I see now you already merged, so looks great :) The comment above was just a nitpick.

Julien4218:
Ho sorry @zsistla I didn't see a suggestion for doing something different, if there is something I should do please point me to an example, I'd be happy to update it. I've copied the work on the handling of collector_host right above it.

PR #619, opened 2022-01-17
Move pre-checks for Kubernetes to the install step

fryckbos:
We still detect Kubernetes based on the presence of kubectl. Other pre-checks are moved to the install step in order to provide the customer with feedback on why the installation is not successful.


Julien4218 (approved):
@fryckbos this looks good. One suggestion perhaps down the road, we should capture an unsupported error code (131), since we actually don't support installing in this conditions. We can do this later if you like.

PR #620, opened 2022-01-18
Golden signal install once

NRhzhao:
[VIRTUOSO-1738] - Install Golden Signal Alert only once

- I have to duplicate the NRQL call to check for alert policy in pre-install.   I tried passing down the variable from NRQL query since it's used in installation as well, but it doesn't seem like pre-install and install share the same shell, so it's unable to.
- Not sure if there is a better way to do this.

### DIFF
--- a/recipes/newrelic/infrastructure/alerts/golden.yml
+++ b/recipes/newrelic/infrastructure/alerts/golden.yml
@@ -22,6 +22,8 @@ preInstall:
       - Low Application Throughput
       This policy can be modified under Alerts & AI\Policies.
   requireAtDiscovery: |
+      if [[ "$NR_CLI_SHOW_GOLDEN_OPTION" == "true" ]] ; then exit 0; fi 
+
       # recipe per account base, will skip if already installed 
       NEW_RELIC_API_URL=$(echo -n 'https://api.newrelic.com')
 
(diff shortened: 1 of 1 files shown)


recipes/newrelic/infrastructure/alerts/golden.yml
```diff
@@ -21,6 +21,34 @@ preInstall:
       - High Application Response Time
       - Low Application Throughput
       This policy can be modified under Alerts & AI\Policies.
+  requireAtDiscovery: |
+      if [[ "$NR_CLI_SHOW_GOLDEN_OPTION" == "true" ]] ; then exit 0; fi
```
Julien4218, line 25: @NRhzhao are we missing a line before that like this:
NR_CLI_SHOW_GOLDEN_OPTION="{{.NR_CLI_SHOW_GOLDEN_OPTION}}"

Julien4218 (commented):
@NRhzhao this looks good, except perhaps the init in the requireAtDiscovery script, could you double check the comment?

recipes/newrelic/infrastructure/alerts/golden.yml
NRhzhao: @Julien4218  Hi Julien, I belive {{.NR_CLI_SHOW_GOLDEN_OPTION}} is part of go-task.
The pre-install is not part of go-task, and it's not getting the variable with that notation.

It just use the regular $ notation.

Julien4218: Makes sense, thanks @NRhzhao I will test. I kind of think we should also have a task at some point to make the preInstall script work the same (support the {{. }} templating, but that may not be a small effort), so we'll see.

Julien4218 (approved):
@NRhzhao this works perfectly. Please deploy when time permits.

PR #621, opened 2022-01-19
Improve Kubernetes recipe

fryckbos:
Improve Kubernetes recipe:
 * check standard path for kubectl
 * detect OLM and adjust Pixie install to avoid conflicting OLM deployment
 * suppress unwanted output when detecting the cluster type
 * fix wrong logic in detecting whether namespace already exists

Tested on cluster with and without OLM installed.


Julien4218 (approved):
This looks good @fryckbos , I'll deploy shortly.

PR #636, opened 2022-02-07
infrastructure/kubernetes: pin latest version compatible with helm v2

roobre:
`nri-bundle` versions older than `3.2.11` are not compatible with Helm v2, as they use a v3-introduced function `get`: https://github.com/newrelic/helm-charts/issues/671

This PR pins version `3.2.11` if Helm v2 is detected during installation.

### DIFF
--- a/recipes/newrelic/infrastructure/kubernetes.yml
+++ b/recipes/newrelic/infrastructure/kubernetes.yml
@@ -339,6 +339,13 @@ install:
               fi
             fi
 
+            if helm version --short --client | grep 'v2' > /dev/null; then
+              echo "Helm version 2.x detected. Support for Helm v2 will be removed in the near future,"
+              echo "please upgrade at your earliest convenience."
+              echo "Will install Kubernetes Integration v2.8.3, last version compatible with Helm v2."
+              ARGS="${ARGS} --version 3.2.11"
+            fi
+
             $SUDO helm upgrade $ARGS
           else
             echo ""


recipes/newrelic/infrastructure/kubernetes.yml
```diff
@@ -339,6 +339,13 @@ install:
               fi
             fi
 
+            if helm version --short --client | grep 'v2' > /dev/null; then
+              echo "Helm version 2.x detected. Support for Helm v2 will be removed in the near future,"
+              echo "please upgrade at your earliest convenience."
+              echo "Will install Kubernetes Integration v2.8.3, last version compatible with Helm v2."
+              ARGS="${ARGS} --version 3.2.11"
```
roobre, line 346: Please note that the difference between these two versions is intended. Here we're pinning the chart version, which includes the aforementioned nri-kubernetes version:

```shell
helm template newrelic-bundle newrelic/nri-bundle --version 3.2.11 --set global.cluster=x --set global.licenseKey=x | grep 'newrelic/infrastructure-k8s'
          image: "newrelic/infrastructure-k8s:2.8.3"
```

roobre:
Closing as we have decided the CLI should not support Helm v2 ([internal thread](https://newrelic.slack.com/archives/C01PTDC51K2/p1644228401353089?thread_ts=1643968956.609059&cid=C01PTDC51K2)).

PR #646, opened 2022-02-16
fix: added infra to node recipe dependencies

noahmmcgivern:


Julien4218 (changes requested):
This work should be part of another MMF.

PR #647, opened 2022-02-18
IMDS Security Update

NolanT:
* Azure: add [required header](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/instance-metadata-service?tabs=linux#security-and-authentication)`Metadata: true`
* AWS: upgraded check for compatibility with both IDMSv1 and IDMSv2
   * `http://169.254.169.254/latest/meta-data/instance-id` returns a `401` error on "IDMSv2 only" EC2 instances
   * token request returns a `200` on either IMDS mode


Julien4218 (changes requested):
@noahmmcgivern to enter separate PR for Azure and ticket for AWS.

PR #648, opened 2022-02-23
Quote special chars

maxlemieux:
Passwords as written to the config file, don't have quotes around them. This causes problems parsing YML if the password has special characters in it.

This change simply adds quotes around the two cases in each Linux recipe which involve passwords in the YML.

TODO: figure out how to do this for the PowerShell recipe.

Super Twos Day Bug Bash!


maxlemieux:
Updating tests then reopening.

maxlemieux:
Re-doing as branch in this project per earlier direction, so it can pass the tests.

maxlemieux:
@Julien4218 I don't have permissions to branch the project, please advise. Thanks

Julien4218:
@maxlemieux I can add you to the group once your profile is added to the GH newrelic org

maxlemieux:
> @maxlemieux I can add you to the group once your profile is added to the GH newrelic org

Should be all set now.

PR #649, opened 2022-02-23
fix: aws imds v1 and v2 support

noahmmcgivern:


Julien4218 (changes requested):
@noahmmcgivern we shouldn't wire solely on v2, the v1 is the version that is supported by default and indefinitely (from AWS doc).
We should try first on v1, and if we don't get detection, retry with v2. Users that have updated their config with v2 can disable v1.

Julien4218:
@noahmmcgivern any update on this? I also left some feedback in slack.

Julien4218 (approved):

PR #650, opened 2022-02-23
fix: azure add imds metadata header

noahmmcgivern:


recipes/newrelic/infrastructure/cloud/azure-linux.yml
```diff
@@ -18,7 +18,7 @@ processMatch: []
 
 preInstall:
   requireAtDiscovery: |
-      code=$(curl -I http://169.254.169.254/metadata/instance/compute/vmId -w %{response_code} -so '/dev/null')
+      code=$(curl -H Metadata:true --noproxy "*" -I http://169.254.169.254/metadata/instance/compute/vmId -w %{response_code} -so '/dev/null')
```
Julien4218, line 21: @noahmmcgivern this doesn't seem to work for me, just like the previous implementation.
This syntax works for me 
curl -H Metadata:true --noproxy "*" "http://169.254.169.254/metadata/instance?api-version=2021-02-01" -w %{response_code} -so '/dev/null'

For reference https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service?tabs=linux

Could you update and validate on an Azure VM (I used Ubuntu 20), and do the same for the windows case?

Julien4218 (changes requested):
Needs fixing

Julien4218 (approved):

PR #659, opened 2022-03-15
chore: add unsupport for infra recipes inside docker or wsl

NRhzhao:


Julien4218 (commented):
@NRhzhao this looks good.
Would you know what the difference is for the logic comparing to what is done during the recipe execution? https://github.com/newrelic/open-install-library/blob/main/recipes/newrelic/infrastructure/amazonlinux2.yml#L85-L96

Should we be consistent, or remove the recipe execution snippet?
If you could also structure the logic as a serie of if statement that returns 131, instead of nesting another if for the WSL check, that would make it easier to read/understand.

Julien4218 (approved):

PR #661, opened 2022-03-16
feat(recipes): added sudo check in rhel infra recipes

mbruzina:


Julien4218 (approved):
@mbruzina we dont need this for debian/ubuntu and suse?

mbruzina:
@Julien4218 yes, there are separate tasks to address other distro families

PR #666, opened 2022-03-29
Enables in-agent log forwarding by default

jaffinito:
Updates .NET Windows and Linux recipes to enabled in-agent log forwarding by default.

Windows:
Uses the newrelic.config to enable log forwarding since this is the primary way Windows users interact with agent config.
Updated configuration task and followed same pattern as other newrelic.config updates.

Linux:
Uses the environment variables to enable log forwarding since this is the primary way Linux users interact with the config.
Updated configuration task and followed same pattern as other config env vars.

Notes on release:
The .NET teams current target for the release that will support these changes is 03/29, but can wait for this to be pushed out on 03/31.

Resolves newrelic/newrelic-dotnet-agent#1020


jaffinito:
I updated the recipes to no longer have explicit DT settings.

Julien4218:
> I updated the recipes to no longer have explicit DT settings.

Much appreciated, thank you.

Julien4218 (approved):
Verified non regression on linux and windows.
Validated updated newrelic config for windows.

PR #668, opened 2022-03-29
chore(install): add nrql validation for k8s to pull entity id post in…

NRhzhao:


Julien4218 (commented):
@NRhzhao I don't think this will work because the cluster name is also determined by asking K8s directly in the recipe.

PR #670, opened 2022-04-04
chore(install): updated APM recipes to return entity guid

NRhzhao:


Julien4218 (changes requested):
@NRhzhao this actually won't work, because user can install the java instrumentation with no application currently running, and those can be started at a later time. This is a similar use case to what we do with windows for .net/iis.
I believe the entityGuid here is intended for the UI, we should either implement a `validationScript` hook in the recipe, or the more abstract JSON file mechanism to read back from the recipe, or alternatively have the UI do the lookup entityGuid logic.
cc @mbruzina

PR #671, opened 2022-04-05
fix(k8s): exit when no running cluster is detected

mbruzina:


mbruzina:
Explanation as to where are heads were at on this one...

We attempted to use `kubectl config get-contexts` to detect any running clusters, though the behavior on OSX differed than Linux using minikube.  

linux w/o running cluster
```
$ kubectl config get-contexts
CURRENT   NAME   CLUSTER   AUTHINFO   NAMESPACE
```

OSX w/o running cluster
```$ kubectl config get-contexts
CURRENT   NAME             CLUSTER          AUTHINFO         NAMESPACE
*         docker-desktop   docker-desktop   docker-desktop
```

Instead we decided on using `kubectl cluster-info dump pod-running-timeout 5` - key being the pod-running-timeout option.  Again, behaviors differed on this command between OSX and Linux

Linux (executed immediately)
```
$ kubectl cluster-info dump pod-running-timeout 5
error: cluster "minikube" does not exist
```

OSX (took 5 seconds
```
$ kubectl cluster-info dump pod-running-timeout 5 all-namespaces true
Unable to connect to the server: EOF
```

When this command _does_ detect a cluster, we do a really primitive grep on the payload and use this to indicate a cluster is running.  If this payload isn't present, we exit with a new code (45)

Julien4218 (commented):
@mbruzina I've tested local on my M1, and also on an EC2 minikube. It worked well for both.
The detection on ec2/minikube is still pretty fast (25ms observed).
However on my mac, it takes the timeout seconds to get back, which is currently 5s.

Could we do a couple of things:
1) decrease the cluster-info timeout to 2s, this is more than enough to wait. Anything longer would potentially slowdown other detection and the overall installation
2) could we add another check/filter before using the `cluster-info` command by doing a `kubectl config get-contexts` and see that we have a non empty config? Checking the config is a local command, so this should be pretty fast and at least we can exit early for some cases when there is a kubectl with no context. We'll still need the `cluster-info` to further filter down the mac usages, but this might speed up the negative detection of some other linux usages.

PR #678, opened 2022-04-29
feat(logging): use winevtlog as default input for windows logs

danybmx:
Hi!

This change is related to https://github.com/newrelic/infrastructure-agent/pull/1125

Fluent-bit has a new input plugin for windows logs that come to "replace" winlog, by now we're updating the infrastructure-agent to accept this new plugin type and the next step will be to set it as the default plugin when installing logging for windows.

We thought on make this change transparent to the users by replacing `winlog > winevtlog` directly when generating the fluent-bit configuration, but the output generated by those plugins have different attribute names and so on, so we discard that option.

Regards!

https://docs.fluentbit.io/manual/pipeline/inputs/windows-event-log-winevtlog
https://fluentbit.io/announcements/v1.9.0/

### DIFF
--- a/recipes/newrelic/infrastructure/logs/windows-logs.yml
+++ b/recipes/newrelic/infrastructure/logs/windows-logs.yml
@@ -44,7 +44,7 @@ install:
         LOGS_CONTENT: |
           logs:
             - name: windows-security
-              winlog:
+              winevtlog:
                 channel: Security
                 collect-eventids:
                 - 4740
@@ -57,7 +57,7 @@ install:
                 - 4648
 
             - name: windows-application
-              winlog:
+              winevtlog:
                 channel: Application
 
             - name: newrelic-cli.log


Julien4218:
Hello @danybmx I've recreated this PR from a branch on the repo, so that the regression tests can run and ensure the recipes are still working properly.
https://github.com/newrelic/open-install-library/pull/763

Would you know if/when this work will be ready to be merged and deployed? I notice the DO NOT MERGE comment in your description (thanks for that by the way).

danybmx:
Indeed @Julien4218 it's ready to be deployed, sorry didn't realise about this one. Go ahead if tests are passing on the new one and thanks a lot for taking care.

danybmx:
As previous comments this is follow up on #763

Julien4218:
Perfect, will do. Thanks @danybmx

PR #681, opened 2022-05-03
Consolidate rhel and debian ohi/nginx recipes

katiehrenchir:
- Add isYumInstalled check to the nginx install recipe and install via yum if true
- Otherwise install with apt-get
- Add param for manual test roles to add status url configuration


natecanfield822 (approved):
Nice work!

PR #687, opened 2022-05-13
chore(infra): Allow root installs without sudo

amolero-nr:
Asserting for root/sudo user and removing sudo package usage from internal script commands

Reference:
https://issues.newrelic.com/browse/NR-6525


Julien4218 (approved):
This looks good

PR #693, opened 2022-05-24
fix: Detect existing status page

natecanfield822:


recipes/newrelic/infrastructure/ohi/apache/rhel.yml
```diff
@@ -57,6 +57,11 @@ preInstall:
       # Supported case
       exit 0
     fi
+    STUB_STATUS_IN_USE=$(httpd -t -D DUMP_INCLUDES | grep '/' | awk '{ print $2 }' | xargs grep '/server-status' | wc -l)
+    if [ $STUB_STATUS_IN_USE -gt 0 ] ; then
+      #Stub status page is enabled and in use, but not accessible from local machine
+      exit 132
```
katiehrenchir, line 63: how did you know to exit with `132`?

natecanfield822: Good question - since this is in the preinstall script this will trigger a detected status in the CLI - meaning we know there's apache installed that could be instrumented, but we can't do it through our recipe right now. See: https://github.com/newrelic/open-install-library/blob/b8f8b67a74f6bd9ec3c860f5a9360488669916c3/docs/error-codes.md

katiehrenchir (approved):

recipes/newrelic/infrastructure/ohi/apache/rhel.yml
katiehrenchir: bookmarked! thanks!

PR #697, opened 2022-05-25
chore(infra): amazon linux consolidated recipe

amolero-nr:
Using the global recipe `vars` section shown below to detect Amazon Linux version (either `2018.03`, `2`, `2022`) and upon that, the right repo folder (`el` for version `2018.03` or `amazonlinux` for versions `2` and `2022`):

```sh
  vars:
    AMAZON_LINUX_VERSION:
      sh: awk -F= '/VERSION_ID/ {print $2}' /etc/os-release
    REPO_DIR:
      sh: if [ {{.AMAZON_LINUX_VERSION}} = "2" ] || [ {{.AMAZON_LINUX_VERSION}} = "2022" ]; then echo "amazonlinux"; else echo "el"; fi
    ARCH:
      sh: uname -m
```

Used them all on the `assert_pre_req` and `install_infra` tasks with appropriate conditional logic to build/switch to the appropriate final URL repo and continue regular recipe flow.


katiehrenchir:
I think somebody else just recently renamed (in [this pr](https://github.com/newrelic/open-install-library/commit/efc97292462e298dc47f3d3caee295fd4a52d776) less than an hour ago) those recipes from `amazonlinux.yml` and `amazonlinux2.yml` -> `amazonlinux_1.yml` and `amazonlinux_2.yml`  - there are some old references to those in the test folder that should probably all be updated to `amazonlinux.yml` e.x. https://github.com/newrelic/open-install-library/blob/577ebcece3501948ed1851e241eb4526aa08d82b/test/definitions-eu/apm/java/rhl2-sysd-custom-sh.json#L39

amolero-nr:
Thank you @katiehrenchir! - Will take a look!

recipes/newrelic/infrastructure/awslinux.yml
```diff
@@ -7,6 +7,11 @@ description: New Relic install recipe for the Infrastructure agent
 repository: https://github.com/newrelic/infrastructure-agent
 
 installTargets:
+  - type: host
+    os: linux
+    platform: amazon
+    platformFamily: rhel
+    platformVersion: "2018.03"
```
Julien4218, line 14: @amolero-nr the amazonlinux_1 file didnt specify a platformVersion, hence cover all of them. Is it safe/good to make it more specific here? Perhaps we know the other versions are not supported by the infra agent?

amolero-nr: Thank you @Julien4218! - I thought to stick to that version and not prior ones after reading [this end-of-life/maintenance support notes from AWS](https://aws.amazon.com/blogs/aws/update-on-amazon-linux-ami-end-of-life/) but will remove that line for now. I have not tested if prior versions are supported  or not by the infra agent.

Julien4218 (approved):
This looks good @amolero-nr 

Please make sure the E2E tests are passing before releasing.

PR #709, opened 2022-06-07
gha validate recipe workflow test

amolero-nr:


amolero-nr:
This was a test and needs not to be merged.

PR #712, opened 2022-06-08
test validate recipe unit tests works with new awslinux.yml file

amolero-nr:


amolero-nr:
This was a test and needs not to be merged.

PR #715, opened 2022-06-09
Address increased error rate for NGINX recipe post-updates

katiehrenchir:
- adds back base nginx recipe changes to prompt for adding a status stub if it isn't already present
- switches to restart nginx using `sudo nginx -s reload` instead of relying on sysctl
- differentiates error messages so it's easier to pinpoint issues on the error dashboard
- fixes a scope issue when checking for the status stub after adding it which caused it to fail every time
- adds new error code 9 indicating the installation could not complete due to missing nginx configuration


recipes/newrelic/infrastructure/ohi/nginx/linux.yml
```diff
@@ -113,27 +125,73 @@ install:
...
+          if [ "$stub_location" == "" ]; then
+              echo -n "An nginx status stub location is required to proceed. Add default status location of /status? Y/N (default: Y)? "
+              read answer
+              echo ""
+              NEW_RELIC_CONTINUE=$(echo "${answer^^}" | cut -c1-1)
+              if [[ -z "$NEW_RELIC_CONTINUE" ]]; then
+                  NEW_RELIC_CONTINUE="Y"
+              fi
+              if [[ "$NEW_RELIC_CONTINUE" == "N" ]]; then
+                  echo "Exiting the installation code 130"
```
natecanfield822, line 137: Maybe a better message here? Something that indicates we stopped because the user selected not to let us add a status location that we needed. Maybe something like "Stopping installation: user opted not to allow automatic status stub page creation"

natecanfield822 (changes requested):
Look great! Just one small change to request

recipes/newrelic/infrastructure/ohi/nginx/linux.yml
katiehrenchir: ope, good catch, I think this was a debug print I left in. changed to your suggestion in this commit: https://github.com/newrelic/open-install-library/pull/715/commits/48ddf3b18f48c4d384a21a8e06aac72badadaae0

Julien4218:
This work is put on-hold and will be re-considered in another MMF in the near future.

PR #721, opened 2022-06-14
chore(install): update logs unsupported recipe and tests

NRhzhao:


recipes/newrelic/infrastructure/logs/linux-logs.yml
```diff
@@ -12,12 +12,24 @@ dependencies:
 installTargets:
   - type: host
```
natecanfield822, line 13: It looks like previously this target would have captured anything linux not caught by the more specific unsupported file - specifically in terms of explicitly supported systems, [SUSE/SLES 12](https://docs.newrelic.com/docs/logs/forward-logs/forward-your-logs-using-infrastructure-agent#requirements) - should an entry be added for that platform/version?

PR #726, opened 2022-06-21
Chore/troubleshooting

mbruzina:


katiehrenchir (approved):

.github/workflows/curl-spike.yml
```diff
@@ -0,0 +1,13 @@
...
+
+
+jobs:
+  which-curl:
+    name: Which curl
+    runs-on: ubuntu-latest
+    steps:
+      - name: check to see where curl lives
+        run: |
+          which curl
```
natecanfield822, line 13: Should we run this delegated to localhost as well? Maybe both with and without that attribute?

natecanfield822: Scratch that I was thinking of ansible YML, not an action

PR #727, opened 2022-06-21
chore(install): Add tests for Amazon linux 2022

NRhzhao:


recipes/newrelic/infrastructure/ohi/redis/rhel.yml
```diff
@@ -14,6 +14,11 @@ installTargets:
     os: linux
     platform: "amazon"
     platformVersion: "2"
+  - type: host
```
Julien4218, line 17: @NRhzhao you might want to check the other OHIs. I can see most of them are filtering on platformVersion: "2" as well, so perhaps the change below should be mirrored to all other OHIs?

NRhzhao: @Julien4218 Hi Julien, I'm going to back out the change for Redis. There is not much support for Amazon Linux 2022 right now, and I don't have time to research and set up tests for each OHI manually (Another priority came up).  
I will monitor our installations; once we get more requests for Amazon Linux 2022 for OS, we can pick up this change again for OHI.

Julien4218 (approved):

PR #729, opened 2022-06-22
chore(e2e): install curl as part of ansible role

mbruzina:


natecanfield822 (approved):
May need to change these three too if this works:
https://github.com/newrelic/open-install-library/blob/chore/ansible-install-curl/test/deploy/assertions/is-reporting/roles/onafterstart/tasks/main.yml
https://github.com/newrelic/open-install-library/blob/chore/ansible-install-curl/test/deploy/assertions/recipe-is-valid/roles/onafterstart/tasks/nrql_has_data.yml
https://github.com/newrelic/open-install-library/blob/chore/ansible-install-curl/test/deploy/assertions/recipe-is-valid/roles/onafterstart/tasks/main.yml

Julien4218 (approved):

PR #731, opened 2022-06-23
add k3s/k0s support for the Kubernetes recipe

bpschmitt:
Updated the install logic to include a check for `k3s` and `k0s` clusters.  Previously, this was failing for users even though Pixie supports both versions.


bpschmitt:
Closing in favor of https://github.com/newrelic/open-install-library/pull/741

PR #742, opened 2022-06-23
fix(ansible): moving curl install to prepare

mbruzina:


Julien4218 (approved):

Julien4218 (approved):

mbruzina:
Closing - need to revisit in the future.  This change was to fix an accidental removal of `curl` from demo-deployer images.  Some of the recipe tests use curl to validate installations and were failing for the command not being found; spirit was to install `curl` as part of the role that uses curl to validate instead of relying on the host having the installation.

PR #745, opened 2022-06-28
chore(e2e): update test to run in multiple steps

NRhzhao:


.github/workflows/nonregression-eu.yml
```diff
@@ -4,16 +4,138 @@ on:
   workflow_dispatch:
   schedule:
     - cron: "0 11 * * 1-5"
-  push:
```
Julien4218, line 7: @rthorn-nr why do we remove the run of the tests on push to main? Wouldn't we want to know that something pushed still passes the test?

.github/workflows/nonregression.yml
```diff
@@ -35,16 +158,36 @@ jobs:
...
+          set -e
+          testDefinitionFile=$(echo $MATRIX | jq -c -r '.testDefinitionFile')
+          echo $testDefinitionFile
           docker run -i\
             -v ${{ github.workspace }}/configs/:/mnt/deployer/configs/\
             -v ${{ github.workspace }}/test/:/mnt/deployer/test/\
-            --entrypoint ruby ghcr.io/newrelic/deployer:latest batch.rb -c configs/gitusdkr${{ github.run_id }}.json -d test/definitions/ -s 20 -m deploy -l debug
+            --entrypoint ruby ghcr.io/newrelic/deployer:latest main.rb -c configs/gitusdkr${{ github.run_id }}.json -d $testDefinitionFile -l debug
+          echo ::set-output name=exit_status::$?
+      - name: Teardown any previous deployment
```
Julien4218, line 172: It would be great to see how we could offload the teardown to another GH run, for example this is the job that clean the resources after running the regression tests https://github.com/newrelic/open-install-library/blob/main/.github/workflows/cleanup.yml
This would speed up the tests further. This can be done outside the scope of this PR though.

.github/workflows/nonregression-eu.yml
Julien4218: Never mind, I see the eu-smoke is doing the build on push to main.

Julien4218 (approved):

.github/workflows/nonregression-eu.yml
rthorn-nr: @Julien4218 my proposal is that we can run the `nonregression-eu-smoke` workflow on push to main and still run full regression testing daily at its regularly-scheduled time. This is only because the EU tests have been flaky when too many of them are running simultaneously, even with `max-parallel` set to 30.

I'm happy to delete the EU smoke test workflow and run full EU regression instead, but it will take longer to run, especially since the smoke tests are run last (just due to the directory structure -- definitely can change the script to run them first, which may speed that workflow up)

.github/workflows/nonregression.yml
rthorn-nr: Agreed! I would be happy to take a look at using the cleanup/-eu workflows rather than triggering teardown in each test individually in another PR.

PR #750, opened 2022-07-01
Add logic to revert to kubectl install when helm 2 is detected

bpschmitt:
This update adds logic that will detect the user's Helm client version.  If Helm 3 is found, Helm will be used to perform the install.  If Helm 2 is found (or Helm is not present at all), then the install will revert to using `kubectl` as the install client.  

In addition to detecting the Helm version, if `kubectl` is used for install and Pixie is selected for install, the guided install will now apply the Pixie CRDs prior to applying the New Relic manifest.  Previously, this did not happen and was generating errors.

### DIFF
--- a/recipes/newrelic/infrastructure/kubernetes.yml
+++ b/recipes/newrelic/infrastructure/kubernetes.yml
@@ -352,7 +352,6 @@ install:
                   echo "Helm not found."
                   echo "Reverting to kubectl install."
                   echo ""
-                  KUBECTL_INSTALL=true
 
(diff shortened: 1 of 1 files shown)


bpschmitt:
One other minor update.  Previously, the `newrelic-k8s.yml` manifest output would end up in `/tmp`.  I've moved this to `$HOME` instead as well as updated the message to help the user find the file.

amolero-nr (approved):

amolero-nr (approved):

PR #751, opened 2022-07-07
add nri-metadata-injection job cleanup to K8s recipe

bpschmitt:
This change adds an if statement that checks for existing `nri-metadata-injection` jobs in the cluster.  If they exist, they will be deleted before proceeding with the install.  This has caused installation conflicts for a number of users and should provide a better experience.

### DIFF
--- a/recipes/newrelic/infrastructure/kubernetes.yml
+++ b/recipes/newrelic/infrastructure/kubernetes.yml
@@ -475,9 +475,9 @@ install:
             fi
 
             # Clean up old nri-metadata-injection jobs if they exist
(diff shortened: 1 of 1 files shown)


NRhzhao (approved):

thakerb:
I noticed this error message in New Relic installation dashboard for an installation done on Nov 02, 2022. Could it be related to this code change because the code change also involved newrelic-k8s.yml? 
execution failed for kubernetes-open-source-integration: exit status 1: open /tmp/newrelic-k8s.yml: permission denied
cc: @bpschmitt @NRhzhao

PR #757, opened 2022-07-13
Changes linux package name from newrelic-netcore20-agent to newrelic-dotnet-agent

vuqtran88:
## Description:
The .Net Agent is going to release the major version of the agent. Part of the major release, we are going to rename a lot of our installer names as well as package names to remove confusion as well as better reflect what .NET frameworks are supported in our new Agent. Because of this change, we need to update our recipes to use the new package name. This PR changes using linux package name from `newrelic-netcore20-agent` to `newrelic-dotnet-agent`.

~~Note: Since the linux packages with the new name will only be available after we release the major version agent, this PR should only be merged after the .NET team releases the major version to avoid breaking. This is also why this PR is labeled as DRAFT.~~


JcolemanNR (approved):
Looks correct to me.

Julien4218 (approved):
Looks good

PR #759, opened 2022-07-15
Mongodb Recipe Enhancements

amolero-nr:
[WIP] This PR is still a `work in progress` as I still have pending: 

- a few pending `TODO`s noted in the code, 
- some clean up or possible refactoring, 
- prompts styling updates, 
- an important update in the use cases flow in the program that I think will make things clearer for the user (regular mongodb creds vs ssl enabled mongodb cases),
- etc.

For those that want to test a single node MongoDB (without authentication enabled), you can use demo deployer with something like this:
```sh
ruby main.rb -c configs/credentials.json -d open-install-library/test/manual/definitions/ohi/linux/mongodb-debian.json
```

Once your instance/mongodb is up, go to mongodb's config file at `/etc/mongod.conf` and comment out these 2 lines:
```sh
replication:
  replSetName: test
```

Then restart mongodb: `sudo systemctl restart mongod` and run the MongoDB integration install.

### DIFF
--- a/recipes/newrelic/infrastructure/ohi/sql/ms-sql.yml
+++ b/recipes/newrelic/infrastructure/ohi/sql/ms-sql.yml
@@ -131,7 +131,7 @@ install:
           $NR_CLI_DB_PASSWORD=   "$env:NR_CLI_DB_PASSWORD"
           if ([string]::IsNullOrWhiteSpace($NR_CLI_DB_PASSWORD)) {
             Add-Type -AssemblyName System.Web;
-            $NR_CLI_DB_PASSWORD = [System.Web.Security.Membership]::GeneratePassword(20,2).replace('-','');
+            $NR_CLI_DB_PASSWORD = [System.Web.Security.Membership]::GeneratePassword(20,2).replace("-","");
           }
 
           $OhiConfig = "C:\\Program Files\\New Relic\\newrelic-infra\\integrations.d\\mssql-config.yml";


recipes/newrelic/infrastructure/ohi/mongodb/debian.yml
```diff
@@ -62,24 +59,344 @@ preInstall:
...
+            IS_TOOL_INSTALLED=$(which ${tool} | wc -l)
+            if [ "$IS_TOOL_INSTALLED" -eq 0 ]
+            then
+              echo -e "{{.RED}}This installation requires '${tool}' to be installed.{{.NOCOLOR}}" >> /dev/stderr
+              exit ${code}
+            fi
+          done
+        - | 
+          IS_MONGO_SHELL_INSTALLED=$(which mongo | wc -l)
+          if [ $IS_MONGO_SHELL_INSTALLED -eq 0 ]; then
```
Julien4218, line 108: @amolero-nr this might need to go in the pre-req. If we cant create a username/password we wont be able to install. We should still detect though.

recipes/newrelic/infrastructure/ohi/mongodb/debian.yml
```diff
@@ -62,24 +59,344 @@ preInstall:
...
+          tee -a /tmp/mongo-role.js > /dev/null <<-"EOT"
+            use admin
+            db.createRole({ role: "listCollections", privileges: [{ resource: { db: "", collection: "" }, actions: [ "listCollections" ] }], roles: [] })
+          EOT
+          if [ -f /tmp/mongo-user.js ]; then
+            rm -f /tmp/mongo-user.js 
+          fi
+          tee -a /tmp/mongo-user.js > /dev/null <<-"EOT"
+            use admin
+            db.createUser({ user: "MONGO_MONITORING_USERNAME", pwd: "MONGO_MONITORING_PASSWORD", roles: [ "clusterMonitor", "listCollections" ] })
```
Julien4218, line 128: @amolero-nr I assume the password will be different? It should either auto-generated or injected through env var.

recipes/newrelic/infrastructure/ohi/mongodb/debian.yml
```diff
@@ -62,24 +59,344 @@ preInstall:
...
+          EOT
+      
+    assert_input:
+      cmds:
+        - |
+          NEW_RELIC_ASSUME_YES="{{.NEW_RELIC_ASSUME_YES}}" # TODO: Implement NEW_RELIC_ASSUME_YES
+          NR_CLI_DB_HOSTNAME="{{.NR_CLI_DB_HOSTNAME}}"
```
Julien4218, line 135: For all the env vars, we should allow them to be passed in explicitly when running the CLI.
Also we should default appropriately when not specified. NR_CLI_DB_HOSTNAME should be localhost as we dont support installing on remote host with this installation.

recipes/newrelic/infrastructure/ohi/mongodb/debian.yml
```diff
@@ -62,24 +59,344 @@ preInstall:
...
+          NR_CLI_CLIENT_CERT_FILE="{{.NR_CLI_CLIENT_CERT_FILE}}"
+          NR_CLI_CERT_AUTH_FILE="{{.NR_CLI_CERT_AUTH_FILE}}"
+
+          # MongoDB defaults
+          NR_CLI_DB_HOSTNAME=${NR_CLI_DB_HOSTNAME:-localhost}
+          NR_CLI_DB_PORT=${NR_CLI_DB_PORT:-27017}
+          NR_CLI_DB_CLUSTERNAME=${NR_CLI_DB_CLUSTERNAME:-mongocluster}
+          NR_CLI_DB_AUTH=${NR_CLI_DB_AUTH:-admin}
+
+          MONGO_MONITORING_USERNAME=newrelic
```
Julien4218, line 151: env var should be configurable from env var too

recipes/newrelic/infrastructure/ohi/mongodb/debian.yml
```diff
@@ -62,24 +59,344 @@ preInstall:
...
+            rm /etc/newrelic-infra/integrations.d/mongodb-config.yml;
+          fi
+          tee /etc/newrelic-infra/integrations.d/mongodb-config.yml > /dev/null <<-"EOT"
+          integrations:
+            - name: nri-mongodb
+              env:
+                # The mongos to connect to
+                HOST: NR_CLI_DB_HOSTNAME
+                # The port the mongos is running on
+                PORT: NR_CLI_DB_PORT
```
Julien4218, line 172: Can we validate the port before installing?

Julien4218 (commented):
@amolero-nr it looks like this is work in progress, which is great.
You might consider having 1 file for both debian and rhel, to avoid a lot of duplications, since the recipe is quite long.

recipes/newrelic/infrastructure/ohi/mongodb/debian.yml
amolero-nr: Ok

recipes/newrelic/infrastructure/ohi/mongodb/debian.yml
amolero-nr: Yes, password will be different each time and is now injected through env var.

recipes/newrelic/infrastructure/ohi/mongodb/debian.yml
amolero-nr: This was corrected.

recipes/newrelic/infrastructure/ohi/mongodb/debian.yml
amolero-nr: Will add to my to do.

PR #764, opened 2022-07-26
Increase max-parallel in nonregression-eu workflow

rthorn-nr:


Julien4218 (approved):
let's try this @rthorn-nr !

PR #765, opened 2022-07-26
Fix raw repo regex in unit test workflow

rthorn-nr:
Fixes the `raw.githubusercontent.com` branch replacement regex to match each URL in a string of multiple URLs rather than just a single URL. This was fixed in the nonregression/-eu workflows but not ported to the validation workflow. 

Before: https://regexr.com/6qjgq
After: https://regexr.com/6qjh3


Julien4218 (approved):
Thanks for providing the regex URL to validate, much easier.

PR #768, opened 2022-07-28
Validate SQLServer NR installation in OHI test

rthorn-nr:


NRhzhao (approved):

Julien4218 (approved):
@rthorn-nr this looks good. I think quite a few other tests might also need similar update.
Do we have a separate ticket @NRhzhao

PR #769, opened 2022-07-29
chore(install): added metadata for mysql

NRhzhao:


recipes/newrelic/infrastructure/ohi/mysql/debian.yml
```diff
@@ -265,7 +266,13 @@ install:
         IS_INITCTL:
           sh: command -v initctl | wc -l
 
-postInstall:
+    collect_meta:
+      cmds:
+        - |
+          user=$(whoami)
+          version=$(mysqld --version)
+          echo {\"Metadata\": {\"user\":\"$user\", \"key2\":\"$version\"}} | tee {{.NR_CLI_OUTPUT}} > /dev/null
```
Julien4218, line 274: @NRhzhao maybe we should name "key2" something more specific? Perhaps just "version" or "mysqlVersion"?

NRhzhao: @Julien4218 Good catch, it was suppose to be just version. Fixing

Julien4218 (approved):

Julien4218:
This looks good.

PR #777, opened 2022-08-23
chore(Memcached): Memchached recipe simplefied

NRhzhao:


NRhzhao:
1. Removing the check for the port number. We see an error after it was introduced, and I don't think it's helpful for most of the use cases.
2. Add try to all package updates fail to continue the install
3. Add single quotes around the password in the config file to allow escape characters

amolero-nr (approved):

PR #785, opened 2022-08-30
update infra agent key path

rogercoll:
As part of a new strategy to rotate packages GPG keys, the path of the current key will be changed. We still maintain the path of the previous key, but from now on, all the packages will be pointed to `infrastructure_agent/keys/` bucket path.

See https://download.newrelic.com/infrastructure_agent/keys/

Let us know if you need further information


amolero-nr:
Closing this PR as changes will be implemented through https://github.com/newrelic/open-install-library/pull/786 instead.

PR #789, opened 2022-09-13
chore(windows): Collect host window version with metadata infra

NRhzhao:


NRhzhao:
Will write window info on the meta file and be sent to need storage:

image

Julien4218 (approved):

PR #791, opened 2022-09-15
Adds a precheck script to .NET Linux installs to speed up detection

jaffinito:
- Updates preInstall-requireAtDiscovery to check if any of the three main .net packages are installed and if none are installed, exit out with status code 3.
- The new script will check both dpkg (Debian) and yum (RHEL-RPM) for the packages and record if the command was successful or not.
- Failed checks will report as a non-0 status code while success is 0.
- The if-then at the end of the check will examine all the status codes and if none are 0, we know that .NET was not installed.  If any are 0, we found .NET and should continue with the introspector.
- Tested locally on Ubuntu and RockyLinux (Centos)

### DIFF
--- a/recipes/newrelic/apm/dotNet/linux-systemd.yml
+++ b/recipes/newrelic/apm/dotNet/linux-systemd.yml
@@ -22,7 +22,45 @@ processMatch: []
 
 preInstall:
   requireAtDiscovery: |
-    # Get .NET introspector tooling
+    # Check if .NET is installed
+    # There are three packages for .NET: dotnet-runtime, aspnetcore-runtime and dotnet-sdk
+    # Checking both debian and rpm based distros together
+    
+    set +e # disable fail on error
+    
+    dpkg -l dotnet-runtime* >/dev/null 2>&1
+    DPKG_DOTNET=$?
+    
+    dpkg -l aspnetcore-runtime* >/dev/null 2>&1
+    DPKG_ASPNETCORE=$?
+
+    dpkg -l dotnet-sdk* >/dev/null 2>&1
+    DPKG_DOTNET_SDK=$?
+    
+    yum -q list installed dotnet-runtime* >/dev/null 2>&1
+    YUM_DOTNET=$?
+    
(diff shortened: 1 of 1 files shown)


recipes/newrelic/apm/dotNet/linux-systemd.yml
```diff
@@ -22,7 +22,45 @@ processMatch: []
...
 preInstall:
   requireAtDiscovery: |
-    # Get .NET introspector tooling
+    # Check if .NET is installed
+    # There are three packages for .NET: dotnet-runtime, aspnetcore-runtime and dotnet-sdk
+    # Checking both debian and rpm based distros together
+    
+    set +e # disable fail on error
+    
+    dpkg -l dotnet-runtime* >/dev/null 2>&1
```
Julien4218, line 31: @jaffinito this is a great idea. could we combine all the dpkg queries into a single call? I notice sometime that invoking a package list is not speedy, I found this to work but feel free to modify as needed "dpkg -l apt notexist notexist2 apt anothernot | grep 'apt\|notexist2\|anothernot' | wc -l" and for rhel "sudo yum -q list installed sudo.* systemd.* notexist | grep 'sudo\|notexist' | wc -l"

jaffinito: Nice! Let me run some tests to see how that works.

JcolemanNR:
Note that this will miss use cases where a .NET application is deployed as a self contained executable. If we are fine with some false negatives (virtuoso doesn't install .NET agent even though there are .NET apps) in this case, then I guess it's ok

recipes/newrelic/apm/dotNet/linux-systemd.yml
Julien4218: Hello @jaffinito just checking to see if you have some time to look into this. Let me know if you need anything. Thank You

Julien4218 (approved):
This looks good thanks @jaffinito

PR #795, opened 2022-10-04
feat(infra): add process-based validation

amolero-nr:
This is to address [NR-53396](https://issues.newrelic.com/browse/NR-53396) which is about adding a second `infrastructure-agent` validation in case it is running on a different port. 

Changes tested in RHEL, Ubuntu, Debian, AWS Linux, Windows.

Testing: https://issues.newrelic.com/browse/NR-55799

### DIFF
--- a/recipes/newrelic/infrastructure/ubuntu.yml
+++ b/recipes/newrelic/infrastructure/ubuntu.yml
@@ -239,6 +239,12 @@ install:
     assert_agent_status_ok:
       cmds:
         - |
+          infra_present(){
+            INFRA_PRESENT=$(ps aux | grep newrelic-infra$ | wc -l)
+            [[ $INFRA_PRESENT -gt 0 ]]
+            return
+          }
+
           MAX_RETRIES=150
           TRIES=0
           echo "Running agent status check attempt..."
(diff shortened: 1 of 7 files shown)


Julien4218:
@amolero-nr this might works yes. It looks like after doing all the attempt to validate with the infra URL, we'll check the process is started and will delegate to the NRQL validation for the final check.

recipes/newrelic/infrastructure/ubuntu.yml
```diff
@@ -261,7 +261,7 @@ install:
                   # Process-based validation attempt
                   if infra_present; then
                     echo "detected newrelic-infra process running"
-                    break
+                    exit 0
```
amolero-nr, line 264: It seems, the `break` statement is not breaking out of the while loop. I do not like using an `exit 0` but it works fine in my tests (and Pavan's). Please: Is using `exit 0` here acceptable in your opinion @Julien4218, @NRhzhao?

image

Julien4218: @amolero-nr seems fine to me, since we've validated infra, I don't see what else is needed to be done in the recipe?
If you do, make sure you make the change in all the recipes.

amolero-nr: Thank you!  Will update the rest of them with this fix.

Julien4218 (approved):
This looks good @amolero-nr

PR #797, opened 2022-10-10
fix(postgres): add support for rhel family

NRhzhao:


NRhzhao:
- Update to prompt for username/password if default postgres user doesn't work
- Allow passing in env for username/password for automatic install

rthorn-nr (approved):

PR #800, opened 2022-10-12
chore(sql): Allow pass in SQL credentials

NRhzhao:


NRhzhao:
Install for SQL recipe fails when default login credential fails. This change allows the user to pass in credential instead of just relying on the default
- If credentials are passed in, we will try run NR script with those first. 
- If it fails or no credential is passed in, it will try to run with default login credentials (existing behavior)
- Part of the effort to troubleshoot SQL 2014 CLI install

Julien4218 (approved):

PR #811, opened 2022-11-02
feat(capture-logs): removed go-task vars in favor of recipe metadata 

mbruzina:
Instead of using CAPTURE_CLI_LOGS to trigger which recipes should capture for potential log collection, use our recipe metadata functionality instead.

We're adding more metadata around our log capturing opt-in functionality, and having both of these flags within our recipe's `Metadata` collection is a cleaner implementation.

This library change should be released with the corresponding newrelic-cli change from PR: https://github.com/newrelic/newrelic-cli/pull/1386

### DIFF
--- a/recipes/newrelic/infrastructure/awslinux.yml
+++ b/recipes/newrelic/infrastructure/awslinux.yml
@@ -71,12 +71,12 @@ install:
       sh: if [ {{.AMAZON_LINUX_VERSION}} = "2" ] || [ {{.AMAZON_LINUX_VERSION}} = "2022" ]; then echo "amazonlinux"; else echo "el"; fi
     ARCH:
       sh: uname -m
-    CAPTURE_CLI_LOGS: true
 
 
   tasks:
     default:
       cmds:
+        - task: write_recipe_metadata
         - task: assert_pre_req
         - task: cleanup
         - task: setup_license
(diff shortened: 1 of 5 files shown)


NRhzhao:
1. We should silence the error; this might fail if somehow the file gets locked up and stops the installation. (Should be an optional step)
`echo {\"Metadata\": {\"user\":\"$user\", \"version\":\"$version\"}} | tee {{.NR_CLI_OUTPUT}} > /dev/null`
2. We also use the same file on the other recipes [example](https://github.com/newrelic/open-install-library/blob/a8fa80f8e05508ce1ec543199f83ec5a80843604/recipes/newrelic/infrastructure/ohi/mysql/rhel.yml#L273)
Is that going to override this setting?

mbruzina:
> 1. We should silence the error; this might fail if somehow the file gets locked up and stops the installation. (Should be an optional step)
>    `echo {\"Metadata\": {\"user\":\"$user\", \"version\":\"$version\"}} | tee {{.NR_CLI_OUTPUT}} > /dev/null`

Got it, I can add this

> 2. We also use the same file on the other recipes [example](https://github.com/newrelic/open-install-library/blob/a8fa80f8e05508ce1ec543199f83ec5a80843604/recipes/newrelic/infrastructure/ohi/mysql/rhel.yml#L273)
>    Is that going to override this setting?

It shouldn't.  Each recipe execution has its own execution context in which we create these files, and we build the temp files using the recipe name.  For example, when running our infra recipes locally the cli output file is

`/var/folders/kq/llpxws8x4zvgn9ppq4b2r83m0000gn/T/infrastructure-agent-installer_out4017780260`

This file removal is removed after the given recipe's execution has completed/failed/etc (it's deferred)

mbruzina:
@NRhzhao updated metadata writing in linux infra recipes:
`echo '{"Metadata":{"CapturedCliOutput":"true"}}' | tee {{.NR_CLI_OUTPUT}} > /dev/null`

NRhzhao (approved):

amolero-nr (approved):

PR #823, opened 2022-11-18
chore: adding manual deploy config to create an empty eks cluster

Julien4218:
This PR is dependent on this demo-deployer PR to add the capability to associate an instance profile to an EC2 instance, so the host can have the proper permission to create an EKS cluster.
https://github.com/newrelic/demo-deployer/pull/104

JIRA https://issues.newrelic.com/browse/NR-67454


rthorn-nr (approved):
Tested SockShop deployment as well, very streamlined!

PR #829, opened 2022-11-29
fix: error when writing to existing file

svetlanabrennan:
This pr fixes this error [caused by this line](https://github.com/newrelic/open-install-library/blob/main/recipes/newrelic/infrastructure/kubernetes.yml#L500):

```
execution failed for kubernetes-open-source-integration: exit status 1: open /tmp/newrelic-k8s.yml: permission denied
```

If this script is re-ran by the user, this error happens because it can't overwrite the existing file. This pr ensures that we remove the file if it exists and then write to it.

### DIFF
--- a/recipes/newrelic/infrastructure/kubernetes.yml
+++ b/recipes/newrelic/infrastructure/kubernetes.yml
@@ -497,6 +497,7 @@ install:
             fi
 
             BODY="${BODY}}"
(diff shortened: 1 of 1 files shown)


Julien4218 (approved):
@svetlanabrennan I tested and saw no regression on an EKS cluster.

PR #830, opened 2022-11-30
chore: adding min and max nodes param when creating cluster

Julien4218:


rthorn-nr (approved):

rthorn-nr:
@jerard the k8-linux unit test was failing due to the unsupported k8s version 1.25. Figured we needed to get the changes from main into this branch, so I rebased and pushed. Going to run one more manual test before approving

rthorn-nr (approved):

PR #831, opened 2022-11-30
Use K8s v1.24.8 for non-regression testing

rthorn-nr:


Julien4218 (approved):
@rthorn-nr this looks good. We should probably also mirror this change in the manual section https://github.com/newrelic/open-install-library/blob/main/test/manual/definitions/kubernetes/minikube-empty.json

PR #836, opened 2022-12-05
fix(alert): Log error message instead of just code

NRhzhao:
1. Update discovery to search for the specific alert to avoid paging issues when the user has too many policies
2. Update to log error message instead of just exit code if the install fails. Examples of error when triggered failure on purpose.
image
image

### DIFF
--- a/recipes/newrelic/infrastructure/alerts/golden.yml
+++ b/recipes/newrelic/infrastructure/alerts/golden.yml
@@ -37,13 +37,12 @@ preInstall:
       POLICY_RESULT=$(curl -sX POST $NEW_RELIC_API_URL'/graphql' \
           -H "Api-Key:$NEW_RELIC_API_KEY" \
           -L -H 'Content-Type: application/json' \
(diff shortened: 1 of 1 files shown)


recipes/newrelic/infrastructure/alerts/golden.yml
```diff
@@ -86,6 +85,14 @@ install:
               -L -H 'Content-Type: application/json' \
               -d @/tmp/policy.json
           )
+
+          # Check if got error while try to create policy
+          POLICY_RESULT_ERR=$(echo $POLICY_RESULT | /usr/local/bin/newrelic utils jq '.|select(.errors != null)|.errors[0].message ')
+          if [[ ! -z "$POLICY_RESULT_ERR" ]]; then
+            >&2 echo 'Could not create a new alert policy for {{.ALERT_POLICY_NAME}} got the following error:' "$POLICY_RESULT_ERR"
```
Julien4218, line 92: @NRhzhao I think if we redirect to stderr we're actually not getting to see the error message. Given this makes little difference, I think it would be better to display in stdout, instead of trying to send to the error stream.
There are other similar redirect below.

Julien4218 (approved):
looks good, thanks @NRhzhao

PR #840, opened 2022-12-08
Add option to allow user to uninstall a previously installed kubernetes integration 

csongnr:
i tested:  
- no existing k8s integration prior => prompt is skipped 
- passing in a weird answer to script => all answers that start w/ y or n will be interpreted as a yes or no, no answer (as in just hit enter) defaults as a yes, and weird answers get prompted w/ the question again. 
- passing weird answer to script and then acceptable answer => uninstall skips or attempts based on acceptable answer 
- saying no while existing k8s integration under different namespace => install attempts, but errors out as expected
- saying no while existing k8s integration under same namespace => install attempts, doesn't error out (expected, used to work this way before too depending on some specifics.) 
- saying yes while existing k8s integration under different namespace  => uninstall completes, install then succeeds
- saying yes while existing k8s integration under same namespace => uninstall completes, install then succeeds
- completely different helm release in other namespace => uninstall completes correct uninstall, leaves alone irrelevant release, install then succeeds

### DIFF
--- a/recipes/newrelic/infrastructure/kubernetes.yml
+++ b/recipes/newrelic/infrastructure/kubernetes.yml
@@ -393,6 +393,30 @@ install:
 
             KUBECTL_INSTALL=false
 
+            # find if nri bundle already installed, ask customer if they wanna uninstall, and then move forward 
+
+            GREEN='\033[0;32m'
+            NC='\033[0m'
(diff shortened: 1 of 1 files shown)


mbruzina:
This PR's functionality was merged in https://github.com/newrelic/open-install-library/pull/841

PR #841, opened 2022-12-08
feat(k8s): add uninstall option

mbruzina:
Branched off of `cs/uninstall-option` to test failing golden signal test on PR.  
Seems there was a mismatch in tests on main and the kubernetes branch was behind.  Was able to rebase main and golden signals was no longer detected (which is correct)

Also wrapped the functionality added by @csongnr to respect `-y`/`--assumeYes` flags on guided install

### DIFF
--- a/recipes/newrelic/infrastructure/kubernetes.yml
+++ b/recipes/newrelic/infrastructure/kubernetes.yml
@@ -393,6 +393,36 @@ install:
 
             KUBECTL_INSTALL=false
 
+            # Check if nri-bundle already installed, ask customer if they want uninstall before re-installation, and then move forward.
+            GREEN='\033[0;32m'
+            NC='\033[0m'
+            if $SUDO helm list --all-namespaces | grep nri-bundle > /dev/null 2>&1; then
+              while :; do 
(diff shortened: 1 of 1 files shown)


mbruzina:
Simply running unit tests...

csongnr (approved):

csongnr:
adding the testing comments here for visibility: 
i tested:

- no existing k8s integration prior => prompt is skipped
- passing in a weird answer to script => all answers that start w/ y or n will be interpreted as a yes or no, no answer (as in just hit enter) defaults as a yes, and weird answers get prompted w/ the question again.
- passing weird answer to script and then acceptable answer => uninstall skips or attempts based on acceptable answer
- saying no while existing k8s integration under different namespace => install attempts, but errors out as expected
- saying no while existing k8s integration under same namespace => install attempts, doesn't error out (expected, used to work this way before too depending on some specifics.)
- saying yes while existing k8s integration under different namespace => uninstall completes, install then succeeds
- saying yes while existing k8s integration under same namespace => uninstall completes, install then succeeds
- completely different helm release in other namespace => uninstall completes correct uninstall, leaves alone irrelevant release, install then succeeds

PR #848, opened 2022-12-14
feat(k8s): capture additional metadata with user actions

svetlanabrennan:
Need to capture if the user chose to uninstall or skip uninstalling nri-bundle (this question is only prompted if the bundle was already installed). 

Also capturing the helm upgrade command before it's executed to help us troubleshoot errors with this command if there's an error with helm upgrade args.

### DIFF
--- a/recipes/newrelic/infrastructure/kubernetes.yml
+++ b/recipes/newrelic/infrastructure/kubernetes.yml
@@ -411,11 +411,13 @@ install:
                 fi
                 if [[ "$UNINSTALL_ANSWER" == "N" ]]; then
                   echo "Attempting install over existing integration."
+                  echo "{\"Metadata\":{\"uninstallNriBundleAnswer\":\"No\"}}" | tee -a {{.NR_CLI_OUTPUT}} > /dev/null
                   break 
                 fi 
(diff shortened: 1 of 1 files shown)


recipes/newrelic/infrastructure/kubernetes.yml
```diff
@@ -411,11 +411,13 @@ install:
                 fi
                 if [[ "$UNINSTALL_ANSWER" == "N" ]]; then
                   echo "Attempting install over existing integration."
+                  echo "{\"Metadata\":{\"uninstallNriBundleAnswer\":\"No\"}}" | tee -a {{.NR_CLI_OUTPUT}} > /dev/null
```
csongnr, line 414: can you add metadata for "PreviousNRIBundleDetected" : "yes" before the while loop in this block, and then edit this block to have an `else` statement at the end that adds  "PreviousNRIBundleDetected" : "no"?

svetlanabrennan: Done

csongnr (approved):

noahmmcgivern (approved):

PR #849, opened 2023-01-06
Fix smoke/mysql-rhel

rthorn-nr:
One of our smoke tests `smoke/mysql-rhel.json` has been failing consistently when downloading the mysql community release rpm file, either with a read timeout or a null error `'NoneType' object is not callable`. 

Using `get_url` instead of `yum` for the download uncovered a 403 when trying to download the file. Tried again with `wget`, and the issue was resolved. 

Other changes:
- become: yes -> true (to resolve linting warnings)
- increased lock timeout for yum rpm install

### DIFF
--- a/test/definitions/smoke/mysql-rhel.json
+++ b/test/definitions/smoke/mysql-rhel.json
@@ -1,44 +1,49 @@
 {
-    "global_tags": {
-        "owning_team": "virtuoso",
-        "Environment": "development",
-        "Department": "product",
-        "Product": "virtuoso"
-    },
-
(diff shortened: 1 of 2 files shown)


Julien4218 (commented):
@rthorn-nr this looks good, however I'm not seeing the impacted E2E tests running as part of this PR. I believe they need either a modification in the recipe file, or the test file. Could you do one of those so that we can see if the tests are passing with this change?

Julien4218 (approved):
Looks great, thanks @rthorn-nr

PR #851, opened 2023-01-10
Move tests with OHI validation timeout to separate workflow

rthorn-nr:
Creates a separate workflow (`ohi-validation-timeout.yml`) for the OHI tests in the nonregression workflow which are failing due to a validation timeout and filters those same tests out of the nonregression workflow:

```
const outputTestFilesMap = testDefinitions
  .filter((testDefinitionFile) => !isOHIValidationTimeout(testDefinitionFile))
```

For any other tests that need to be added to the workflow, add the path to the test (starting from the project root) to the array in `.github/workflows/scripts/ohiValidationTimeout.js`.

Other changes are due to autoformatting

### DIFF
--- a/.github/workflows/ohi-validation-timeout.yml
+++ b/.github/workflows/ohi-validation-timeout.yml
@@ -1,11 +1,6 @@
-name: Non Regression Testing
+name: OHI Validation Timeout
 
-on:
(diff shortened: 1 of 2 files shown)


Julien4218 (approved):
looks good

PR #855, opened 2023-01-18
fix: catch command errors

svetlanabrennan:
Catch command errors for helm add, helm update and kubectl apply

### DIFF
--- a/recipes/newrelic/infrastructure/kubernetes.yml
+++ b/recipes/newrelic/infrastructure/kubernetes.yml
@@ -428,8 +428,15 @@ install:
               echo "{\"Metadata\":{\"previousNriBundleDetected\":\"No\"}}" | tee -a {{.NR_CLI_OUTPUT}} > /dev/null
             fi
 
-            $SUDO helm repo add newrelic https://helm-charts.newrelic.com
-            $SUDO helm repo update
+            if [[ $($SUDO helm repo add newrelic https://helm-charts.newrelic.com) ]]; then
+            else
+              exit 131
+            fi
+
+            if [[ $($SUDO helm repo update) ]] ; then
+            else
+              exit 131
+            fi
 
             ARGS="--install newrelic-bundle newrelic/nri-bundle"
             ARGS="${ARGS} --set global.licenseKey=${NEW_RELIC_LICENSE_KEY}"
--- a/recipes/newrelic/infrastructure/kubernetes.yml
+++ b/recipes/newrelic/infrastructure/kubernetes.yml
@@ -562,7 +562,10 @@ install:
               done
             fi
 
-            $SUDO $KUBECTL apply -f $KUBECTL_YAML_DIR/newrelic-k8s.yml
+            if [[ $($SUDO $KUBECTL apply -f $KUBECTL_YAML_DIR/newrelic-k8s.yml) ]]; then
+            else
+              exit 131
+            fi
           fi
 
           HELM_COMMAND="helm not used"


recipes/newrelic/infrastructure/kubernetes.yml
```diff
@@ -428,8 +428,15 @@ install:
               echo "{\"Metadata\":{\"previousNriBundleDetected\":\"No\"}}" | tee -a {{.NR_CLI_OUTPUT}} > /dev/null
             fi
 
-            $SUDO helm repo add newrelic https://helm-charts.newrelic.com
-            $SUDO helm repo update
+            if [[ $($SUDO helm repo add newrelic https://helm-charts.newrelic.com) ]]; then
+            else
+              exit 131
```
thakerb, line 433: I would suggest adding an echo "ERROR = $? from $SUDO helm repo add newrelic https://helm-charts.newrelic.com" on the failure path so that we can understand the reason for the failure.

recipes/newrelic/infrastructure/kubernetes.yml
```diff
@@ -428,8 +428,15 @@ install:
...
+            fi
+
+            if [[ $($SUDO helm repo update) ]] ; then
+            else
+              exit 131
```
thakerb, line 438: I would suggest adding an echo "ERROR = $? from  $SUDO helm repo update" on the failure path so that we can understand the reason for the failure.

recipes/newrelic/infrastructure/kubernetes.yml
```diff
@@ -555,7 +562,10 @@ install:
               done
             fi
 
-            $SUDO $KUBECTL apply -f $KUBECTL_YAML_DIR/newrelic-k8s.yml
+            if [[ $($SUDO $KUBECTL apply -f $KUBECTL_YAML_DIR/newrelic-k8s.yml) ]]; then
+            else
+              exit 131
```
thakerb, line 567: I would suggest adding an echo "ERROR = $? from $SUDO $KUBECTL apply -f $KUBECTL_YAML_DIR/newrelic-k8s.yml" on the failure path so that we can understand the reason for the failure.

recipes/newrelic/infrastructure/kubernetes.yml
svetlanabrennan: @thakerb 
If we add a custom echo like you suggested, then the user (and our team) won't see the exact helm error message in the terminal and in our dashboard (where we track our errors). We will just see that there was an error when "helm repo add" command was ran. 

Example errors:
- custom echo error message:
```
ERROR 1 returned from: helm repo add newrelic https://helm-charts.newrelict.com/
```

- letting the cli capture the helm error message
```
Error: looks like "https://helm-charts.newrelict.com/" is not a valid chart repository or cannot be reached: Get "https://helm-charts.newrelict.com/index.yaml": dial tcp: lookup helm-charts.newrelict.com on 10.1.241.220:53: no such host
```

That's why I chose to go with letting the CLI capture the helm error. Thoughts?

recipes/newrelic/infrastructure/kubernetes.yml
svetlanabrennan: Same reasoning for choosing to not set a custom echo as I mentioned in the comment above.

recipes/newrelic/infrastructure/kubernetes.yml
svetlanabrennan: Same reasoning for choosing to not set a custom echo as I mentioned in the comment above.

csongnr (approved):

PR #857, opened 2023-01-26
Add support for Kubernetes v1.26

vuqtran88:
This PR adds support for Kubernetes v1.26. More specifically, this does the followings:
- Lifts up the integration for k8s v1.26 .
- Detects k8s version. For k8s >= v1.25, install KSM from the v2.7.0 image, otherwise install the v2.6.0 image. This is because k8s v1.26 can only work KSM image v2.7.0,  v1.23<= k8s <= v1.25 can work with either KSM image v2.6.0 or v2.7.0, k8s < v1.23 can only work with KSM image v2.6.0.

I have tested the recipe with a minikube cluster with the following test scenarios:
1. Helm is installed + minikube cluster with k8s v1.26
2. Helm is installed + minikube cluster with k8s v1.24
3. Helm is NOT installed + minikube cluster with k8s v1.26
4. Helm is NOT installed + minikube cluster with k8s v1.24

### DIFF
--- a/recipes/newrelic/infrastructure/kubernetes.yml
+++ b/recipes/newrelic/infrastructure/kubernetes.yml
@@ -160,8 +160,8 @@ install:
             exit 131
           fi
 
-          if [[ "$SERVER_MAJOR_VERSION" -lt 1 ]] || [[ "$SERVER_MAJOR_VERSION" -eq 1 && "$SERVER_MINOR_VERSION" -lt 16 ]] || [[ "$SERVER_MAJOR_VERSION" -eq 1 && "$SERVER_MINOR_VERSION" -gt 25 ]]; then
-            echo "Installation failed. Kubernetes v1.16 to v1.25 is required, found v${SERVER_MAJOR_VERSION}.${SERVER_MINOR_VERSION}" >&2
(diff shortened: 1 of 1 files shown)


recipes/newrelic/infrastructure/kubernetes.yml
```diff
@@ -481,6 +481,13 @@ install:
               ARGS="${ARGS} --devel"
             fi
             ARGS="${ARGS} --set ksm.enabled=${NR_CLI_KSM}"
+
+            if [[ $SERVER_MAJOR_VERSION -eq 1 && $SERVER_MINOR_VERSION -lt 23 ]]; then
```
nrepai, line 485: this makes everything k8s >=1.24 use KSM 2.7.0?

this chart seems to imply that KSM 2.7.0 should be used for k8s >=1.25: https://github.com/kubernetes/kube-state-metrics#compatibility-matrix

vuqtran88: Hi @nrepai, great find! This actually makes everything k8s v>= 1.23 use KSM 2.7.0. Even though I tested k8s v1.23 with KSM v2.7.0, but it makes sense to follow the guidance from the compatibility matrix. I will make the change.

amolero-nr (approved):

PR #862, opened 2023-02-08
fix(gcp): added timeout to 2s to GCP detection

kidk:
It was missing, which means the default of 'forever' is chosen. Most of the time it was 60 seconds. 

This is kinda a blind fix, hope we have some tests for this 😆


Julien4218 (approved):
Verified on GCP compute instance

PR #866, opened 2023-02-14
chore(logging): output SSL ciphers for debian infra recipes

mbruzina:
Dumping SSL ciphers to assist troubleshooting of apt failures.  Example output:
```
==> Installing Infrastructure Agent
Detecting available SSL ciphers...
 - SSLv3
 - TLSv1
 - TLSv1.2
 - TLSv1.3
```

If any command within the new `log_ssl_ciphers` task fails, installation will continue and the failure output will be logged.

### DIFF
--- a/recipes/newrelic/infrastructure/debian.yml
+++ b/recipes/newrelic/infrastructure/debian.yml
@@ -61,6 +61,7 @@ install:
         - task: cleanup
         - task: setup_license
         - task: setup_proxy
+        - task: log_ssl_ciphers
         - task: update_apt
         - task: install_gnupg
         - task: add_gpg_key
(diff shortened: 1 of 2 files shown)


recipes/newrelic/infrastructure/debian.yml
```diff
@@ -183,6 +184,18 @@ install:
...
 
+    log_ssl_ciphers:
+      cmds:
+        - |
+          IS_OPENSSL_INSTALLED=$(which openssl | wc -l)
+          IS_SORT_INSTALLED=$(which sort | wc -l)
+          IS_UNIQ_INSTALLED=$(which uniq | wc -l)
+          if [ $IS_OPENSSL_INSTALLED -gt 0 ] && [ $IS_SORT_INSTALLED -gt 0 ] && [ $IS_UNIQ_INSTALLED -gt 0 ]; then
+            echo "Detecting available SSL ciphers..."
+            openssl ciphers -v | awk '{print " - " $2}' | sort | uniq
```
Julien4218, line 195: @mbruzina this is an interesting idea. I'm not sure the value of sort/uniq is worth adding a constraint and not log this output when any of those tools is not there, they might not be that common on some bare linux.
Maybe an alternative could be to concatenate all in a single line, with the "Detecting available SSL:" message? Just a thought.

noahmmcgivern (approved):

recipes/newrelic/infrastructure/debian.yml
mbruzina: That can work too...was just grabbing a command from a recent slack conversation.  There are tons of dupes without sort/unique though.  Let me see what I can do using the utils we bundle with the CLI 👍

Julien4218: Sure, dont feel like you have to change or spend a lot of time doing differently. We could always make more changes if we see we're missing most of the usages.

noahmmcgivern (approved):

PR #881, opened 2023-03-13
fix(ms-sql): added tls/ssl checks

NSSPKrishna:


recipes/newrelic/infrastructure/ohi/sql/ms-sql.yml
```diff
@@ -91,7 +91,19 @@ install:
         - task: assert_infra
         - task: stop_infra
         - task: input_install
+        - task: download
 
+    download:
```
amolero-nr, line 96: Thank you @NSSPKrishna! Please, consider the following comments:

1) The `download` task should run before the `input_install` task for `WebClient` to pick up the right protocols, if needed. So, the order of lines 93 and 94 should be fixed.

2) Rename the `download` task to something like `set_security_protocols` or `set_protocols` to make it more descriptive of what the task is actually doing.

3) Position the `download` task (after being correctly renamed as requested in item 2) to the right place in the code to keep things organized.

NSSPKrishna: Made the changes. Please review it again

amolero-nr: Can you please move all the block of [`set_security_protocols`](https://github.com/newrelic/open-install-library/blob/pk/NR-91116/recipes/newrelic/infrastructure/ohi/sql/ms-sql.yml#L96-L106) to line 407 to keep all tasks code in order inside the file?

NSSPKrishna:
Tested with windows 2019 with sql 2019 Standard and windows 2016 with sql 2019. No errors found with the updated recipe. Created an EC2 instance with windows 2012 R2 SQL server 2016 standard. Failed to create a user with demo deployer. Will raise another PR for fixing windows 2012 R2.

recipes/newrelic/infrastructure/ohi/sql/ms-sql.yml
NSSPKrishna: Refactored to match the order of task.

Julien4218:
> @Julien4218, we validated the changes with recipe that creates Windows Server 2019 and MSSQL 2019. As suggested by you, we will validate these changes against Windows 2016 server and/or Windows 2012 R2 and will let the team know.
> 
> cc @amolero-nr @NSSPKrishna

Great thank you. For Windows 2012 R2, you can you just launch an instance in AWS if it's easier. Remember to terminate when you're done testing.

amolero-nr:
Testing `tls13` in Windows Server 2012 R2: a) simple command check, b) in the install command and c) in the infra agent;  it seems to work okay (FYI: @Julien4218, @NSSPKrishna):

image

image

image

amolero-nr (approved):

PR #877, opened 2023-03-06
feat(apm): Application Name Envar

amolero-nr:
Removes old `inputVars` method prompting user for application name and restart option. Adds support for an optional new `NEW_RELIC_APPLICATION_NAME` environment variable.

PR #889, opened 2023-03-24
PHP recipe application name

rthorn-nr:
Handles whitespace in the application name by writing the application name and restart value to separate lines in the temporary file. 

Also includes a change for the application name regex to detect for any value of `newrelic.appname` and not just "PHP Application", which fixes the long-standing behavior in redhat.yml which did not allow for multiple changes of `newrelic.appname`

### DIFF
--- a/recipes/newrelic/apm/php/debian.yml
+++ b/recipes/newrelic/apm/php/debian.yml
@@ -208,19 +208,22 @@ install:
             read -r APPLICATION_RESTART
             APPLICATION_RESTART=${APPLICATION_RESTART:-y}
 
-            echo "$APPLICATION_NAME $APPLICATION_RESTART" > "{{.TMP_INSTALL_DIR}}/{{.USER_INPUT}}"
+            echo "$APPLICATION_NAME" > "{{.TMP_INSTALL_DIR}}/{{.USER_INPUT}}"
+            echo "$APPLICATION_RESTART" >> "{{.TMP_INSTALL_DIR}}/{{.USER_INPUT}}"
           fi
 
           # Interactive install with NEW_RELIC_APPLICATION_NAME envar set, uses envar and also defaults to service restart
           if [[ "{{.NEW_RELIC_ASSUME_YES}}" != "true" && ! -z $NEW_RELIC_APPLICATION_NAME ]]; then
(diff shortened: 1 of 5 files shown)


michaelkro (commented):
Tested:

* php-apache-fpm-wordpress-linux2
   * Set name with envar ✅
   * Set name with prompt ✅
   * Allow CLI to use default by passing `-y` ✅
 * php-apache-wordpress-ubuntu18
   * Set name with envar ✅
   * Set name with prompt ✅
   * Allow CLI to use default by passing `-y` ✅

recipes/newrelic/apm/php/debian.yml
```diff
@@ -208,19 +208,22 @@ install:
...
+            echo "$NEW_RELIC_APPLICATION_NAME" > "{{.TMP_INSTALL_DIR}}/{{.USER_INPUT}}"
+            echo "y" >> "{{.TMP_INSTALL_DIR}}/{{.USER_INPUT}}"
           fi
 
     verify_continue:
       cmds:
         - |
           if [[ "{{.NEW_RELIC_ASSUME_YES}}" != "true" ]]; then
-            read -r APPLICATION_NAME APPLICATION_RESTART < "{{.TMP_INSTALL_DIR}}/{{.USER_INPUT}}"
+            read -r APPLICATION_NAME < "{{.TMP_INSTALL_DIR}}/{{.USER_INPUT}}"
```
Julien4218, line 225: @rthorn-nr why don't we just consolidate the step user_input with verify_continue, and avoid the writing/reading from a temp file?

rthorn-nr: Great idea @Julien4218 , I'll take a look at that this afternoon

rthorn-nr: Actually @Julien4218 these are used in a later task when writing to newrelic.ini, so we would need to collapse all of those tasks to be able to scope them to a single task or write them to a temp file anyway. I'm trying to find alternatives to this but don't want to spend too much time on it. 

This strategy of using a temp file was merged a few weeks ago in https://github.com/newrelic/open-install-library/pull/877

Julien4218: Ok, I didn't see it was used somewhere else. Something to change later then as we're looking into removing our dependency on go tasks.

amolero-nr (approved):

PR #879, opened 2023-03-08
Collect More Error Messages from K8s Recipe

xqi-nr:
# Description

A customer encounter the error message while install K8s integration
```
Error: looks like "https://helm-charts.newrelic.com/" is not a valid chart repository or cannot be reached: Get "https://newrelic.github.io/helm-charts/index.yaml": Forbidden
```

# Changes
For debugging purpose, I add code to collect more error info at several places before error cause exit.

PR #890, opened 2023-03-24
Fix helm repo add error msg

xqi-nr:
# Description
In installation recipe, when customer does not have repo newrelic installed, the script will exit due to a bug introduced in https://github.com/newrelic/open-install-library/pull/879
The error message is like `helm (version v3.11.2+g912ebc1) repo add failed due to: "newrelic" has been added to your repositories`
which should not lead to exist.

# Changes

modify the code to exit only when the helm repo add command results in non zero code

# Tests 
I test the modified script with the following commands on K8s 1.17 and 1.26 with and without nri-bundle preinstalled.

All the testing installation succeeded

```
xqi@TFXCKKXMP9 newrelic-cli_0.65.1_Darwin_arm64 % NEW_RELIC_CLI_SKIP_CORE=1 NR_CLI_CLUSTERNAME=k8scluster117 NR_CLI_NAMESPACE=newrelic NR_CLI_PRIVILEGED=true NR_CLI_LOW_DATA_MODE=true NR_CLI_KSM=true NR_CLI_KUBE_EVENTS=true NR_CLI_PROMETHEUS_AGENT=true NR_CLI_PROMETHEUS_AGENT_LOW_DATA_MODE=true NR_CLI_CURATED=false NR_CLI_NEWRELIC_PIXIE=true NR_CLI_PIXIE_API_KEY=XXX NR_CLI_PIXIE=true NR_CLI_PIXIE_DEPLOY_KEY=XXX NEW_RELIC_API_KEY=XXX NEW_RELIC_ACCOUNT_ID=XXX ./newrelic install -c /Users/xqi/Documents/GitHub/open-install-library/recipes/newrelic/infrastructure/kubernetes.yml
```
I also printout the $? value with and without newrelic repo pre-existed.  in both cases, the value is 0.

### DIFF
--- a/recipes/newrelic/infrastructure/kubernetes.yml
+++ b/recipes/newrelic/infrastructure/kubernetes.yml
@@ -489,7 +489,7 @@ install:
             if [[ $? != 0 ]]; then
               if [[ "${ERROR}" !=  "" ]]; then
                 echo "helm (version $HELM_VERSION) repo add failed due to: $ERROR"
-              else 
+              else
                 echo "helm (version $HELM_VERSION) repo add failed."
               fi
               exit 131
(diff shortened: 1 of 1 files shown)


Julien4218 (approved):
E2E test passing, no regression

PR #895, opened 2023-03-30
chore: Update Amazon Linux AMI from 2022 to 2023

NRhzhao:


NRhzhao:
Update date test AMI from AL2022 to 2023, as AL2022 AMI no longer exist

1. Update alert test to use AL2023 image
2. Remove infra AL2022 test, there is already a AL2023 test
3. Update Ubuntu Nginx AMI, the existing filter no longer exist

mbruzina (approved):

PR #906, opened 2023-04-13
chore: reduce php recipe redundancies

noahmmcgivern:


sanderblue (approved):

recipes/newrelic/apm/php/debian.yml
```diff
@@ -143,59 +154,6 @@ install:
...
 
-    assert_pre_req:
-      cmds:
-        - |
-          if [ -z "$SUDO_USER" ]; then
-            echo -e "{{.RED}}This installation recipe for the New Relic PHP agent must be run under sudo.{{.NOCOLOR}}" >> /dev/stderr
-            exit 7
-          fi
-        - |
-          IS_SYSTEMCTL_INSTALLED=$(which systemctl | wc -l)
```
zsistla, line 154: Why was this check removed?  Both address specific cases where PHP installation can fail.  Without the check the user has no way of knowing why installation failed.

PR #915, opened 2023-04-25
fix(alerts): check if the policy file is available

NSSPKrishna:
Fixes [NR-108091](https://issues.newrelic.com/browse/NR-108091)


recipes/newrelic/infrastructure/alerts/golden.yml
```diff
@@ -78,6 +78,11 @@ install:
             "query": "mutation {alertsPolicyCreate(accountId: {{.NEW_RELIC_ACCOUNT_ID}}, policy: { name: POLICY_NAME incidentPreference: PER_CONDITION_AND_TARGET }) { id name } }"
           }
           EOT
+          if [ $? -eq 0 ]; then 
+            echo "File '/tmp/policy.json' is created"
+          else 
+            echo "Unable to create the file '/tmp/policy.json'. Please retry using the right permissions"
+          fi 
           sudo sed -i "s/POLICY_NAME/\"\"{{.ALERT_POLICY_NAME}}\"\"/g" /tmp/policy.json
```
mbruzina, line 86: What happens with these two `sed` commands if we cannot create `/tmp/policy.json`?  I see we eventually exit if a curl fails a few lines down, but am not sure if this error handling handles both sed failures and the curl failure (if that makes sense).  Can we simply exit from the `else` if we're unable to create a temp file?

NSSPKrishna: Good call. We can exit from it. Thinking of a better way instead of echoing "unable to create a file".

mbruzina:
I'm thinking it may suffice to check permissions on the `/tmp` directory only a few times...once for removing files, and again for file creation.  Since this is all in the same task, we probably only need to do it once per operation - so checking on the removal/write for `/tmp/policy.json` may be all we need

https://github.com/newrelic/open-install-library/blob/36677e6362a2d41cbcda58e21a4ff26c653f075f/recipes/newrelic/infrastructure/alerts/golden.yml#L73-L86  

or to be certain on `/tmp/condition.json` add an exit if we cannot delete, keep the `exit 2` you already have , but then remove the remaining `exit 2` checks from subsequent `condition.json` file interactions

https://github.com/newrelic/open-install-library/blob/36677e6362a2d41cbcda58e21a4ff26c653f075f/recipes/newrelic/infrastructure/alerts/golden.yml#L114-L127

mbruzina (approved):

PR #954, opened 2023-08-04
feat(php): add helpful error messaging for some important preInstall checks

sanderblue:
This PR adds clear, helpful error messages for the following error scenarios.

Recipe Checklist
- [x] Debian
- [x] RedHat

1. When PHP is not installed
2. When php-fpm and Apache are not installed or running
3. When Apache is running, but the PHP module is not enabled

### DIFF
--- a/recipes/newrelic/apm/php/debian.yml
+++ b/recipes/newrelic/apm/php/debian.yml
@@ -40,6 +40,7 @@ preInstall:
       IS_PHP_INSTALLED=$(which php | wc -l)
       if [ "$IS_PHP_INSTALLED" -eq 0 ]; then
         # PHP not installed, no detection
+        echo "Error: PHP is not installed. Please install PHP, configure and enable your web server (Apache or NGINX), then retry the installation." >&2
         exit 1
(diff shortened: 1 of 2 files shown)


recipes/newrelic/apm/php/debian.yml
```diff
@@ -40,6 +40,7 @@ preInstall:
       IS_PHP_INSTALLED=$(which php | wc -l)
       if [ "$IS_PHP_INSTALLED" -eq 0 ]; then
         # PHP not installed, no detection
+        echo "Error: PHP is not installed. Please install PHP, configure and enable your web server (Apache or NGINX), then retry the installation." >&2
```
Julien4218, line 43: @sanderblue are those messages displayed to the user, since it's in the "requireAtDiscovery" detection step?

sanderblue: No, I guess they aren't since it's in this early step, but they do show up when using `--debug`.

Julien4218 (approved):

PR #973, opened 2023-09-08
fix(tests): Update AMIs

amolero-nr:
Update several test definition files with a newer AMI name.

### DIFF
--- a/test/definitions-eu/otel/rhel/awslinux23-otel.json
+++ b/test/definitions-eu/otel/rhel/awslinux23-otel.json
@@ -9,7 +9,7 @@
     "resources": [{
         "id": "host1",
         "provider": "aws",
-        "ami_name": "al2023-ami-2023.0.*-x86_64",
+        "ami_name": "al2023-ami-2023.1.*-x86_64",
         "type": "ec2",
         "size": "t3.nano"
     }],
--- a/test/definitions-eu/otel/rhel/awslinux23-otel.json
+++ b/test/definitions-eu/otel/rhel/awslinux23-otel.json
@@ -7,7 +7,7 @@
     },
 
     "resources": [{
-        "id": "host1",
+        "id": "host2",
         "provider": "aws",
         "ami_name": "al2023-ami-2023.*-x86_64",
         "type": "ec2",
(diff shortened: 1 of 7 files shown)


test/definitions-eu/otel/rhel/awslinux23-otel.json
```diff
@@ -9,7 +9,7 @@
     "resources": [{
         "id": "host1",
         "provider": "aws",
-        "ami_name": "al2023-ami-2023.0.*-x86_64",
+        "ami_name": "al2023-ami-2023.1.*-x86_64",
```
rthorn-nr, line 12: Is this digit something we'd want to wildcard? Would be a pain if this changes to `2`

amolero-nr: I guess we could, but maybe is useful to make sure to test with the oldest minor version of the AL2023 AMI?

From what I [read](https://docs.aws.amazon.com/linux/al2023/ug/naming-and-versioning.html), when version `2` is released (e.g.: `al2023-ami-2023.2.*-x86_64`), version `1` is not automatically removed, they will be there for some time (standard and extended support).

rthorn-nr: That sounds reasonable to me, even though it will mean another update in the future

rthorn-nr (dismissed):

test/definitions-eu/otel/rhel/awslinux23-otel.json
amolero-nr: Done in 9c1ca48f8772aace21f16d4b7a4def96a8f42107

rthorn-nr (approved):

test/definitions-eu/otel/rhel/awslinux23-otel.json
```diff
@@ -7,9 +7,9 @@
     },
 
     "resources": [{
-        "id": "host1",
+        "id": "host2",
```
NSSPKrishna, line 10: 👋 , is there a specific reason behind changing the id?

PR #976, opened 2023-09-28
Fix new config setup

alvarocabanas:
Fix new config setup for super-agent release.
Logs-super-agent no longer has a dependency.

### DIFF
--- a/recipes/newrelic/infrastructure/super-agent/debian.yml
+++ b/recipes/newrelic/infrastructure/super-agent/debian.yml
@@ -335,6 +335,12 @@ install:
     config_supervisors:
       cmds:
         - |
+          if [ "{{.NR_CLI_INFRA_AGENT}}" == "false" && "{{.NR_CLI_NRDOT}}" == "false"] ; then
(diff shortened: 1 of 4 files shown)


NRhzhao:
@alvarocabanas Looks like we can't just remove `super-agent` dependency from logs. This will cause the regular guided install to offer logs again (Super-agent logs).
Not sure why it wasn't working for you, but super-agent logs were installed when I tried earlier. 
If you want to disable the super-agent log recipes for now, you can probably just do a exit 1 in the assert_pre_req step.
But, you would need to fix all the super-agent log tests, or they would all fail.


image

alvarocabanas:
> @alvarocabanas Looks like we can't just remove `super-agent` dependency from logs. This will cause the regular guided install to offer logs again (Super-agent logs). Not sure why it wasn't working for you, but super-agent logs were installed when I tried earlier. If you want to disable the super-agent log recipes for now, you can probably just do a exit 1 in the assert_pre_req step. But, you would need to fix all the super-agent log tests, or they would all fail.
> 
> image

Sure, I have just rollbacked the logs modification and just left the super-agent ones.

NRhzhao (approved):

PR #994, opened 2023-10-30
Super Agent recipe, add all repository to allow Fluentbit installation and fix bugs

alvarocabanas:
- Add all repository to allow Fluentbit and integrations installation
- Remove dependency from logging causing it to fail
- Fix bugs on Host Monitoring

### DIFF
--- a/recipes/newrelic/infrastructure/super-agent/logs/linux-logs.yml
+++ b/recipes/newrelic/infrastructure/super-agent/logs/linux-logs.yml
@@ -68,6 +68,11 @@ installTargets:
     platformFamily: rhel
     platformVersion: "((7|8)\\.?.*)"
     kernelArch: aarch64
+  - type: host
+    os: linux
+    platformFamily: suse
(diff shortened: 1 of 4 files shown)


recipes/newrelic/infrastructure/super-agent/logs/linux-logs.yml
```diff
@@ -6,9 +6,6 @@ displayName: Logs Integration
 description: New Relic install recipe for Logging in super-agent
 repository: https://github.com/newrelic/infrastructure-agent/tree/master/assets/examples/logging
 
-dependencies:
```
NRhzhao, line 9: @alvarocabanas Why are we removing the dependencies here?

alvarocabanas: Because it's failing and saying incompatible. 
We should have this dependency in the future but we can leave it without for the moment.

NRhzhao (approved):

PR #1004, opened 2023-12-15
feat(kubernetes): Add gke autopilot check to script

svetlanabrennan:
Updating the script to check if user is installing in a GKE Autopilot. If they are, it will turn off privileged mode, control plane and then turn off pixie since it requires privileged access. It also turns off logging which is not currently supported in GKE autopilot.

### DIFF
--- a/recipes/newrelic/infrastructure/kubernetes.yml
+++ b/recipes/newrelic/infrastructure/kubernetes.yml
@@ -632,6 +632,14 @@ install:
               fi
             fi
           else
+            # We currently dont' support installing in GKE Autopilot using kubectl
+            if [[ "$GKE_AUTOPILOT_ANSWER" == "Y" ]]; then
+             echo ""
+             echo -e "\033[0;31mUsing kubectl to install the Kubernetes integration is currently not supported.\033[0m" >&2
+             echo -e "\033[0;31mPlease install helm and try again.\033[0m" >&2
+             exit 131
+            fi
+
             echo ""
             echo "Using kubectl to install the Kubernetes integration"
             echo ""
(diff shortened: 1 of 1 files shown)


recipes/newrelic/infrastructure/kubernetes.yml
```diff
@@ -190,6 +190,30 @@ install:
           # DEBUG_MESSAGE is used to collect all necessary debug info we need.
           DEBUG_MESSAGE=""
 
+          # Determine if user is installing in gke autopilot
+          while :; do
+            echo -e -n "${GREEN}Are you installing in a GKE Autopilot cluster? Y/N? ${NC} "
```
csongnr, line 195: I think you're going to have to add an
`if [[ "{{.NEW_RELIC_ASSUME_YES}}" == "true" ]]; then
                  GKE_AUTOPILOT_ANSWER="N"
`
somewhere in this conditional, its for when the install process allows you to set certain flags, to answer all the questions asked in the install process. I think they use this flag in some automated testing suites and it'll break if theres no default answer defined. I'm going to assume that we're probably going to have to assume that gke autopilot is not the default mode, so answer is 'N' not 'Y' but give it a thought and see what you think 

https://github.com/newrelic/newrelic-cli/blob/main/docs/cli/newrelic_install.md 

take a look and see what you think about the default answer

svetlanabrennan: Done

csongnr (approved):

recipes/newrelic/infrastructure/kubernetes.yml
```diff
@@ -190,6 +190,40 @@ install:
...
+              echo ""
+              GKE_AUTOPILOT_ANSWER=$(echo "${ans^^}" | cut -c1-1)
+              if [[ "$GKE_AUTOPILOT_ANSWER" == "N" ]]; then
+                echo "Continuing with installing the integration."
+                echo "{\"Metadata\":{\"gkeAutopilotAnswer\":\"No\",\"DebugMessage\":\"$DEBUG_MESSAGE\"}}" | tee -a {{.NR_CLI_OUTPUT}} > /dev/null
+                break
+              fi
+              if [[ "$GKE_AUTOPILOT_ANSWER" == "Y" ]]; then
+                echo "{\"Metadata\":{\"gkeAutopilotAnswer\":\"Yes\",\"DebugMessage\":\"$DEBUG_MESSAGE\"}}" | tee -a {{.NR_CLI_OUTPUT}} > /dev/null
+                echo "GKE Autopilot does not allowed privileged access. Turning off privileged mode." >&2
```
NSSPKrishna, line 215:
```suggestion
                echo -e "\033[0;31mGKE Autopilot does not allow privileged access. Turning off privileged mode.\033[0m" >&2
```

NSSPKrishna (changes requested):
Everything looks good to me. Just a nit.

amolero-nr (approved):

PR #1010, opened 2024-01-02
chore(nodejs): exit installation if Node.js version is unsupported

sanderblue:


mbruzina (dismissed):

recipes/newrelic/apm/node/linux.yml
```diff
@@ -24,17 +24,25 @@ preInstall:
       if [ -z "$ASSERT_NODE_INSTALLED" ] ; then
         exit 1
       fi
+      LOWEST_SUPPORTED_MAJOR_NODE_VERSION=16
+      INSTALLED_MAJOR_NODE_VERSION=$(node -v | awk -F '[^0-9]+' '{ print $2 }')
```
stevula, line 28: It's very common to use `nvm` to manage multiple node versions, so the system node version might not be the node version that is being used to run the app. We could parse the node version from the output of `pm2 list`, similar to what's being done here:
https://github.com/newrelic/open-install-library/pull/1010/files#diff-465f7b5cc2681ce268fd0072c1926be849dedca67a68ba4a5b5776cf9dffd688L111

mbruzina: Agreed on checking `nvm`, good call - though could we get by with simply performing `which` to determine then nvm directly?  Wouldn't want to install the introspector if they're not supported

stevula (changes requested):
I tested this on an EC2 instance with node 14 and 16 managed via nvm and a test app running via pm2. I locally compiled newrelic-cli using the code from this branch and ran the following install script:

```
curl -Ls https://download.newrelic.com/install/newrelic-cli/scripts/install.sh | bash && sudo NEW_RELIC_API_KEY=NRAK-**** NEW_RELIC_ACCOUNT_ID=**** NEW_RELIC_CLI_SKIP_CORE=1 ./newrelic install -n node-agent-installer --debug
```

#### Node 14
Expected behavior: 
Script should fail with error "Node.js version not supported [...]"

Actual behavior: 
🔴  Script shows node agent as "Installed" but the entity does show up in NR UI under All Entities view. 
🔴 Debug output shows errors in the `preInstall` commands:
```
DEBUG "node": executable file not found in $PATH
DEBUG "node": executable file not found in $PATH
DEBUG 7:4: not a valid test operator: 16
DEBUG 7:4: not a valid test operator: 16
DEBUG 7:4: 16 must be followed by a word
DEBUG 7:4: 16 must be followed by a word
```

#### Node 16 
Expected behavior:
Script should show Node agent as installed

Actual Bheavior:
✅  Script shows Node agent as installed.
🟡 Debug output shows errors in the `preInstall` commands (same as above).

recipes/newrelic/apm/node/linux.yml
sanderblue: I'm making sure this works by getting the path to the binary via `command -v node`. Just confirmed it works on the EC2 instance I'm testing with and it uses `nvm` to manage node versions.

`NODEJS_BINARY_PATH=$(command -v node)` results in the following output:

```
$ command -v node
/home/ec2-user/.nvm/versions/node/v14.16.1/bin/node
```

Confirmed this works in the recipe as well.

mbruzina (approved):

sanderblue:
> I tested this on an EC2 instance with node 14 and 16 managed via nvm and a test app running via pm2. I locally compiled newrelic-cli using the code from this branch and ran the following install script:
> 
> ```
> curl -Ls https://download.newrelic.com/install/newrelic-cli/scripts/install.sh | bash && sudo NEW_RELIC_API_KEY=NRAK-**** NEW_RELIC_ACCOUNT_ID=**** NEW_RELIC_CLI_SKIP_CORE=1 ./newrelic install -n node-agent-installer --debug
> ```
> 
> #### Node 14
> Expected behavior: Script should fail with error "Node.js version not supported [...]"
> 
> Actual behavior: 🔴 Script shows node agent as "Installed" but the entity does not show up in NR UI under All Entities view. 🔴 Debug output shows errors in the `preInstall` commands:
> 
> ```
> DEBUG "node": executable file not found in $PATH
> DEBUG "node": executable file not found in $PATH
> DEBUG 7:4: not a valid test operator: 16
> DEBUG 7:4: not a valid test operator: 16
> DEBUG 7:4: 16 must be followed by a word
> DEBUG 7:4: 16 must be followed by a word
> ```
> 
> #### Node 16
> Expected behavior: Script should show Node agent as installed
> 
> Actual Behavior: ✅ Script shows Node agent as installed. 🟡 Debug output shows errors in the `preInstall` commands (same as above).
> 
> A possible problem here is that the install script is run with sudo and node/nvm might not be on the secure path.

Should be working now after my recent changes. Let me know if you run into anymore issues. Thanks!

PR #1013, opened 2024-01-11
chore(nodejs): exit installation if Node.js version is unsupported

sanderblue:


stevula (approved):
- [x] Node 16 => status "Installed" and entity appears in NR entities page.

- [ ] Node 14 => status "Failed" with message:
> Error: Node.js version not supported. Please upgrade to the latest supported version.


- [ ] No Node processes => status "Canceled" with message:
> No Node processes found running on the host

PR #1022, opened 2024-01-24
fix(ohi): Ingest current cassandra logs only

amolero-nr:
Enables log ingestion of `Cassandra`'s currently active `system/debug` logs only, excluding any rotated ones (system.log.1, etc). Ref: [NR-216306](https://new-relic.atlassian.net/browse/NR-216306).

### DIFF
--- a/test/deploy/linux/cassandra/install/debian/roles/prepare/tasks/installCassandra.yml
+++ b/test/deploy/linux/cassandra/install/debian/roles/prepare/tasks/installCassandra.yml
@@ -7,7 +7,7 @@
   become: true
 
 - name: add cassandra keys
-  shell: wget -q -O - https://www.apache.org/dist/cassandra/KEYS | sudo apt-key add -
+  shell: wget -q -O - https://downloads.apache.org/cassandra/KEYS | sudo apt-key add -
   become: true
 
 - name: Install cassandra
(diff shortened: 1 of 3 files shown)


test/deploy/linux/cassandra/install/debian/roles/prepare/tasks/installCassandra.yml
```diff
@@ -7,7 +7,7 @@
   become: true
 
 - name: add cassandra keys
-  shell: wget -q -O - https://www.apache.org/dist/cassandra/KEYS | sudo apt-key add -
+  shell: wget -q -O - https://downloads.apache.org/cassandra/KEYS | sudo apt-key add -
```
amolero-nr, line 10: To fix:

image

amolero-nr: The **_wrong_** KEYS file URL should be the one causing the unit test validation to fail.

rthorn-nr (approved):
The error in the failing test is similar to the one I saw for `ca-certificates-java` install on Debian 10 in this PR: https://github.com/newrelic/open-install-library/pull/1008 This means we'll probably need to move this test to Debian 11. 

Since this change is so transparent and the test failure is an unrelated issue, I'll go ahead and approve this, and we can take care of the test in a later PR

PR #1041, opened 2024-02-13
chore(dotnet): Modifying ports to avoid port address issue

Avinash-Sanpala:
JIRA: https://new-relic.atlassian.net/browse/NR-218247

As part of this ticket I'm modifying the existing dotnet sample application installation ports to avoid the port address already in use issue during instrumentation


pchinthapenta (approved):
looks good

PR #1043, opened 2024-02-13
chore:update go deploy config with go mod and go sum

pchinthapenta:
Added creation of go mod and go sum to go deploy config


Avinash-Sanpala (approved):
looks good to me!

PR #1051, opened 2024-02-26
chore(tests): Update nginx-linux-2 ami

amolero-nr:
Fixes image not found for `nginx-linux-2-ami` test (Ref: [NR-232715](https://new-relic.atlassian.net/browse/NR-232715))


rthorn-nr:
Hey @amolero-nr , I'm seeing another config with the same pattern in an EU smoke test: `test/definitions-eu/smoke/nginx-linux2-ami.json` -- seems like we may need to update that one as well

rthorn-nr (approved):

PR #1067, opened 2024-04-18
ungate kubernetes agent install on v128+

csongnr:


ramkrishankumarN (approved):

nrepai (approved):

NSSPKrishna:
Code looks good!  A minor change to use constants for the supported Kubernetes version range would further improve readability and maintainability. This makes it easier to understand the supported versions and simplifies future modifications if the range changes.

NSSPKrishna (changes requested):
Code looks good!  A minor change to use constants for the supported Kubernetes version range would further improve readability and maintainability. This makes it easier to understand the supported versions and simplifies future modifications if the range changes.

csongnr:
@NSSPKrishna given how convoluted this recipe already is, I actually think its more readable when the actual version number that we minimally/maximally support is in the inline check rather than as a constant, as that would require having to find the constant definition and then referencing the actual value of the constant. We also only use the minimally and maximally supported kubernetes versions once in the entire recipe, so it wouldn't necessarily simplify future modifications in this instance, although I understand the motivation for your change request. We use constants here, however, to capture the output of the user's setup for easy reference in the conditionals. Let me know what you think, thanks.

NSSPKrishna (approved):

PR #1068, opened 2024-04-23
chore: Updated AMI name in ruby deploy config

pchinthapenta:
Updated AMI name in ruby deploy config, as the old AMI is no longer available in Mumbai region


Nandu-pns (approved):

Avinash-Sanpala (approved):
LGTM!

PR #1070, opened 2024-04-25
feat(superagent): Add fleet

gsanchezgavier:
This PR:
- Adds the `fleet_id` configuration to the super-agent recipe
- Align some differences between scripts for different OS. Sadly unifying thes task is not currently supported by newrelic-cli

### DIFF
--- a/recipes/newrelic/infrastructure/super-agent/debian.yml
+++ b/recipes/newrelic/infrastructure/super-agent/debian.yml
@@ -78,6 +78,7 @@ install:
         - task: install_super_agent
         - task: update_otel_license_key
         - task: config_supervisors
+        - task: config_fleet_id
         - task: config_opamp
         - task: config_host_monitoring
         - task: update_otel_mem_limit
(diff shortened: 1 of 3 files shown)


recipes/newrelic/infrastructure/super-agent/debian.yml
```diff
@@ -390,6 +391,14 @@ install:
             sed -i '/^\s*#\s*agent_type: "newrelic\/io\.opentelemetry\.collector/s/#//' /etc/newrelic-super-agent/config.yaml
           fi
 
+    config_fleet_id:
+      cmds:
+        - |
+          if [ ! -z "{{.NR_CLI_FLEET_ID}}" ] ; then
+            sed -i 's/s*#\s*fleet_id:/fleet_id:/g' /etc/newrelic-super-agent/config.yaml
```
rubenruizdegauna, line 398: Is this regexp ☝️  correct? I would say that the first `s` lacks a `\`

recipes/newrelic/infrastructure/super-agent/debian.yml
```diff
@@ -390,6 +391,14 @@ install:
...
+            sed -i 's/fleet_id: FLEET_ID_HERE/fleet_id: {{.NR_CLI_FLEET_ID}}/g' /etc/newrelic-super-agent/config.yaml
```
rubenruizdegauna, line 399: Is it correct to hardcode `FLEET_ID_HERE` ? why are we not using a regex here?

recipes/newrelic/infrastructure/super-agent/debian.yml
rubenruizdegauna: (Same for the rest below)

recipes/newrelic/infrastructure/super-agent/debian.yml
rubenruizdegauna: (same for the rest below)

gsanchezgavier: I followed the same pattern used for API_KEY_HERE, where that [example](https://github.com/newrelic/newrelic-super-agent/blob/main/build/examples/super-agent-config-no-agents.yaml) is replaced by the recipe

recipes/newrelic/infrastructure/super-agent/debian.yml
gsanchezgavier: good catch, yes that was totally not ineffective

recipes/newrelic/infrastructure/super-agent/debian.yml
rubenruizdegauna: :) got it, thanks!

rubenruizdegauna (approved):
LGTM

NSSPKrishna (approved):

PR #1094, opened 2024-07-11
feat(super-agent): support upgrades, do not touch configs

DavSanchez:
This allows to have some degree of idempotence for installs of the super-agent, not touching config files when an installation has run successfully before. It achieves this by placing a sentinel file when an installation runs successfully, and relevant installation tasks first check for the existence of this file to decide if actions are performed or not.

### DIFF
--- a/recipes/newrelic/infrastructure/super-agent/debian.yml
+++ b/recipes/newrelic/infrastructure/super-agent/debian.yml
@@ -225,6 +225,7 @@ install:
           fi
         - |
           if [ -f /etc/newrelic-super-agent/.nr-cli ] ; then
+            echo "skipping setup_infra_license sub-task"
             exit 0
           fi
           if [ "{{.NR_CLI_INFRA_AGENT}}" != "false" ] ; then
@@ -425,6 +426,7 @@ install:
           fi
         - |
           if [ -f /etc/newrelic-super-agent/.nr-cli ] ; then
+            echo "skipping config_supervisors sub-task"
             exit 0
           fi
           if [ "{{.NR_CLI_INFRA_AGENT}}" = "false" ]; then
@@ -436,6 +438,7 @@ install:
           fi
         - |
           if [ -f /etc/newrelic-super-agent/.nr-cli ] ; then
+            echo "skipping config_supervisors sub-task"
             exit 0
           fi
           if [ "{{.NR_CLI_NRDOT}}" = "false" ]; then
@@ -480,6 +483,7 @@ install:
           fi
         - |
           if [ -f /etc/newrelic-super-agent/.nr-cli ] ; then
+            echo "skipping config_opamp sub-task"
             exit 0
           fi
           if [ "{{.NEW_RELIC_REGION}}" = "STAGING" ]; then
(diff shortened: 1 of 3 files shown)


recipes/newrelic/infrastructure/super-agent/debian.yml
```diff
@@ -166,14 +176,22 @@ install:
     cleanup:
       cmds:
         - |
-          if [ "{{.NR_CLI_INFRA_AGENT}}" != "false" ] ; then
+          # Log if a previous execution of this recipe was detected, indicating we are skipping this task
+          if [ -f /etc/newrelic-super-agent/.nr-cli ] ; then
+            echo "A previous execution of this recipe was detected. Skipping "cleanup" task."
+          fi
```
rubenruizdegauna, line 182: If we need to skip the whole step, I would use an early return to increase readability
```
if [ -f /etc/newrelic-super-agent/.nr-cli ] ; then
  echo "A previous execution of this recipe was detected. Skipping "config_supervisors" task."
  exit 0
fi
```
I tested it, and it works as I would expect: exits that step and continues in the next one. wdyt?

@pranav-new-relic , do you see any problem with this?

pranav-new-relic: 👍 makes sense, @rubenruizdegauna. cc @DavSanchez

DavSanchez: Great suggestion! Implemented in https://github.com/newrelic/open-install-library/pull/1094/commits/0bc100e640a43619d0ce3940344a1e3905708a7c

recipes/newrelic/infrastructure/super-agent/debian.yml
```diff
@@ -423,21 +490,34 @@ install:
...
             if [ "{{.NR_CLI_NRDOT}}" != "false" ]; then
               mkdir -p /etc/newrelic-super-agent/fleet/agents.d/nr-otel-collector/values
-              cp /etc/newrelic-super-agent/examples/values-nr-otel-collector-gateway-0.1.0.yaml /etc/newrelic-super-agent/fleet/agents.d/nr-otel-collector/values/values.yaml
+              cp /etc/newrelic-super-agent/examples/values-nr-otel-collector-agent-linux-0.1.0.yaml /etc/newrelic-super-agent/fleet/agents.d/nr-otel-collector/values/values.yaml
+            else
+              if [ "{{.NR_CLI_NRDOT}}" != "false" ]; then
+                mkdir -p /etc/newrelic-super-agent/fleet/agents.d/nr-otel-collector/values
+                cp /etc/newrelic-super-agent/examples/values-nr-otel-collector-gateway-0.1.0.yaml /etc/newrelic-super-agent/fleet/agents.d/nr-otel-collector/values/values.yaml
+              fi
             fi
```
rubenruizdegauna, line 521: This code doesn't seem correct. The second `if` will never happen. A greater mind would explain like this:

> The code contains a logical error in the else block. The condition inside the else block is the same as the condition in the if block, which means the code inside the else block will never be executed.

pranav-new-relic: that's true; thanks for spotting this @rubenruizdegauna :)

DavSanchez: Fixed in https://github.com/newrelic/open-install-library/pull/1094/commits/0b5d29dad079f86e4d978740823a25ae1c445982!

gsanchezgavier:
in favor of https://github.com/newrelic/open-install-library/pull/1103

PR #1102, opened 2024-08-01
Super agent recipe final step check status endpoint

alvarocabanas:
Adapt the recipe to query (with retries) for a healthy status in the SA.

Rename the display name and logs as New Relic Agent Control

PR #1112, opened 2024-08-26
[NR-295076] super-agent: manually install fluent-bit for amazon linux

sigilioso:
This PR updates the super-agent's recipe to install fluent-bit manually on Amazon linux. This is needed because the [recommends package dependency](https://github.com/newrelic/newrelic-super-agent/blob/1abc91ceec135a3e26055ba406e90cdab3f1644a/.goreleaser.yml#L174) used in the super-agent's RPM (which usually installs fluent-bit as a dependency) isn't supported in the default RPM version in the AL2 x86 base image.


paologallinaharbur (approved):

sigilioso:
Failing checks are caused by unsupported OSs which are removed in https://github.com/newrelic/open-install-library/pull/1102

kang-makes (approved):

PR #1116, opened 2024-09-06
Use US OAuth urls instead of EU

davidgit:
IAM doesn't have system-identity-oauth in EU. There are no instances of system identity oauth running in EU. 

This change can be permanent.


DavSanchez (approved):
🚀

Nandu-pns (approved):

PR #1124, opened 2024-10-07
feat(K8s): Update K8s deploy config

abasha1234:
This PR will update the deploy config of Kubernetes to support all the install flow which including installing Helm.
Then deploy config was tested for the all the instrumentation flows (except Pixie) and installation was successful.

Ticket: https://new-relic.atlassian.net/browse/NR-221222


test/deploy/linux/kubernetes/minikube/roles/prepare/tasks/main.yml
```diff
@@ -70,3 +70,20 @@
...
+
+- name: Download the Helm binary
+  shell: 'wget https://get.helm.sh/helm-v3.8.0-linux-amd64.tar.gz'
+  become: true
+
+- name: Extract the Helm archive
+  shell: 'tar -xvf helm-v3.8.0-linux-amd64.tar.gz'
+
+- name: Move the Helm executable to a system directory
+  shell: mv linux-amd64/helm /usr/local/bin/
```
Avinash-Sanpala, line 82: will this `linux-amd64` path be consistent across other Linux distributions? and also I think its better to add permissions before moving the file to prevent the access issues. what do you think?

abasha1234: `linux-amd64` will be consistent for most of the Linux distributions as this contains `helm`. I tried manually and with deploy config locally without changing permissions, didn't see access issues. If you want I will add, not a problem.

Avinash-Sanpala: okay got it and yeah may be its better to add the permissions

Avinash-Sanpala (approved):

PR #1139, opened 2024-12-09
feat(agent-control): L2 identity creation

rubenruizdegauna:
Create L2 identities using L1 identity token in Agent Control installation.

### DIFF
--- a/recipes/newrelic/infrastructure/super-agent/debian.yml
+++ b/recipes/newrelic/infrastructure/super-agent/debian.yml
@@ -592,7 +592,7 @@ install:
                   break
                 fi
   
-                echo "Error creating the new system identity. The API endpoint returned $HTTP_CODE. Retrying ($RETRY/3)..."
+                echo "Error creating L2 system identity. The API endpoint returned $HTTP_CODE. Retrying ($RETRY/3)..."
                 sleep 2
               done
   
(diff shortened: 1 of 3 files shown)


recipes/newrelic/infrastructure/super-agent/debian.yml
```diff
@@ -519,58 +519,143 @@ install:
...
+                    -o "$TEMPORAL_FOLDER/response.json" \
+                    --data-binary @- \
+                    "$IDENTITY_CREATION_ENDPOINT"
+                )
+  
+                if [ $HTTP_CODE -eq 200 ]; then
+                  break
+                fi
+  
+                echo "Error creating the new system identity. The API endpoint returned $HTTP_CODE. Retrying ($RETRY/3)..."
```
sigilioso, line 595: Would it make sense to use slightly different error messages in order to help troubleshooting during the transition? For instance we could add L2 `Error creating L2 system identity...`

sigilioso (commented):
LGTM. I've just left a question

recipes/newrelic/infrastructure/super-agent/debian.yml
rubenruizdegauna: Good point! Changed in f6777411fa6b142a97ef758cb7ba86e18bf63739

sigilioso (approved):
LGTM

PR #1140, opened 2024-12-12
chore!: remove agent type references (NR-337618)

gsanchezgavier:
Remove references from Agent types.

⚠️  It needs released changes on the SA. I'll notify whenever is ready


rubenruizdegauna (approved):
LGTM. Tests will pass once the renaming is merged

PR #1160, opened 2025-02-11
fix: Update Mysql restart logic to handle different processes

Nandu-pns:


shashank-reddy-nr (approved):
Changes looks good to me

PR #1178, opened 2025-03-21
fix(debian-tests): Resolve consul-debian and cassandra-debian failures

Aashirwadjain:
### Description

This PR aims to resolve consul-debian and cassandra-debian failures

### Fixes: 
- The method of adding a repository GPG key using apt-key is deprecated. To solve this issue, we have added the GPG key directly to the trusted keyrings directory.
- Added signed-by to reference the GPG key, ensuring APT trusts the repository using the specified GPG key.
- Configured `openjdk-11-jdk` separately before installing cassandra.

### DIFF
--- a/test/deploy/linux/consul/install/debian/roles/configure/tasks/main.yml
+++ b/test/deploy/linux/consul/install/debian/roles/configure/tasks/main.yml
@@ -19,13 +19,28 @@
     state: latest
   become: yes
 
-- name: Add official HashiCorp GPG Key
-  shell: curl -fsSL https://apt.releases.hashicorp.com/gpg | sudo apt-key add -
-  become: true
+- name: Download HashiCorp GPG key
+  shell: |
+    curl -fsSL https://apt.releases.hashicorp.com/gpg -o /tmp/hashicorp.asc
+  args:
+    creates: /tmp/hashicorp.asc
+  become: yes
 
-- name: Add official HashiCorp Linux Repository
-  shell: apt-add-repository "deb [arch=amd64] https://apt.releases.hashicorp.com $(lsb_release -cs) main"
-  become: true
+- name: Add HashiCorp GPG key to APT keyring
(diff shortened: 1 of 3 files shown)


shashank-reddy-nr (approved):
LGTM

test/deploy/linux/consul/install/debian/roles/configure/tasks/main.yml
```diff
@@ -19,13 +19,28 @@
...
+    creates: /tmp/hashicorp.asc
+  become: yes
 
-- name: Add official HashiCorp Linux Repository
-  shell: apt-add-repository "deb [arch=amd64] https://apt.releases.hashicorp.com $(lsb_release -cs) main"
-  become: true
+- name: Add HashiCorp GPG key to APT keyring
+  shell: |
+    gpg --dearmor -o /usr/share/keyrings/hashicorp-archive-keyring.gpg /tmp/hashicorp.asc
+  args:
```
shashank-reddy-nr, line 32: @Aashirwadjain Can you please add reasoning for this change?

Aashirwadjain: @shashank-reddy-nr @vagrawal-newrelic 
The method of adding a repository GPG key using apt-key is deprecated. This approach is no longer recommended because it affects the entire system's trust settings. We can instead add the GPG key directly to the trusted keyrings directory to resolve this issue. So, the above change converts the downloaded GPG key to a format that APT can use and places it in the `/usr/share/keyrings/` directory.

Additionally, the change uses `signed-by=/usr/share/keyrings/hashicorp-archive-keyring.gpg` to tell APT to trust the repository using the specified GPG key. 

The changes avoids the deprecated apt-key method and it uses signed-by to explicitly tell APT which GPG key to use for verifying the repository, avoiding potential conflicts.

PR #1188, opened 2025-04-01
fix: Update recipe to automate nagios plugin install

Nandu-pns:
JIRA - https://new-relic.atlassian.net/browse/NR-380307

This PR automates the installation of Nagios plugins if they are not present on the user's host. We have noticed numerous installation failures due to missing plugins, even though this requirement is mentioned as a prerequisite in the installation UI.

### Test Results

- RHEL tests when plugins are not found.

image

- RHEL tests when plugins are already installed.

image

- Debian tests when plugins are not found.

image

- Debian tests when plugins are already installed.

Pasted Graphic 1

### DIFF
--- a/recipes/newrelic/infrastructure/ohi/nagios/debian.yml
+++ b/recipes/newrelic/infrastructure/ohi/nagios/debian.yml
@@ -102,7 +102,7 @@ install:
               tar -xzf nagios-plugins.tar.gz || { echo "Failed to extract plugins." >&2; exit 2; }
           
               # Extract version from filename to use in cd command
-              VERSION=$(echo "$DOWNLOAD_URL" | grep -oE 'nagios-plugins-([0-9]+\.[0-9]+\.[0-9]+)\.tar\.gz' | sed 's/nagios-plugins-//' | sed 's/.tar.gz//')
+              VERSION=$(echo "$DOWNLOAD_URL" | grep -oE 'nagios-plugins-([0-9]+\.[0-9]+(\.[0-9]+)?).tar\.gz' | sed 's/nagios-plugins-//' | sed 's/.tar.gz//')
               cd "nagios-plugins-$VERSION" || { echo "Failed to enter directory." >&2; exit 2; }
           
               # Install prerequisites if missing
--- a/recipes/newrelic/infrastructure/ohi/nagios/debian.yml
+++ b/recipes/newrelic/infrastructure/ohi/nagios/debian.yml
@@ -64,9 +64,61 @@ install:
           fi
         - |
           PLUGINS_EXIST=$(ls /usr/local/nagios/libexec | grep check_users | wc -l)
-          if [ $PLUGINS_EXIST -eq 0 ] ; then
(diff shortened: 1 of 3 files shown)


recipes/newrelic/infrastructure/ohi/nagios/rhel.yml
```diff
@@ -64,9 +64,58 @@ install:
...
+            if curl -sL "$DOWNLOAD_URL" -o nagios-plugins.tar.gz; then
+              tar -xzf nagios-plugins.tar.gz || { echo "Failed to extract plugins." >&2; exit 2; }
+          
+              # Extract version from filename to use in cd command
+              VERSION=$(echo "$DOWNLOAD_URL" | grep -oE 'nagios-plugins-([0-9]+\.[0-9]+\.[0-9]+)\.tar\.gz' | sed 's/nagios-plugins-//' | sed 's/.tar.gz//')
+              cd "nagios-plugins-$VERSION" || { echo "Failed to enter directory." >&2; exit 2; }
+          
+              # Install prerequisites if missing
+              echo "[INFO] Installing prerequisites... Please wait, this may take some time."
+              sudo yum install -y gcc make > /dev/null 2>&1
```
Nandu-pns, line 96: For Reviewers/Self-Reference -  While the Nagios plugins official support documentation ([link](https://support.nagios.com/kb/article/nagios-plugins-installing-nagios-plugins-from-source-569.html#RHEL)) outlines several prerequisite tools, this PR only installs the specific plugins that have caused errors due to their absence. This approach minimizes the initial impact. We can expand the scope to include all prerequisite tools listed in the documentation if future issues arise.

Nandu-pns: Furthermore I noticed that all of these prerequisites are essential during the installation of Nagios Core. Hence, by the time the plugins are being installed, these pre-requisite tools are typically already installed or available on the user hosts.

recipes/newrelic/infrastructure/ohi/nagios/debian.yml
```diff
@@ -63,10 +63,76 @@ install:
...
+              echo "Failed to fetch the latest Nagios plugins version from the download page." >&2
+              exit 2
+            fi
+          
+            # Construct the full download URL
+            DOWNLOAD_URL="https://nagios-plugins.org/download/$DOWNLOAD_URL"
+          
+            # Download and install plugins
+            if curl -sL "$DOWNLOAD_URL" -o nagios-plugins.tar.gz; then
+              tar -xzf nagios-plugins.tar.gz || { echo "Failed to extract plugins." >&2; exit 2; }
```
shashank-reddy-nr, line 102: do we need to keep version here? why are we not keeping version

Nandu-pns: This simply extracts the tar file and creates a directory with this name. Including a version is optional, but not necessary.

shashank-reddy-nr (approved):
LGTM

PR #1190, opened 2025-04-04
fix(varnish): Added default values for instance_name and params config file

rpaliwal1997:
**Issue :-** The INSTANCE_NAME variable in the Varnish Cache integration recipe was not correctly defaulting to new_relic when the user did not explicitly define it. This resulted in the following error during installation:

**Error:-**

> `DEBUG validation error encountered: exit status 1: [ERR] Error input validation failure: instance_name must not be blank
> `

[Jira Ticket](https://new-relic.atlassian.net/browse/NR-386250)

**Resolution:** 

- The issue has been resolved by explicitly handling the default value for INSTANCE_NAME in the recipe. The updated recipe ensures that the variable is never blank and defaults to new_relic when not provided.

- After fixing the recipe, varnish-cache-open-source-integration is installing successfully when user don’t define the value for instance_name.

**Testing after the fix:**

- Verified that the INSTANCE_NAME defaults to new_relic when not explicitly defined by the user.

- Confirmed that the /etc/newrelic-infra/integrations.d/varnish-config.yml file is correctly populated with the default value.

- Ensured that the error no longer occurs during installation.

**Case 1:-** When user don’t input the instance_name, default value  'new_relic' will be set, I tested this scenario, integration was successfully, please see below.

image

**Case 2:-** When user input the instance_name, I tested this scenario, integration was successfully, please see below.

image

**Case 3:-**  When user select ‘Yes(-y)’ flag to set all prompts as Yes.

image

### DIFF
--- a/recipes/newrelic/infrastructure/ohi/varnish/debian.yml
+++ b/recipes/newrelic/infrastructure/ohi/varnish/debian.yml
@@ -40,8 +40,10 @@ validationNrql: "SELECT count(*) from VarnishSample FACET entityGuid SINCE 10 mi
 inputVars:
   - name: "NR_CLI_INSTANCE_NAME"
     prompt: "Instance Name (used to identify data in New Relic)"
+    default: "new_relic"
   - name: "NR_CLI_PARAMS_CONFIG_FILE"
     prompt: "Location of varnish.params (omit to check defaults)"
+    default: "/etc/default/varnish/varnish.params"
 
 preInstall:
   info: |2
--- a/recipes/newrelic/infrastructure/ohi/varnish/debian.yml
+++ b/recipes/newrelic/infrastructure/ohi/varnish/debian.yml
@@ -39,10 +39,10 @@ validationNrql: "SELECT count(*) from VarnishSample FACET entityGuid SINCE 10 mi
 
 inputVars:
   - name: "NR_CLI_INSTANCE_NAME"
-    prompt: "Instance Name (used to identify data in New Relic)"
+    prompt: "Instance Name (used to identify data in New Relic, default name is : new_relic)"
     default: "new_relic"
   - name: "NR_CLI_PARAMS_CONFIG_FILE"
-    prompt: "Location of varnish.params (omit to check defaults)"
+    prompt: "Location of varnish.params (omit to check defaults) default location is : /etc/default/varnish/varnish.params"
     default: "/etc/default/varnish/varnish.params"
 
 preInstall:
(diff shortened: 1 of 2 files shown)


recipes/newrelic/infrastructure/ohi/varnish/debian.yml
```diff
@@ -39,9 +39,11 @@ validationNrql: "SELECT count(*) from VarnishSample FACET entityGuid SINCE 10 mi
 
 inputVars:
   - name: "NR_CLI_INSTANCE_NAME"
-    prompt: "Instance Name (used to identify data in New Relic)"
+    prompt: "Instance Name (used to identify data in New Relic, default name is : new_relic)"
```
vagrawal-newrelic, line 42: @rpaliwal1997 can u please make this prompt also same as `NR_CLI_PARAMS_CONFIG_FILE` prompt so that both are consistent, 
`Instance Name (used to identify data in New Relic) default name is : new_relic`

recipes/newrelic/infrastructure/ohi/varnish/debian.yml
```diff
@@ -39,9 +39,11 @@ validationNrql: "SELECT count(*) from VarnishSample FACET entityGuid SINCE 10 mi
...
+    default: "new_relic"
```
shashank-reddy-nr, line 43: Here, the term "Instance name" refers to the entity name that will be created in New Relic, correct? I'm considering whether we should use a more specific name instead of just referring to it as "New Relic.

shashank-reddy-nr: I just want to check whether is this entity name that is being created? I undertsood that you copied from On-Host but whether is it entity_name or not is the question, if yes, I think we should be more specific I think, As customers will not understand what this entity is if we keep New Relic

recipes/newrelic/infrastructure/ohi/varnish/debian.yml
rpaliwal1997: Hi @vagrawal-newrelic , It will look weird if I keep like this 'Instance Name (used to identify data in New Relic) default name is : new_relic' as default value also will be added after this prompt, might be confusing to the user(It will be look like 'Instance Name (used to identify data in New Relic) default name is : new_relic new_relic').

shashank-reddy-nr (approved):
LGTM

recipes/newrelic/infrastructure/ohi/varnish/debian.yml
rpaliwal1997: Hi @shashank-reddy-nr , Since instance_name is not the original entity like host_name but rather an identifier in the logs to help identify data, keeping new_relic as the default value seems appropriate here.

PR #1193, opened 2025-04-07
feat(ac): install only one agent (infra|otel)

rubenruizdegauna:
This PR enforces installing only one HOST instrumentation agent in the Agent Control installation. Infra Agent or Opentelemetry Collector, but not both.


vjripoll (approved):
LGTM

PR #1194, opened 2025-04-10
fix(regression-tests): Exclude tests that use deprecated os.

vagrawal-newrelic:
excluded tests that are using deprecated OS. 
Please find the details of the excluded tests and further analysis on the deprecated os in this [doc](https://docs.google.com/document/d/1qv885CXmdi9E4F6E69Wis4EXb7BX3kRwP0z2Hew-aKw/edit?tab=t.0)

Jira ticket:  https://new-relic.atlassian.net/browse/NR-396172

image
image

PR #1199, opened 2025-04-15
fix(deprecated-tests): Moved deprecated CentOS tests to separate folder

shashank-reddy-nr:
### Changes:
- Since the CentOS (7, 8) AMI images we use for regression tests have been deprecated and we currently lack relevant verified standalone CentOS 7, 8 images in the AMI marketplace which are compatible with demo deployer, we plan to move these test cases into a deprecated folder until we find suitable alternative AMI images.

- I have reverted a few changes from [PR#1194](https://github.com/newrelic/open-install-library/pull/1194) that were initially made to bypass CentOS 7, 8 regression tests. The reason for the reversion is that while we are bypassing these tests on GitHub regression jobs(non regression, Non regression EU), the PR builds are still running these CentOS tests and failing in PRs. As a result, authors may be unable to determine whether test failures are genuine, since the daily scheduled job is skipping them due to the aforementioned PR#1194.


pranav-new-relic:
👍 

looks good to me; the only consideration I'm trying to make is - to achieve this, we can either go with the solution you've proposed in this PR, or we can update the validate workflow similar to the non-regression (US, EU) workflows to exclude test files (using the script Vishal added), so that the validate workflow triggered on PR runs _also_ excludes the tests needing exclusion, along with the runs of non-regression (US, EU). In my opinion, I find this a cleaner and easier solution to basically exclude these tests from everywhere and bring them back when needed but I'd also like to know @vagrawal-newrelic's opinion since he's the author of #1194 :)

also I think the purpose of different folders for US, EU in the new deprecated folder is completely fine

shashank-reddy-nr:
I spoke with Vishal, and we agreed on the point that adding unnecessary code to bypass would make the workflow unnecessarily complex. It's better to implement these changes to keep the tests in a separate folder.

pranav-new-relic:
👍👍 I agree, too

PR #1200, opened 2025-04-17
fix(Cassandra-test): fix failing Cassandra gpg key error.

vagrawal-newrelic:
Context:
Cassandra test was failing because the apt package was not able to find the gpg key as it was being added using `apt-key add` method which is an old method of gpg key addition, which has been deprecated in Cassandra. 
Fix:
Updated the gpg key addition method and correct Cassandra repository path. Followed the steps provided in Cassandra [official doc](https://cassandra.apache.org/doc/latest/getting_started/installing.html#debian) for installing Cassandra and updated the ansible script according to that.

Local test success validation:
image


shashank-reddy-nr (approved):
LGTM

Nandu-pns (approved):

PR #1201, opened 2025-04-17
fix(linux): Updated infra install command to include upgrade as well

shashank-reddy-nr:
GSTE: https://new-relic.atlassian.net/browse/NR-404232

### Changes:
- This PR addresses a bug in the New Relic Infrastructure agent installation process for Linux. The current command (yum -y -q install newrelic-infra) fails to upgrade the agent on AWS linux hosts where an older version is already installed. Instead, it only installs the agent when it's not present. To ensure correct behavior, the installation process has been modified. The revised command now checks for the agent's presence and either upgrades it or installs the latest version appropriately:

`yum list installed newrelic-infra >/dev/null 2>&1 && yum update newrelic-infra -y -q || yum install newrelic-infra -y -q`

This fix will ensure the agent is either upgraded or installed consistently.

image

### DIFF
--- a/recipes/newrelic/infrastructure/awslinux.yml
+++ b/recipes/newrelic/infrastructure/awslinux.yml
@@ -205,7 +205,7 @@ install:
           fi
           curl -s $REPO_URL -o /etc/yum.repos.d/newrelic-infra.repo
           yum -y -q makecache --disablerepo='*' --enablerepo='newrelic-infra'
-          yum -y -q install newrelic-infra
+          yum list installed newrelic-infra >/dev/null 2>&1 && sudo yum update newrelic-infra -y -q || sudo yum install newrelic-infra -y -q
       silent: true
 
     restart:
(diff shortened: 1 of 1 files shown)


ujwalsah:
There seems to a conflict. Please check

shashank-reddy-nr:
Yes, this issue arose because your team made changes in the same line, and those changes were merged into the main branch yesterday. I have now resolved the conflict. Please take a look.

mbolla22 (approved):
Looks good for me 
Small doubt , is this issue is only with Awslinux or for all remaining linux distributions ?

mbolla22:
Small doubt , is this issue is only with Awslinux or for all remaining linux distributions ?

shashank-reddy-nr:
Nice catch. I see same command in centOS_rhel as well. Let me try to reproduce on rhel env and if it exists, I will update the recipe.

shashank-reddy-nr:
@mbolla22 in CentOS RHEL hosts the current command is working as expected and upgradation is happening. So, we are good to merge this change as we are having issue only on AWS linux hosts.

image

PR #1202, opened 2025-04-22
fix(RHEL8): Updated deprecated AMI images with new ones

shashank-reddy-nr:
### Changes:
- Updated the deprecated RHEL8 AMI images to new ones.

Before changes Error:
image


Aashirwadjain (approved):
Changes Looks good to me

PR #1208, opened 2025-05-07
feat: remove newrelic-opentelemetry-collector recipe

kb-newrelic:
### Summary
- Removing `nr-otel-collector` recipes as it is no longer maintained in the form it was originally added here. The ownership of the project has changed and with the technical and directional change of the product, we'll want to re-integrate at a later point in time with a more holistic vision regarding OTel as a while.
- Release [0.8.5](https://github.com/newrelic/nrdot-collector-releases/releases/tag/nr-otel-collector-0.8.5) was the last `nr-otel-collector` release published to the referenced deb/rpm/zypper repos, so while the install script is not broken, it will grow stale over time


pranav-new-relic:
For the reviewer a.k.a @pranav-new-relic 
- all references to files named with `otel` / `newrelic-opetelemetry-collector` (the name of the recipe) have been eliminated
image
image

- the test runs are failing because of the Validate Recipe workflow's `git diff` identifying these test files to have been changed (but the change is a delete; which is why in subsequent steps, the "changed" recipe identified by this command is not identifiable by the GitHub action that tries accessing the file at the identified path of the changed recipe, since it was deleted
- keep in mind that this PR's merge and release should immediately be followed by a revert of changes in the CLI https://github.com/newrelic/newrelic-cli/pull/1469/files https://github.com/newrelic/newrelic-cli/pull/1474/files and then, the CLI shall be released (**DO NOT** release the CLI pointing to the version of OIL with this PR's changes merged without reverting the aforementioned changes in the CLI)

Nandu-pns (approved):

PR #1220, opened 2025-06-04
fix: Replaced deprecated ubuntu18.04 AMI images with alt ones

shashank-reddy-nr:
### Changes:
- Replaced deprecated ubuntu18.04 AMI images with alternative ones


pranav-new-relic (approved):
LGTM

PR #1225, opened 2025-06-20
fix: Updated deprecated ubuntu 20 AMI images

shashank-reddy-nr:
### Changes:
- Updated deprecated ubuntu 20.04 AMI images to new images
New Free Image used: `ubuntu-pro-server/images/hvm-ssd/ubuntu-focal-20.04-arm64-pro-server-20250530`

- Updated apache maven version to 3.9.10 from 3.8.8. As it is deprecated a couple of days back.


Aashirwadjain (approved):
Changes Looks good to me.

PR #1227, opened 2025-06-23
fix: Update APT repo configuration steps in dotnet linux recipe

nr-ahemsath:
This PR updates the .NET agent Linux install recipe to use a new public signing key and also replace the use of `apt-key add` (which is deprecated) with the current best practice of saving the key to a specific file and configuring the APT repo definition to be signed by that file.

Note: this is in draft status until the .NET agent does it's next release, which we plan to do early this week.


shashank-reddy-nr (approved):
LGTM

PR #1229, opened 2025-06-25
fix: Update APT repo configuration steps in dotnet linux recipe (#1227)

shashank-reddy-nr:
This PR updates the .NET agent Linux install recipe to use a new public signing key and also replace the use of apt-key add (which is deprecated) with the current best practice of saving the key to a specific file and configuring the APT repo definition to be signed by that file.

Note: this is in draft status until the .NET agent does it's next release, which we plan to do early this week.

### DIFF
--- a/recipes/newrelic/apm/dotNet/linux-systemd.yml
+++ b/recipes/newrelic/apm/dotNet/linux-systemd.yml
@@ -267,8 +267,8 @@ install:
           IS_YUM_INSTALLED=$(which yum 2>&1 || true)
           if [[ -z "$IS_YUM_INSTALLED" ]]
           then
-            echo 'deb http://apt.newrelic.com/debian/ newrelic non-free' | sudo tee /etc/apt/sources.list.d/newrelic.list > /dev/null
-            wget -q -O- {{.NEW_RELIC_DOWNLOAD_URL}}548C16BF.gpg | sudo apt-key add -
+            echo 'deb [signed-by=/usr/share/keyrings/newrelic-apt.gpg] http://apt.newrelic.com/debian/ newrelic non-free' | sudo tee /etc/apt/sources.list.d/newrelic.list > /dev/null
+            wget -q -O- {{.NEW_RELIC_DOWNLOAD_URL}}NEWRELIC_APT_2DAD550E.public | sudo gpg --dearmor -o /usr/share/keyrings/newrelic-apt.gpg
             sudo apt-get -o Acquire::Check-Valid-Until=false update >/dev/null
             sudo apt-get install newrelic-dotnet-agent -y -qq > /dev/null
           else


chynesNR (approved):

nrcventura (approved):

nr-ahemsath:
Hi @shashank-reddy-nr:  I haven't tested these changes using guided install specifically.  I haven't worked on these recipes since 2021 so I don't remember what is required and how to do it, but I do remember it being a bit complex.

We made the same change in our agent's CI, specifically a test that runs after we deploy a new agent version to `apt.newrelic.com`: https://github.com/newrelic/newrelic-dotnet-agent/pull/3163/files

And that test passed when we did an agent release/deploy yesterday: https://github.com/newrelic/newrelic-dotnet-agent/actions/runs/15860680315/job/44717638024

Let me know if this is sufficient.  Otherwise I'm happy to test the new guided install recipe locally but I'll need some guidance on how to do that.

nr-ahemsath:
@shashank-reddy-nr Cool, thanks.  Is there an option to make it work with a staging account and license key?

shashank-reddy-nr:
You can use staging accounts by adding NEW_RELIC_REGION=staging in above command. But on there is WAF restrictions on staging and from AWS external world we can't access. So, I rrquest you to test using prod.

nr-ahemsath:
So, I'm running into the following snags when trying to test these changes locally:

1. I'm on Windows, so the most convenient way for me to test on Linux is to either use a WSL2 VM, or a Docker container.
2. The install recipe depends on `systemd` being configured and running on the Linux system in order to work.  It will not attempt to install the agent if it can't find any systemd services for .NET applications.
3. Additionally, the install recipe depends on the "dotnet introspector" which is a tool that our team wrote back when Guided Install was first developed approx. four years ago.  This tool was written in an old version of .NET Core (not sure which yet, I'm checking into that) that depends on v1.x of `libssl`.
4. Ubuntu 22.04+ do not have v1.x of libssl.  Ubuntu 20.04 does have v1.x of libssl.
5. However, since Ubuntu 20.04 is EOL, I can't create an AWS EC2 instance running 20.04.

So, I can't test on WSL or Docker because they don't have systemd, and I can't test on AWS EC2 because it won't run Ubuntu 20.04.

By the way, the missing libssl v1 dependency means that guided install as it is currently implemented for .NET on Linux simply will not work for anybody on a modern, supported Ubuntu.  I checked with our GTS team and they could only find one instance of a customer complaining about that failure.  So it looks like almost nobody is using guided install on Linux for .NET.

**Long term**, we need to fix the introspector to work on modern Linux distros.  We also need to update the integration tests in this repo to test with modern Linux distros and using a supported version of the .NET runtime (8.0+)
**Short term**, given the low usage of this feature, I don't think I can afford to spend any more time trying to test this locally.

PR #1230, opened 2025-06-25
feat: change ac recipes to use new auth cli

vjripoll:
Move from direct curls calls with lot of logic to the new auth cli that simplifies the script into de recipies.


sigilioso (approved):
LGTM 🚀

PR #1233, opened 2025-06-26
chore(recipes/php): update debian recipe

lavarou:
Update PHP's debian recipe to avoid using deprecated `apt-key` utility as well as legacy public key that uses week algorithm. This aligns the recipe with Debian/Ubuntu install instructions provided to customers in public facing documentation.


pranav-new-relic:
LGTM, as the tests consuming the debian recipe also seem to work perfectly fine 👍

PR #1248, opened 2025-07-17
fix: Fixed Infra linux 2 ver specific install failure

shashank-reddy-nr:
### Changes:
- Fixed version specific install failure on amazon linux 2 issue.

Ticket: https://new-relic.atlassian.net/browse/NR-308200

### Fix Test Results:
**1: AMazon Linux 2 :**
image

**2: Amazon linux 2023:**
image

### DIFF
--- a/recipes/newrelic/infrastructure/awslinux.yml
+++ b/recipes/newrelic/infrastructure/awslinux.yml
@@ -223,17 +223,16 @@ install:
             echo "FIPS mode not enabled. Installing $PACKAGE_NAME${VERSION_SUFFIX}..."
           fi
 
-          # Install specific version if provided, otherwise install latest
           if [ -n "$AGENT_VERSION" ]; then
             # Install specific version (handles install/update)
-            INSTALL_OUTPUT=$(yum -y install $PACKAGE_NAME${VERSION_SUFFIX} 2>&1)
-            # If install did not complete, attempt to downgrade the package.
-            if ! echo "$INSTALL_OUTPUT" | grep -q "Complete!"; then
+            yum -y -q install $PACKAGE_NAME${VERSION_SUFFIX}
+            # If specific version was not installed, attempt to downgrade the package.
+            if ! rpm -q $PACKAGE_NAME${VERSION_SUFFIX} >/dev/null 2>&1; then
                 yum -y -q downgrade $PACKAGE_NAME${VERSION_SUFFIX}
             fi
           else
             # No version specified - install or update to latest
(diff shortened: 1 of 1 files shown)


recipes/newrelic/infrastructure/awslinux.yml
```diff
@@ -225,8 +225,12 @@ install:
 
           # Install specific version if provided, otherwise install latest
           if [ -n "$AGENT_VERSION" ]; then
-            # Install specific version (handles install/update/downgrade)
-            yum -y -q install $PACKAGE_NAME${VERSION_SUFFIX} --allowerasing
```
shashank-reddy-nr, line 229: The --allowerasing option is NOT supported across all the Amazon Linux versions (linux 1,2) it seems.

recipes/newrelic/infrastructure/awslinux.yml
```diff
@@ -225,8 +225,12 @@ install:
...
+            # Install specific version (handles install/update)
+            INSTALL_OUTPUT=$(yum -y install $PACKAGE_NAME${VERSION_SUFFIX} 2>&1)
```
shashank-reddy-nr, line 229: Trying to install or upgrade first, as install command will not work for downgrading purpose without --alloweasing flag. So, conditionally doing downgrade if required.

recipes/newrelic/infrastructure/awslinux.yml
```diff
@@ -225,8 +225,12 @@ install:
...
+            # If install did not complete, attempt to downgrade the package.
+            if ! echo "$INSTALL_OUTPUT" | grep -q "Complete!"; then
```
vagrawal-newrelic, line 231: @shashank-reddy-nr May be checking the Installation of the package using "Complete!" keyword pattern matching might not be a very robust validation. As it is leading to a very tight coupling with the installation output!!
Rather than we can use `rpm` to check directly weather the particular package is installed or not something like this
image

And RPM works on all version of Amazon linux
Also RPM is a low level utility tool so running a command would be faster than Searching a keyword in Installation output using grep

Just a suggestion!!!

shashank-reddy-nr: Nice suggestion @vagrawal-newrelic. I will give a try with this. Thanks for suggesting alternative best approach as well.

shashank-reddy-nr: Addressed this

pranav-new-relic (approved):
nice find

vagrawal-newrelic (approved):
LGTM

PR #1265, opened 2025-09-23
feat: add public key url to config

danielorihuela:
Configures the public key url for onhost.

#related-to NR-462134

### DIFF
--- a/recipes/newrelic/infrastructure/agent-control/debian.yml
+++ b/recipes/newrelic/infrastructure/agent-control/debian.yml
@@ -381,7 +381,7 @@ install:
             fi
           fi
       vars:
-        NEW_RELIC_AGENT_VERSION: "0.48.0"
+        NEW_RELIC_AGENT_VERSION: "1.0.0"
       silent: true
 
     # If configured to do so, migrate the newrelic-infra configuration for usage with New Relic Agent Control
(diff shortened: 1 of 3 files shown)


recipes/newrelic/infrastructure/agent-control/debian.yml
```diff
@@ -478,10 +478,13 @@ install:
         - |
           if [ "{{.NEW_RELIC_REGION}}" = "STAGING" ]; then
             sed -i 's/\(endpoint: https:\/\/opamp.\).*/\1'"staging-service.newrelic.com\/v1\/opamp"'/' /etc/newrelic-agent-control/config.yaml
+            sed -i '/^fleet_control:/a \  signature_validation:\n    public_key_server_url: https:\/\/staging-publickeys.newrelic.com\/r\/blob-management\/GLOBAL\/AgentConfiguration\/jwks.json' /etc/newrelic-agent-control/config.yaml
```
gsanchezgavier, line 481: nit, it would be a bit more clear if we replace a placeholder like <PLACEHOLDER> with the url , perhaps is it hard to do the change now + it would be nice to change all this script to use some more clear tool like yq

gsanchezgavier (approved):

PR #1284, opened 2025-11-13
feat: add service account credentials support for non-admin Windows installs

NSSPKrishna:
~Add NRIA_USER and NRIA_PASS input variables for service account authentication~
 ~- Update permission checks to allow non-admin users with valid credentials~
  ~- Modify MSI installation to pass credentials as parameters~
 ~- Ensure installation fails if non-admin user lacks required credentials~
 ~- Maintain backward compatibility for admin user installations~

~IMPORTANT:  Non-admin users now require NRIA_USER and NRIA_PASS parameters~

Infra agent will start to support the update of the agent, removing the agent from the OIL will break the infra agent setup if already configured. Removing the uninstallation from the reciepe.


Nandu-pns:
Confirmed with @nravada — the API key exposed in this PR, flagged by the GitHub secret scanning [alert](https://github.com/newrelic/open-install-library/security/secret-scanning/4), has been revoked/deleted and is no longer valid.

PR #1327, opened 2026-02-18
feat(ac): remove agents

paologallinaharbur:
Removing agents and the ` --agent-set` flag option.

This can be merged only once all agents can on OCI packages


recipes/newrelic/infrastructure/agent-control/rhel.yml
gsanchezgavier: these looks like left overs too , now that we are not installing any other recipe. perhaps unrelated too

paologallinaharbur: some of them are used, I am going to remove awk and cat only

gsanchezgavier (dismissed):

gsanchezgavier (approved):

PR #1334, opened 2026-03-16
feat: modify ac post_install text and remove old newrelic_cli_option from test

alvarocabanas:
- modify ac post_install text and remove old newrelic_cli_option from test


vjripoll (approved):
LGTM

PR #1353, opened 2026-04-13
feat(nginx): add nginx nrdot recipes

sairaj18:
Summary
                                                                                                                                            
  - Add two new NRDOT OpenTelemetry Collector recipes for NGINX monitoring:
    - recipes/newrelic/infrastructure/nrdot/nginx/debian.yml — Debian/Ubuntu targets                                                        
    - recipes/newrelic/infrastructure/nrdot/nginx/rhel.yml — RHEL, CentOS, Amazon Linux 2/2023 targets                                      
  - Add two new NRDOT recipes for NGINX Plus monitoring (prometheus-exporter based):                                                        
    - recipes/newrelic/infrastructure/nrdot/nginx-plus/debian.yml — Debian/Ubuntu targets                                                   
    - recipes/newrelic/infrastructure/nrdot/nginx-plus/rhel.yml — RHEL, CentOS, Amazon Linux 2/2023 targets                                 
  - Add Ansible deploy playbooks under test/deploy/linux/nrdot-nginx/ (debian + rhel variants) that install nginx and expose stub_status on 
  127.0.0.1:8080/nginx_status                                                                                                               
  - Add 6 E2E test definition files under test/definitions/nrdot/nginx/ covering: Ubuntu 22.04, Ubuntu 24.04, RHEL 9, Amazon Linux 2, Amazon
   Linux 2023 (x86_64 + arm64)                                                                                                              
                  
Testing details:                                                                                                                                 
                  
All 4 recipes were manually validated end-to-end across 8 EC2 instances before this PR. Each instance was tested against UC1 (fresh install), UC3 (idempotent re-run), and UC2 (additive: host collector running + nginx recipe added on top). NRQL validation confirmed nginx.% / nginxplus_% metrics flowing with correct host.id on all instances.

### DIFF
--- a/recipes/newrelic/infrastructure/nrdot/nginx-plus/debian.yml
+++ b/recipes/newrelic/infrastructure/nrdot/nginx-plus/debian.yml
@@ -43,20 +43,10 @@ preInstall:
       exit 1
     fi
 
-    # Exit 1 if this is not NGINX Plus (standard nginx is handled by nrdot-collector-nginx).
-    # nginx -v outputs to stderr; NGINX Plus builds include "nginx-plus" in the version string.
+    # Exit 1 if this is not NGINX Plus — standard nginx is handled by nrdot-collector-nginx.
     if ! nginx -v 2>&1 | grep -qi "nginx-plus"; then
       exit 1
     fi
-
-    # Exit 1 if the prometheus exporter is not reachable on the default port.
-    # The exporter must be running before the recipe can configure the collector.
-    code=$(curl -s -o /dev/null -w "%{http_code}" "http://127.0.0.1:9113/metrics" 2>/dev/null || echo "000")
-    if [ "$code" = "200" ]; then
-      exit 0
-    fi
-
-    exit 1
 
 inputVars:
   - name: NR_CLI_NGINX_PLUS_PROMETHEUS_ENDPOINT
@@ -124,6 +114,20 @@ install:
         - |
           if ! nginx -v 2>&1 | grep -qi "nginx-plus"; then
             echo "This recipe requires NGINX Plus. Standard NGINX is handled by nrdot-collector-nginx." >&2
+            exit 132
+          fi
+        # Exit 132: NGINX Prometheus Exporter must be running before installation.
+        - |
+          PROMETHEUS_ENDPOINT="{{.NR_CLI_NGINX_PLUS_PROMETHEUS_ENDPOINT}}"
+          if [ -z "$PROMETHEUS_ENDPOINT" ]; then
+            PROMETHEUS_ENDPOINT="127.0.0.1:9113"
+          fi
+          PROMETHEUS_ENDPOINT=$(echo "$PROMETHEUS_ENDPOINT" | sed 's|^https\?://||' | sed 's|/.*||')
+          code=$(curl -s -o /dev/null -w "%{http_code}" "http://${PROMETHEUS_ENDPOINT}/metrics" 2>/dev/null || echo "000")
+          if [ "$code" != "200" ]; then
+            echo "Could not reach the NGINX Prometheus Exporter at http://${PROMETHEUS_ENDPOINT}/metrics (HTTP ${code})." >&2
+            echo "Install and start nginx-prometheus-exporter before re-running this recipe." >&2
+            echo "See https://github.com/nginx/nginx-prometheus-exporter?tab=readme-ov-file#running-the-exporter-binary for instructions." >&2
             exit 132
           fi
 
--- a/recipes/newrelic/infrastructure/nrdot/nginx-plus/debian.yml
+++ b/recipes/newrelic/infrastructure/nrdot/nginx-plus/debian.yml
@@ -1,0 +1,427 @@
+# Visit our schema definition for additional information on this file format.
+# https://github.com/newrelic/open-install-library/blob/main/docs/recipe-spec/recipe-spec.md#schema-definition
+
+name: nrdot-collector-nginx-plus
+displayName: NRDOT Collector - NGINX Plus
+description: New Relic install recipe for NGINX Plus monitoring via the NRDOT OpenTelemetry Collector (Debian/Ubuntu)
+repository: https://github.com/newrelic/nrdot-collector-releases
+
+installTargets:
+  - type: host
+    os: linux
+    platformFamily: debian
+    platformVersion: "(10|11|12)\\.*"
+  - type: host
+    os: linux
+    platform: ubuntu
+    platformVersion: "(18|20|22|24)\\.04"
+
+keywords:
+  - NRDOT
+  - OpenTelemetry
+  - OTel
+  - Collector
+  - NGINX Plus
+  - nginx-plus
+  - Debian
+  - Ubuntu
+
+processMatch:
+  - nginx
+
+validationNrql: "SELECT count(*) FROM Metric WHERE metricName LIKE 'nginxplus_%' SINCE 10 minutes ago"
+
+preInstall:
+  info: |2
+      To capture data from the NGINX Plus integration, you'll first need to install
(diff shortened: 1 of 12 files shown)


recipes/newrelic/infrastructure/nrdot/nginx-plus/debian.yml
```diff
@@ -0,0 +1,427 @@
...
+
+    # Exit 1 if this is not NGINX Plus (standard nginx is handled by nrdot-collector-nginx).
+    # nginx -v outputs to stderr; NGINX Plus builds include "nginx-plus" in the version string.
+    if ! nginx -v 2>&1 | grep -qi "nginx-plus"; then
+      exit 1
+    fi
+
+    # Exit 1 if the prometheus exporter is not reachable on the default port.
+    # The exporter must be running before the recipe can configure the collector.
+    code=$(curl -s -o /dev/null -w "%{http_code}" "http://127.0.0.1:9113/metrics" 2>/dev/null || echo "000")
```
Nandu-pns: It looks the current `requireAtDiscovery` step exits 1 if the prometheus exporter isn't reachable at `127.0.0.1:9113`. Any non-zero exit from `requireAtDiscovery` causes the recipe to be silently excluded from guided install with no user-facing message. This means if someone has NGINX Plus but hasn't set up the exporter yet, the recipe just won't show up — which could be a confusing experience.                                                    
                  
 Would it make sense to move the exporter check out of `requireAtDiscovery` and move into `assert_pre_req`  and add an error message? That way the recipe still shows up in guided install when NGINX Plus is detected, and if the exporter isn't running they get a clear error message telling them what to set up before re-running. Applies to both `debian.yml` and `rhel.yml`. Thanks !

sairaj18: Good catch. I've addressed the above in [this](https://github.com/newrelic/open-install-library/pull/1353/changes/7ed0de943bf71b3983134c26b12b144c7d2aaacf) commit.

Nandu-pns: Similar to the nginx-plus change — the `requireAtDiscovery` step for nginx (non-plus) recipes exits 1 if `stub_status` isn't reachable, which would silently exclude the recipe from guided install for users who have nginx installed but haven't configured stub_status yet. Worth addressing this the same way too ?

sairaj18: I have addresses the issue in nrdot nginx recipes.

recipes/newrelic/infrastructure/nrdot/nginx/rhel.yml
```diff
@@ -0,0 +1,450 @@
...
+  info: |2
+      To capture data from the NGINX integration, you'll first need to configure
+      the HTTP stub status module in your nginx config before running this install.
+      See https://docs.newrelic.com/docs/opentelemetry/integrations/nginx/nginx-otel-host/ for instructions.
+
+  requireAtDiscovery: |
+    # Exit 1 (skip recipe) if nginx is not installed.
+    if ! command -v nginx > /dev/null 2>&1; then
+      exit 1
+    fi
```
Nandu-pns, line 60: On `nginx/debian.yml` line 44 and `nginx/rhel.yml` line 60 (end of requireAtDiscovery block):

The `requireAtDiscovery` block only checks if nginx is installed, but doesn't exclude NGINX Plus. Since both `nginx/debian.yml` and `nginx-plus/debian.yml` use **processMatch: [nginx]**, a system running NGINX Plus will match both recipes during discovery. The standard recipe will then incorrectly attempt stub_status-based monitoring on NGINX Plus.

Maybe we should consider adding an exclusion as below ? Could you please double check ? 

 ```
Please add an NGINX Plus exclusion:
 # Exit 1 if this is NGINX Plus — handled by nrdot-collector-nginx-plus.
 if nginx -v 2>&1 | grep -qi "nginx-plus"; then
  exit 1
 fi
```

**Comment by Nandu-pns** [comment]
This isn't a blocker but I see the nginx-plus recipes (`nginx-plus/debian.yml`, `nginx-plus/rhel.yml`) have no test coverage. It would be good to add automated test definitions (similar to the existing `test/definitions/nrdot/nginx/` ones) so that future changes to these recipes are automatically validated by CI. This helps catch regressions early without relying on manual testing.

File: recipes/newrelic/infrastructure/nrdot/nginx-plus/rhel.yml [hunk 3]
```diff
@@ -0,0 +1,441 @@
+# Visit our schema definition for additional information on this file format.
+# https://github.com/newrelic/open-install-library/blob/main/docs/recipe-spec/recipe-spec.md#schema-definition
+
+name: nrdot-collector-nginx-plus-rhel
```
Nandu-pns: The recipe name `nrdot-collector-nginx-plus-rhel` has a `-rhel` suffix that none of the other recipes in this PR use — `nrdot-collector-nginx`, `nrdot-collector-nginx-plus` (debian) are both platform-agnostic     
names with installTargets handling the platform differentiation. If an install flow references this recipe by name, it would need to call `nrdot-collector-nginx-plus` for Debian and  `nrdot-collector-nginx-plus-rhel` for RHEL, which is inconsistent. I'd suggest renaming to `nrdot-collector-nginx-plus` to align with the rest.

recipes/newrelic/infrastructure/nrdot/nginx-plus/debian.yml
Nandu-pns: It looks this is partially fixed. The `stub_location` [check](https://github.com/newrelic/open-install-library/pull/1353/changes#diff-fcf77b40566b86c043f7a985ceedb1da3ffe55a9ff85a99ba5c3c47e6ef05fb8R67-R68) is still in `requireAtDiscovery` for both `nginx/debian.yml` and `nginx/rhel.yml`. Similar to the nginx-plus fix, if `stub_status` isn't configured, the recipe will silently disappear from guided install with no user-facing message.
                                                                                                                                                                                                                 
If I'm understanding this correct, the `assert_pre_req` stub_status validation only runs when `NR_CLI_NGINX_STATUS_URL` is explicitly provided, so the blank URL case would have no validation until deep in the install task. Would it make sense to move the `stub_location` check into `assert_pre_req` as a fallback for when no URL is provided, so users always get a clear error message instead of a silent skip ?

recipes/newrelic/infrastructure/nrdot/nginx-plus/rhel.yml
sairaj18: Yes, this should be `nrdot-collector-nginx-plus`. I've updated it.

recipes/newrelic/infrastructure/nrdot/nginx/rhel.yml
sairaj18: Thanks. I've addressed this issue.

sairaj18:
> This isn't a blocker but I see the nginx-plus recipes (`nginx-plus/debian.yml`, `nginx-plus/rhel.yml`) have no test coverage. It would be good to add automated test definitions (similar to the existing `test/definitions/nrdot/nginx/` ones) so that future changes to these recipes are automatically validated by CI. This helps catch regressions early without relying on manual testing.

I'm not sure if we will be able to add the tests for `nginx-plus` because it needs installing licensed packages which we don't have access to.

recipes/newrelic/infrastructure/nrdot/nginx-plus/debian.yml
sairaj18: I've addressed this in the latest commit.

Nandu-pns (approved):
Nice addition ! Changes looks good to me.

PR #1354, opened 2026-04-15
feat(rabbitmq): add rabbitmq nrdot recipes

ujwalsah:
## Summary                                                                                                                                                                                                                                                                 
  - Add NRDOT Collector recipes for RabbitMQ monitoring on Debian/Ubuntu and RHEL/CentOS/Amazon Linux                                                                                                                                                                               
  - Add deployer test definitions and Ansible deploy scripts for CI validation                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 
  ## Test coverage                                                                                                                                                                                                                                                           
  - `rhel9-amd64` — RHEL 9                                                                                                                                                                                                                                                   
  - `ubuntu22-amd64` — Ubuntu 22.04                                                                                                                                                                                                                                          
  - `ubuntu24-amd64` — Ubuntu 24.04
  - `al2023-amd64` — Amazon Linux 2023

### DIFF
--- a/recipes/newrelic/infrastructure/nrdot/rabbitmq/debian.yml
+++ b/recipes/newrelic/infrastructure/nrdot/rabbitmq/debian.yml
@@ -1,0 +1,295 @@
+# Visit our schema definition for additional information on this file format.
+# https://github.com/newrelic/open-install-library/blob/main/docs/recipe-spec/recipe-spec.md#schema-definition
+
+name: nrdot-rabbitmq
+displayName: NRDOT RabbitMQ
+description: New Relic install recipe for RabbitMQ monitoring via NRDOT Collector on Debian/Ubuntu.
+repository: https://github.com/newrelic/nrdot-collector-releases
+
+installTargets:
+  - type: host
+    os: linux
+    platformFamily: debian
+  - type: host
+    os: linux
+    platform: ubuntu
+
+keywords:
+  - RabbitMQ
+  - NRDOT
+  - OpenTelemetry
+  - OTel
+  - Messaging
+  - Queue
+  - Linux
+  - Debian
+  - Ubuntu
+
+processMatch: []
+
+preInstall:
+  discoveryMode:
+    - targeted
+
+inputVars:
+  - name: NR_CLI_RABBITMQ_DEPLOYMENT_NAME
+    prompt: "RabbitMQ deployment name (used to identify this instance in New Relic): "
+  - name: NR_CLI_RABBITMQ_ENDPOINT
+    prompt: "RabbitMQ management API endpoint (e.g. http://localhost:15672): "
+    default: "http://localhost:15672"
+  - name: NR_CLI_RABBITMQ_USERNAME
+    prompt: "RabbitMQ username: "
+    default: "admin"
+  - name: NR_CLI_RABBITMQ_PASSWORD
+    prompt: "RabbitMQ password: "
+    secret: true
+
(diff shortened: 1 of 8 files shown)


recipes/newrelic/infrastructure/nrdot/rabbitmq/debian.yml
```diff
@@ -0,0 +1,295 @@
...
+                    - set(unit, "")
+          exporters:
+            otlphttp:
+              endpoint: ${NEW_RELIC_OTLP_ENDPOINT}
+              headers:
+                api-key: ${NEW_RELIC_LICENSE_KEY}
+              compression: gzip
+          extensions:
+            health_check:
+              endpoint: 0.0.0.0:13133
```
pranav-new-relic, line 219: @ujwalsah just curious, I could certainly be wrong :) is this actually meant to be `0.0.0.0` and not `localhost` instead 🤔

recipes/newrelic/infrastructure/nrdot/rabbitmq/debian.yml
```diff
@@ -0,0 +1,295 @@
...
+
+processMatch: []
+
+preInstall:
+  discoveryMode:
+    - targeted
+
+inputVars:
+  - name: NR_CLI_RABBITMQ_DEPLOYMENT_NAME
+    prompt: "RabbitMQ deployment name (used to identify this instance in New Relic): "
```
pranav-new-relic, line 36: @ujwalsah we might want to consider having a default name here, to prevent an empty value going into the OTel config, etc if this ends up not being defined

recipes/newrelic/infrastructure/nrdot/rabbitmq/debian.yml
```diff
@@ -0,0 +1,295 @@
...
+  version: "3"
+  silent: true
+
+  vars:
+    IS_SYSTEMCTL:
+      sh: command -v systemctl | wc -l
+    NRDOT_ARCH:
+      sh: if [ "$(uname -m)" = "aarch64" ]; then echo "arm64"; else echo "amd64"; fi
+    NRDOT_VERSION:
+      sh: curl -sf "https://api.github.com/repos/newrelic/nrdot-collector-releases/releases/latest" | grep tag_name | cut -d '"' -f 4 | sed 's/^v//' || echo "1.11.1"
```
pranav-new-relic: @ujwalsah I had Claude review this for tinier nitpicks, and it seems to have found that the latest version is 1.13.0 - has this been set to 1.11.1 on purpose (e.g. to fallback to a stabler, known version)? If not, let's correct this (if appropriate to do so) :)

pranav-new-relic:
@ujwalsah I think it might be helpful to add an AL2023 (or an AL, in general) test definition too, given we have a `rhel` recipe added too with `amazon` in the `installTargets`

recipes/newrelic/infrastructure/nrdot/rabbitmq/debian.yml
ujwalsah: 0.0.0.0:13133 is correct — this is the standard OTel health check binding that listens on all interface.

recipes/newrelic/infrastructure/nrdot/rabbitmq/debian.yml
ujwalsah: I agree, added the default name.

recipes/newrelic/infrastructure/nrdot/rabbitmq/debian.yml
ujwalsah: Good catch @pranav-new-relic. It was set to 1.11.1 on purpose for testing before but I missed revert before pushing the changes. Updated it to latest version.

pranav-new-relic (approved):
LGTM

PR #1357, opened 2026-04-17
feat(es-otel): Add es otel nrdot recipies and tests

vagrawal-newrelic:
Below Changes has been added as part of this PR:
Added elasticsearch NRDOT OTEL instrumentation recipe for Debian and rhel OS.
Added tests for respective OS to test NRDOT installation

📓 Tests are successful.

### DIFF
--- a/recipes/newrelic/infrastructure/nrdot/elasticsearch-otel/debian.yml
+++ b/recipes/newrelic/infrastructure/nrdot/elasticsearch-otel/debian.yml
@@ -108,6 +108,9 @@ install:
             echo "NEW_RELIC_LICENSE_KEY={{.NEW_RELIC_LICENSE_KEY}}" | \
               tee -a {{.CONFIG_DIR}}/{{.COLLECTOR_DISTRO}}.conf > /dev/null
 
+          # 600: only root should read the license key.
+          chmod 600 {{.CONFIG_DIR}}/{{.COLLECTOR_DISTRO}}.conf
+
     create_collector_config:
       cmds:
         - |
@@ -689,6 +692,9 @@ install:
           fi
           printf "NEW_RELIC_OTLP_ENDPOINT=%s\n" "$OTLP_ENDPOINT" >> "$ES_ENV_FILE"
           printf "NEW_RELIC_LICENSE_KEY={{.NEW_RELIC_LICENSE_KEY}}\n" >> "$ES_ENV_FILE"
+
+          # 600: only root should read sensitive credentials (license key, ES password).
+          chmod 600 "$ES_ENV_FILE"
 
           echo "Collector config written to $COLLECTOR_CONFIG"
 
--- a/recipes/newrelic/infrastructure/nrdot/elasticsearch-otel/debian.yml
+++ b/recipes/newrelic/infrastructure/nrdot/elasticsearch-otel/debian.yml
@@ -1,0 +1,748 @@
+# Visit our schema definition for additional information on this file format
+# https://github.com/newrelic/open-install-library/blob/main/docs/recipe-spec/recipe-spec.md#schema-definition
+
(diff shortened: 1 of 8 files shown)


shashank-reddy-nr (dismissed):
LGTM

recipes/newrelic/infrastructure/nrdot/elasticsearch-otel/debian.yml
```diff
@@ -0,0 +1,802 @@
...
+      cmds:
+        - |
+          echo "Installing NRDOT Collector v{{.NRDOT_VERSION}}..."
+          NRDOT_DEB="/tmp/{{.COLLECTOR_DISTRO}}_{{.NRDOT_VERSION}}_linux_{{.NRDOT_ARCH}}.deb"
+          curl "https://github.com/newrelic/nrdot-collector-releases/releases/download/{{.NRDOT_VERSION}}/{{.COLLECTOR_DISTRO}}_{{.NRDOT_VERSION}}_linux_{{.NRDOT_ARCH}}.deb" \
+            --location --output "$NRDOT_DEB" 2>/dev/null
+          dpkg -i "$NRDOT_DEB" > /dev/null
+          if [ $? -ne 0 ]; then
+            dpkg --configure -a > /dev/null
+            apt-get -o DPkg::Lock::Timeout=60 install -f -y > /dev/null
```
Nandu-pns: Hi @vagrawal-newrelic  - we've seen scenarios where `dpkg --configure` / `--reconfigure` fails for customers due to various reasons. The current fallback in here has no exit check, which can silently continue into the configure step and lead to confusing downstream errors rather than a clear install failure. Can you handle this the same way it's done for RHEL [here](https://github.com/newrelic/open-install-library/pull/1357/changes#diff-f4069947312c20bdb3fc1e8e13fa91e1978a8f48d5db1d13041f334327c8081fR119-R121) — by explicitly checking the exit code after the fallback and failing fast with a clear error message?

test/deploy/linux/elasticsearch-otel/install/debian/roles/configure/tasks/main.yml
```diff
@@ -0,0 +1,31 @@
+---
+- debug:
+    msg: Install ElasticSearch 7.16+ for NRDOT Testing (Basic - No Auth)
```
Nandu-pns, line 3: I see all the four test definitions provision Elasticsearch 7.x which reached EOL on 15 April 2025 as seen [here](https://www.elastic.co/support/eol) . I believe it's better to update this to install Elasticsearch 8.x or to any latest version as the upstream repo may stop receiving security updates and running tests on a EOL version wouldn't be of much help to us. WDYT @vagrawal-newrelic ?

recipes/newrelic/infrastructure/nrdot/elasticsearch-otel/debian.yml
vagrawal-newrelic: @Nandu-pns  Thanks for pointing this out updated with the required exception handling. Thanks

test/deploy/linux/elasticsearch-otel/install/debian/roles/configure/tasks/main.yml
vagrawal-newrelic: @Nandu-pns Use the same as we are using the same for native agent, Due to some other priority items this cannot be addressed right now as we need to completely update the anisble script to incorporate the security standards that got changed in ES 8.x So created a ticket [here](https://new-relic.atlassian.net/browse/NR-561319) and will take this up later

Nandu-pns (dismissed):
Changes looks good to me !

Nandu-pns (approved):
LGTM.

PR #1371, opened 2026-05-22
feat(redis-otel): Add redis otel recipie for nrdot

vagrawal-newrelic:
Add NRDOT Redis OpenTelemetry recipe with test definitions for automated CI       
  validation.                                               
                                                                                    
  Recipe (recipes/newrelic/infrastructure/nrdot/redis/):                            
  - debian.yml — Debian/Ubuntu install via dpkg
  - rhel.yml — RHEL/CentOS/Amazon Linux install via rpm

  Supports configurable endpoint, identifier pattern (composite vs
  redis.instance.id), config preset (comprehensive vs optimized), optional log
  collection, host metrics, and custom attributes.

  Tests (test/definitions/nrdot/redis/):
  - Ubuntu 22.04, Ubuntu 24.04, Amazon Linux 2, Amazon Linux 2023 (all amd64)
  - Deploy scripts install Redis via package manager, then run the recipe with
  newrelic-cli

  Test plan

  - npm run check (schema validator) passes for both recipe files
  - Run test definitions against staging environment
  - Verify validationNrql detects redis metrics with instrumentation.provider =
  'opentelemetry'


recipes/newrelic/infrastructure/nrdot/redis/debian.yml
```diff
@@ -0,0 +1,757 @@
...
+            echo "OTELCOL_OPTIONS=\"${OTELCOL_OPTIONS}\"" | tee -a "$CONFIG_FILE" > /dev/null
+          fi
+
+          echo "OTELCOL_OPTIONS updated in ${CONFIG_FILE}: ${OTELCOL_OPTIONS}"
+
+    restart_nrdot:
+      cmds:
+        - |
+          systemctl daemon-reload
+          systemctl reload-or-restart {{.COLLECTOR_DISTRO}}.service
```
pranav-new-relic: **Commented by Claude**  |  `[P0 — blocker]`

@vagrawal-newrelic Same bug we just fixed on the IBM-MQ PR (see #1386 commit `048c6037`). `reload-or-restart` isn't quite what you want here.

If the collector is already running and its systemd unit defines `ExecReload=`, this command sends `SIGHUP` instead of doing a full restart. `SIGHUP` makes the process re-read its config file, but it does **NOT** cause systemd to re-evaluate the `OTELCOL_OPTIONS` change you just wrote in `configure_service`. So the collector stays on its old command line — silent failure, customer thinks it worked.

Fix: `systemctl enable` + `systemctl restart` (same pattern haproxy uses + the ibm-mq fix). Same issue in `rhel.yml`.

recipes/newrelic/infrastructure/nrdot/redis/debian.yml
```diff
@@ -0,0 +1,757 @@
...
+          REDIS_OTEL_CONFIG="/etc/nrdot-collector/redis-collector-config.yaml"
+          CONFIG_FILE="/etc/nrdot-collector/nrdot-collector.conf"
+
+          EXISTING_OPTIONS=$(grep "^OTELCOL_OPTIONS=" "$CONFIG_FILE" 2>/dev/null | sed 's/^OTELCOL_OPTIONS=//' | tr -d '"')
+          if [ "{{.COLLECTOR_WAS_RUNNING}}" = "1" ]; then
+            if echo "$EXISTING_OPTIONS" | grep -q "redis-collector-config.yaml"; then
+              OTELCOL_OPTIONS="$EXISTING_OPTIONS"
+            else
+              echo "Overriding default NRDOT config with Redis collector config."
+              OTELCOL_OPTIONS="--config=${REDIS_OTEL_CONFIG}"
```
pranav-new-relic, line 727: **Commented by Claude**  |  `[P0 — blocker]`

@vagrawal-newrelic This is where composability breaks. If the collector was already running with, say, `haproxy-config.yaml` in `OTELCOL_OPTIONS`, this line wipes it out and leaves only the redis config. Any prior nrdot integration on the host goes silent.

HAProxys recipe solved this by *appending*, not overwriting:
```bash
OTELCOL_OPTIONS="${EXISTING_OPTIONS} --config=${REDIS_OTEL_CONFIG}"
```

The whole reason to use `OTELCOL_OPTIONS` over a systemd drop-in is composability with other integrations. Overwriting defeats that. Your `postInstall.info` tells users to fix it manually — thats the tell. Same issue in `rhel.yml`.

recipes/newrelic/infrastructure/nrdot/redis/debian.yml
```diff
@@ -0,0 +1,757 @@
...
+
+    assert_nrdot_status_ok:
+      cmds:
+        - |
+          MAX_RETRIES=30
+          TRIES=0
+          echo "Waiting for NRDOT Collector to start..."
+          while [ $TRIES -lt $MAX_RETRIES ]; do
+            TRIES=$((TRIES + 1))
+            if systemctl is-active --quiet {{.COLLECTOR_DISTRO}}.service; then
```
pranav-new-relic: **Commented by Claude**  |  `[P0 — blocker]`

@vagrawal-newrelic `systemctl is-active` flips true the moment systemd starts the process. If the collector then panics on a bad config a second later, this loop may have already exited and reported success. The recipe would look green while the collector is actually dead.

You already declare a `health_check` extension on `:13133` — verify it too:
```bash
if systemctl is-active --quiet ... && curl -sf http://localhost:13133 > /dev/null 2>&1; then
```

Same fix Shashank landed on IBM-MQ. Same issue in `rhel.yml`.

recipes/newrelic/infrastructure/nrdot/redis/debian.yml
```diff
@@ -0,0 +1,757 @@
...
+
+          # Metrics configuration
+          if [ "$CONFIG_PRESET" = "comprehensive" ]; then
+            cat >> "$COLLECTOR_CONFIG" << 'EOF'
+              metrics:
+                redis.maxmemory:
+                  enabled: true
+                redis.role:
+                  enabled: true
+                redis.cmd.calls:
```
pranav-new-relic, line 288: **Commented by Claude**  |  `[P0 — customer $$ risk]`

@vagrawal-newrelic `redis.cmd.calls` and `redis.cmd.usec` emit one datapoint per unique Redis command name. Redis has hundreds of built-in commands, plus custom Lua scripts, so this can easily be tens of thousands of unique time series per host.

The `optimized` preset has a `filter/cardinality` processor that whitelists ~10 common commands (get/set/del/etc.). The `comprehensive` preset — which is the **default** — has no such filter. So the out-of-the-box install ships unbounded cardinality straight to NR, which the customer pays for by ingest volume.

Two options:
1. Apply the same `filter/cardinality` in comprehensive too.
2. Enable these metrics only in optimized (opt-in) and keep comprehensive to well-known bounded metrics.

Either way — the default shouldnt be a footgun. Same issue in `rhel.yml`.

recipes/newrelic/infrastructure/nrdot/redis/debian.yml
```diff
@@ -0,0 +1,757 @@
...
+              send_batch_max_size: 4096
+              timeout: 10s
+          EOF
+          fi
+
+          # Exporters section
+          cat >> "$COLLECTOR_CONFIG" << 'EOF'
+
+          exporters:
+            otlp_http:
```
pranav-new-relic, line 635: **Commented by Claude**  |  `[P1]`

@vagrawal-newrelic Every other nrdot recipe in this repo uses the canonical OTel Contrib name `otlphttp` (no underscore). Youre using `otlp_http`.

Tests pass, so `otlp_http` is apparently accepted as an alias in the current nrdot-collector — but (a) its inconsistent with haproxy / elasticsearch-otel / nginx, (b) aliases like this can be removed in future collector versions with zero warning, and (c) anyone reading your recipe alongside the others has to mentally translate.

Same story with `file_log/redis` on line 316 — the canonical name is `filelog`. That one isnt exercised by tests (nothing sets `NR_CLI_REDIS_LOG_PATH`), so we dont even know if it currently works. Use the canonical names. Same issue in `rhel.yml`.

recipes/newrelic/infrastructure/nrdot/redis/debian.yml
```diff
@@ -0,0 +1,757 @@
...
+  version: "3"
+  silent: true
+
+  vars:
+    IS_SYSTEMCTL:
+      sh: command -v systemctl | wc -l
+    NRDOT_ARCH:
+      sh: if [ "$(uname -m)" = "aarch64" ]; then echo "arm64"; else echo "amd64"; fi
+    NRDOT_VERSION:
+      sh: curl -sf "https://api.github.com/repos/newrelic/nrdot-collector-releases/releases/latest" | grep tag_name | cut -d '"' -f 4 | sed 's/^v//' || echo "1.11.1"
```
pranav-new-relic, line 49: **Commented by Claude**  |  `[P1]`

@vagrawal-newrelic The `|| echo "1.11.1"` doesnt do what it looks like.

Shell pipelines return the exit code of the **last** command (`sed`). When `curl -sf` fails on a network hiccup, `grep`/`cut`/`sed` all run on empty input and `sed` exits 0. The whole pipeline is "successful," `|| echo "1.11.1"` never fires, and `NRDOT_VERSION` becomes an empty string.

Youre saved by the `if [ -z "$COLLECTOR_VERSION" ]` check at line 110 — so this isnt a real bug **today**. But the fallback is misleading dead code, and if anyone ever adds `set -o pipefail` to Tasks shell, the hardcoded stale `1.11.1` would start firing silently. Simplest fix: just remove `|| echo "1.11.1"` and rely on the empty-string check.

Same pattern exists in kafka/rabbitmq/redis — worth a follow-up sweep once this lands.

recipes/newrelic/infrastructure/nrdot/redis/debian.yml
```diff
@@ -0,0 +1,757 @@
...
+            redis:
+              endpoint: "${REDIS_ENDPOINT}"
+              collection_interval: 10s
+          EOF
+
+          # Add authentication if provided
+          if [ -n "$REDIS_USERNAME" ] && [ -n "$REDIS_PASSWORD" ]; then
+            cat >> "$COLLECTOR_CONFIG" << EOF
+              username: "${REDIS_USERNAME}"
+              password: "${REDIS_PASSWORD}"
```
pranav-new-relic: **Commented by Claude**  |  `[P1 — security]`

@vagrawal-newrelic Youre writing the Redis password to `/etc/nrdot-collector/redis-collector-config.yaml` in plaintext, but this file never gets `chmod 600`. Meanwhile the license-key config file (line 187) IS chmod 600.

Any non-root user on the box can currently `cat` this file and read the Redis password. Two ways to fix:
- **Simple:** `chmod 600` this file too after writing it.
- **Better:** put the password in the env file (like the license key) and reference it as `password: ${env:REDIS_PASSWORD}` in the collector config. That keeps secrets out of the config file entirely.

Same issue in `rhel.yml`.

recipes/newrelic/infrastructure/nrdot/redis/debian.yml
vagrawal-newrelic: Fixed — changed to systemctl enable + systemctl restart.

recipes/newrelic/infrastructure/nrdot/redis/debian.yml
vagrawal-newrelic: By design. Multi-config composition (--config=a.yaml --config=b.yaml) causes conflicts when both configs define same processors with different settings. Override is deliberate; postInstall informs user how to restore composability if they handle conflicts manually.

recipes/newrelic/infrastructure/nrdot/redis/debian.yml
vagrawal-newrelic: Fixed — added curl -sf http://localhost:13133 to the assert loop alongside systemd check.

recipes/newrelic/infrastructure/nrdot/redis/debian.yml
vagrawal-newrelic: By design. User explicitly selects comprehensive vs optimized in the install flow. Comprehensive = full visibility (all metrics). Optimized = cost-efficient with cardinality filtering. Same pattern as Elasticsearch/Kafka. The UI clearly states: Provides maximum visibility with all available telemetry data.

recipes/newrelic/infrastructure/nrdot/redis/debian.yml
vagrawal-newrelic: Intentional. NRDot 1.19.0 emits deprecation warnings for old names (otlphttp, resourcedetection, filelog, hostmetrics). New canonical names are otlp_http, resource_detection, file_log, host_metrics. Tested and confirmed working. Other recipes in this repo should be updated too.

recipes/newrelic/infrastructure/nrdot/redis/debian.yml
vagrawal-newrelic: Fair point — the fallback is unreachable due to shell pipeline behavior. The empty-string guard at line 110 handles the failure case properly. Harmless but acknowledged as misleading. Low priority.

recipes/newrelic/infrastructure/nrdot/redis/debian.yml
vagrawal-newrelic: Fixed — added chmod 600 on the collector config file after writing it.

pranav-new-relic: **Commented by Claude** — follow-up on the fix

@vagrawal-newrelic Your `chmod 600` fixed the security concern but broke the collector — thats why the tests just went red. Ive pushed a follow-up commit removing the chmod so CI unblocks, but the underlying issue you were trying to address is still open. Quick explanation of why the two similar chmods behave differently:

**License-key file (`nrdot-collector.conf`, line 187) — `chmod 600` works fine.**
That file is listed as `EnvironmentFile=` in the systemd unit. **systemd** (running as root) reads it, extracts `KEY=VALUE` pairs, and injects them into the collector processs environment before exec. The collector process itself never opens the file — so its own UID doesnt matter for reading permissions.

**Collector config yaml (`redis-collector-config.yaml`) — `chmod 600` breaks it.**
This one is passed as `--config=/etc/nrdot-collector/redis-collector-config.yaml` on the command line. The **collector binary itself** opens and parses it at runtime. That binary runs as a non-root service user (per the debian packages systemd unit). A `root:root` mode-600 file locks that user out, so the collector fails with `permission denied`, systemd restarts it 5×, gives up, unit is `failed`, recipe exits 31.

HAProxys recipe demonstrates the pattern: it `chmod 600`s **only** its `.conf` env file (line 192), never its `haproxy-config.yaml`. Same for nginx and elasticsearch-otel.

**Proper fix** for the security concern (which I noted as the "better" option in the original comment): move `REDIS_USERNAME` and `REDIS_PASSWORD` into the env file in `configure_env`, then reference them in the yaml as `${env:REDIS_USERNAME}` and `${env:REDIS_PASSWORD}` — same pattern the license key already uses. That keeps the secrets out of the yaml file entirely, so the yaml can safely stay world-readable.

Sketch:
```bash
# in configure_env, after the license-key block:
if [ -n "${NR_CLI_REDIS_PASSWORD:-}" ]; then
  if grep -q "^REDIS_PASSWORD=" "$CONFIG_FILE" 2>/dev/null; then
    sed -i "s|^REDIS_PASSWORD=.*|REDIS_PASSWORD=${NR_CLI_REDIS_PASSWORD}|" "$CONFIG_FILE"
  else
    echo "REDIS_PASSWORD=${NR_CLI_REDIS_PASSWORD}" >> "$CONFIG_FILE"
  fi
fi
# ... same for REDIS_USERNAME ...
```

And in `create_collector_config`:
```yaml
receivers:
  redis:
    endpoint: "${REDIS_ENDPOINT}"
    password: ${env:REDIS_PASSWORD}   # not "$REDIS_PASSWORD"
    username: ${env:REDIS_USERNAME}
```

Happy to push this myself if youd like, but I figured youd want to own the actual fix. My revert commit unblocks CI in the meantime.

vagrawal-newrelic: @pranav-new-relic  pushed the fix!!

recipes/newrelic/infrastructure/nrdot/redis/debian.yml
```diff
@@ -0,0 +1,776 @@
...
+
+          # Use sed to update in-place if the key already exists, otherwise append.
+          # This makes configure_env idempotent across recipe re-runs and upgrades.
+          if grep -q "^NEW_RELIC_LICENSE_KEY=" "$CONFIG_FILE" 2>/dev/null; then
+            sed -i "s|^NEW_RELIC_LICENSE_KEY=.*|NEW_RELIC_LICENSE_KEY={{.NEW_RELIC_LICENSE_KEY}}|" "$CONFIG_FILE"
+          else
+            echo "NEW_RELIC_LICENSE_KEY={{.NEW_RELIC_LICENSE_KEY}}" | tee -a "$CONFIG_FILE" > /dev/null
+          fi
+
+          if [ "{{.NEW_RELIC_REGION}}" = "EU" ]; then
```
Nandu-pns, line 172: @vagrawal-newrelic, Could you also add jp specific OTLP endpoint ( just similar to how it's done for other recipes, for example [here](https://github.com/newrelic/open-install-library/blob/main/recipes/newrelic/infrastructure/nrdot/ibm-mq/rhel.yml#L235) ).

vagrawal-newrelic: Added JP endpoint support. Note: our recipe uses `otlp_http` (HTTP exporter) so no port is needed — HTTPS defaults to 443. The IBM-MQ example uses `otlp` (gRPC) which requires `:4317`.

pranav-new-relic (approved):

PR #1373, opened 2026-05-28
feat: add ubuntu 26 support for ac and bump version to 1.16.1

vjripoll:
Add ubuntu 26 for AC debian recipe and test for it + bump ac version to 1.16.1


gsanchezgavier (approved):
LGTM

PR #1384, opened 2026-06-24
test: Add Windows Server 2022 and 2025 test definitions for the .NET agent

nr-ahemsath:
Add Windows Server 2022 and 2025 test definitions for IIS + the .NET APM agent (full and core-only variants) to both the EU and JP regions, mirroring the existing 2019 set. Uses the Windows_Server-2022/2025-English-Full-Base-* AMIs already referenced by the agent-control definitions.

I also updated the version of the ASP.NET Core hosting module being installed in `test/deploy/windows/iis/deploy-webapplication/roles/configure/tasks/main.yml` to be .NET 10; the old link had expired and so both existing and new tests were failing.

There is also a new `ansible.builtin.wait_for_connection` step added to `test/deploy/windows/iis/install/roles/prepare/tasks/main.yml` in order to fix some test flakiness observed.

Finally: in order to get all of these tests to pass in their current form, I had to run each one individually - when they run simultaneously, they fail due to an AWS security group collision issue.  I was able to temporarily work around that problem (see the full commit history) by making unique resource IDs for the main server in each test.  Claude Code's analysis indicates that the issue is that the AWS security group is based on the filename + resource ID, but we have duplicate filenames which are bit-for-bit identical for the EU and JP versions of each test.  I don't think fixing this problem is in the scope of this PR, and I didn't want to create an anti-pattern by leaving the unique resource IDs in place to work around this. (Update: I have reinstated the region-specific resource IDs at the request of @pranav-new-relic to ease the process of merging this PR.)


pranav-new-relic:
image

pranav-new-relic (approved):
LGTM. Resource-ID rename fixed the AWS security-group collision (16/16 test deploys green on the previous head), and the design doc is now in a properly named location (`docs/test-framework/dotnet-linux-test-modernization.md`) that matches the existing conventions there. Squash-merging.

PR #1386, opened 2026-06-26
feat: Add NRDOT IBM MQ install recipe

shashank-reddy-nr:
## Summary

Adds a `nrdot-collector-ibm-mq` install recipe that configures the NRDOT Collector to monitor IBM MQ via OpenTelemetry.

- `recipes/newrelic/infrastructure/nrdot/ibm-mq/{debian,rhel}.yml` — **config-only** recipe: installs the NRDOT Collector and points it at an **existing** IBM MQ `mq-metric-samples` (`mq_prometheus`) exporter, exporting `ibmmq_*` metrics to New Relic over **OTLP/gRPC** (`otlp` exporter, `https://otlp.nr-data.net:4317` and the EU/staging/JP equivalents). It does not install MQ or the exporter.
- `test/definitions/nrdot/ibm-mq/*.json` — amd64 test definitions (debian, ubuntu22, rhel9, al2023).
- `test/deploy/linux/nrdot-ibm-mq/...` — Ansible roles that stand up IBM MQ + the exporter (`MQ_ENABLE_METRICS`, `:9157`) via Docker for the test harness.

## How it works

- **Identity** — resource detection (`env, ec2, gcp, azure, system`) sets `host.id`, which `transform/ibmmq-cleanup` copies to `target.name` (the IBM MQ entity GUID prefix). `cluster.name` comes from the optional `NR_CLI_IBMMQ_CLUSTER_NAME` input. The raw `qmgr`/`queue` labels and `ibmmq_*` metric names are preserved so `IBMMQ_MANAGER` / `IBMMQ_QUEUE` entities synthesize.
- **Config via env expansion** — the collector config is written verbatim and reads `${env:...}` values from a `0600` env file (`ibmmq-env.conf`, holding the license key, OTLP endpoint, exporter endpoint, cluster name). A systemd drop-in wires `EnvironmentFile=` + `--config`. No in-place `sed` edits, so the license key never lands in the config file. Queue filters use `[.]` literal-dot character classes (heredoc-safe) rather than backslash escapes.
- **Pipeline** — `memory_limiter` first (OTel best practice), then `resourcedetection` → `filter` (drop exporter self-metrics + high-cardinality `SYSTEM.*`/`AMQ.*` queues) → `transform` cleanup → `cumulativetodelta` (Prometheus counters → delta) → `batch`.
- **Idempotent + verified start** — the install step skips the download when the latest version is already present; `assert_nrdot_status_ok` waits on both `systemctl is-active` **and** the health-check endpoint (`:13133`) before reporting success, so a crash-looping collector can't pass.

## Testing

Validated end-to-end on **all four host types** against the New Relic staging account, each with a distinct `NR_CLI_IBMMQ_CLUSTER_NAME`:

| Distro | Recipe |
|---|---|
| Debian 13 | `debian.yml` |
| Ubuntu 22 | `debian.yml` |
| Amazon Linux 2023 | `rhel.yml` |
| Amazon Linux 2 | `rhel.yml` |

On every host `newrelic install -n nrdot-collector-ibm-mq` installs, validates, and `ibmmq_*` metrics arrive with `instrumentation.provider=opentelemetry`, `qmgr=QM1`, the configured `cluster.name`, and `target.name` = the EC2 instance id (`host.id`). Re-running the recipe is a no-op (idempotency confirmed).


recipes/newrelic/infrastructure/nrdot/ibm-mq/debian.yml
```diff
@@ -0,0 +1,269 @@
...
+  version: "3"
+  silent: true
+
+  vars:
+    IS_SYSTEMCTL:
+      sh: command -v systemctl | wc -l
+    NRDOT_ARCH:
+      sh: if [ "$(uname -m)" = "aarch64" ]; then echo "arm64"; else echo "amd64"; fi
+    NRDOT_VERSION:
+      sh: curl -sf "https://api.github.com/repos/newrelic/nrdot-collector-releases/releases/latest" | grep tag_name | cut -d '"' -f 4 | sed 's/^v//' || echo "1.18.0"
```
pranav-new-relic: **Commented by Claude**

@shashank-reddy-nr Subtle shell bug worth fixing here. The `||` in a pipeline only checks the exit code of the **last** command, which is `sed`. So when `curl -sf` fails, `grep`/`cut`/`sed` still run on empty input — and `sed` exits 0 on empty input. The pipeline as a whole "succeeds," `|| echo "1.18.0"` never fires, and `NRDOT_VERSION` ends up as an empty string. The download URL then becomes `..._linux_amd64.deb` (missing version) and the recipe fails several steps later at `dpkg` with a misleading "Failed to install" error.

Second-order trap: if the fallback ever *did* fire, hosts already on `1.18.0` would silently pass the idempotency check and freeze on that version even when newer releases exist.

Hint: capture into `$()`, test for empty, fail fast — and drop the hardcoded fallback. Same pattern is in `rhel.yml` (and also in kafka/rabbitmq recipes, worth a follow-up sweep).

pranav-new-relic (approved):
LGTM.

PR #1390, opened 2026-07-07
fix(mssql): self-hosted pre-install check now supports SQL Auth credentials and fixes TCP timeout

pkudikyala:
**Problem**

The self-hosted MSSQL `requireAtDiscovery` pre-install check and install task connection checks used Windows Authentication exclusively, ignoring `NEW_RELIC_MSSQL_SQL_USERNAME and NEW_RELIC_MSSQL_SQL_PASSWORD` environment variables that users may provide.

This caused the installer to fail with exit status 3 ("unsupported") for any user whose Windows identity has no SQL Server login — even when valid SQL credentials were supplied. The Azure SQL path already supported explicit SQL credentials via `NEW_RELIC_MSSQL_SQL_USERNAME`, but the self-hosted path did not.

A second issue was discovered during investigation: all self-hosted sqlcmd connection checks used -l 1 (1-second login timeout), which is too short for TCP connections and causes reliable timeouts when a port number is appended to the connection string.

**Root Cause**

In the else branch (self-hosted MSSQL) of both requireAtDiscovery and the install task, sqlcmd was called without -U/-P flags:

 Before — always Windows Auth, always times out on TCP
`$name=(sqlcmd -S $connection -Q "SELECT @@SERVICENAME" -l 1 2>&1)`

The Azure SQL branch already read `$env:NEW_RELIC_MSSQL_SQL_USERNAME and passed -U/-P to sqlcmd`. The self-hosted branch lacked this same modularity.

**Fix**

1. preInstall.requireAtDiscovery (self-hosted branch):
- Added `$sqlUsername / $sqlPassword reading from NEW_RELIC_MSSQL_SQL_USERNAME / NEW_RELIC_MSSQL_SQL_PASSWORD env vars (with NR_CLI_SQL_USERNAME / NR_CLI_SQL_PASSWORD as fallbacks)`
- Added `$sqlCredArgs` array splatting — injects -`U/-P` when credentials are set, empty array (Windows Auth) when not
- Changed -l 1 → -l 30 on all 3 connection attempts

2. Install task (self-hosted branch connection checks):
- Added `$installSqlCredArgs using $NEW_RELIC_MSSQL_SQL_USERNAME` (already read earlier in the task)
- Changed -l 1 → -l 30 on all 3 connection attempts

**Test screenshots:**

**Before fix:**
The integration is failing with the username and password.
mssql-selfhosted-error

**After fix:**
After fix using the username and password, it's integrated successfully.
mssql-self-hosted-fix

### DIFF
--- a/recipes/newrelic/infrastructure/ohi/sql/ms-sql.yml
+++ b/recipes/newrelic/infrastructure/ohi/sql/ms-sql.yml
@@ -106,14 +106,21 @@ preInstall:
 
         $NEW_RELIC_MSSQL_DB_HOSTNAME = if ($env:NEW_RELIC_MSSQL_DB_HOSTNAME) { $env:NEW_RELIC_MSSQL_DB_HOSTNAME } else { $env:NR_CLI_DB_HOSTNAME };
         if ([string]::IsNullOrWhiteSpace($NEW_RELIC_MSSQL_DB_HOSTNAME)) {$NEW_RELIC_MSSQL_DB_HOSTNAME = hostname}
-        
+
         $NEW_RELIC_MSSQL_DB_PORT = if ($env:NEW_RELIC_MSSQL_DB_PORT) { $env:NEW_RELIC_MSSQL_DB_PORT } else { $env:NR_CLI_DB_PORT };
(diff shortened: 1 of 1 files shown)


pranav-new-relic (approved):
Approving this, since this is a pointed fix and has been tested (thank you for promptly testing it, @pkudikyala)

PR #1403, opened 2026-07-20
feat(nrdot-mssql): Added the automation cli for mssql otel

RamanaReddy8801:
Jira Ticket: https://new-relic.atlassian.net/browse/NR-595317
Description: Implementation of the New Relic CLI installer for MSSQL OpenTelemetry (OTel) monitoring.

I’ve attached the test [[documentation](https://docs.google.com/document/d/1tnR4WMyzgmaN-X2YhbDIK-cTJwQI6MRMnvQxADe-yZo/edit?tab=t.0)] covering several key integration scenarios for your review.

### CI test coverage (Windows)

Added `test/definitions/nrdot/mssql-otel/mssql-windows2019.json` to test
`nrdot-collector-mssql` on Windows (SQL-auth mode) — this is the first
NRDOT recipe with Windows CI coverage in the repo.

Two supporting changes were needed to make this possible:

- **New role** `test/deploy/windows/nrdot-mssql/install/roles/configure`:
  the only SQL Server AMI available in CI ships with `sa` disabled and
  Windows-Auth-only mode. This role switches SQL Server to Mixed Mode,
  restarts it, then enables and sets a known `sa` password so the
  recipe's SQL-auth path (`NR_CLI_MSSQL_AUTH_MODE=1`) can be tested.
- **Modified** `test/deploy/windows/newrelic-cli/install-recipe/roles/onafterstart/tasks/main.yml`:
  added a generic `env_var` passthrough (PowerShell `$env:` assignments),
  matching the mechanism the Linux test runner already has. Without it,
  there was no way to inject the recipe's `NR_CLI_MSSQL_*` inputVars into
  a Windows test at all. This is additive (`when: env_var is defined`) and
  doesn't affect any existing Windows test definition.


recipes/newrelic/infrastructure/nrdot/mssql-otel/windows.yml
```diff
@@ -0,0 +1,1034 @@
...
+            $account = "{{.NR_CLI_MSSQL_WIN_ACCOUNT}}"
+            $password = "{{.NR_CLI_MSSQL_WIN_PASSWORD}}"
+          } else {
+            $account = "{{.NR_CLI_MSSQL_GMSA_ACCOUNT}}"
+            $password = ""
+          }
+
+          Stop-Service -Name $service -Force -ErrorAction SilentlyContinue
+
+          cmd /c "sc config `"$service`" obj= `"$account`" password= `"$password`""
```
Nandu-pns: Found one more spot with the same problem — in `configure_service_identity`:

```powershell
$account = "{{.NR_CLI_MSSQL_WIN_ACCOUNT}}"
$password = "{{.NR_CLI_MSSQL_WIN_PASSWORD}}"
...
cmd /c "sc config `"$service`" obj= `"$account`" password= `"$password`""
```

`WIN_PASSWORD` is marked `secret: true`, so it's a real password. It's put directly into the script text instead of read from env, and it also goes through `cmd /c`, which reads it again as a command line. Domain passwords often contain characters like `!@#$%^&*`, so this can break or be misused. `WIN_ACCOUNT` and `GMSA_ACCOUNT` have the same issue, lower risk since account names are more constrained.

Since `WIN_PASSWORD` is a real secret, could you take a look before merging?

recipes/newrelic/infrastructure/nrdot/mssql-otel/windows.yml
```diff
@@ -0,0 +1,1034 @@
...
+          $preset = "{{.NR_CLI_MSSQL_CONFIG_PRESET}}"
+          $server = if ($env:NR_CLI_MSSQL_SERVER) { $env:NR_CLI_MSSQL_SERVER } else { "localhost" }
+          $port = if ($env:NR_CLI_MSSQL_PORT) { $env:NR_CLI_MSSQL_PORT } else { "1433" }
+          $region = "{{.NEW_RELIC_REGION}}"
+          $licenseKey = "{{.NEW_RELIC_LICENSE_KEY}}"
+          $configDir = "C:\Program Files\nrdot-collector"
+          $configPath = "$configDir\mssql-config.yaml"
+          $pwFile = "$env:TEMP\nr-mssql-newrelic-pw.tmp"
+
+          New-Item -Path $configDir -ItemType Directory -Force | Out-Null
```
Nandu-pns, line 284: Same issue as the one I flagged on #1412, just on Windows this time — in `create_collector_config`:

```powershell
$configDir = "C:\Program Files\nrdot-collector"
$configPath = "$configDir\mssql-config.yaml"
...
New-Item -Path $configDir -ItemType Directory -Force | Out-Null
...
$configText | Set-Content -Path $configPath -Encoding utf8
```

`mssql-config.yaml` ends up with the SQL Server monitoring password in plaintext (mode 1) and the New Relic license key (always), but the file never gets its ACL restricted — it inherits whatever `C:\Program Files\...` normally allows, which typically means any local user can read it.

One thing to be careful of here: the fix needs to account for which account the `nrdot-collector` service actually runs as, since that's set earlier in `configure_service_identity` and it's different per auth mode — mode 1 leaves it as LocalSystem, but mode 2 switches it to `NR_CLI_MSSQL_WIN_ACCOUNT` (a domain account) and mode 3 to `NR_CLI_MSSQL_GMSA_ACCOUNT` (a gMSA) — neither of which is SYSTEM or a local Administrator. A lockdown that only grants `SYSTEM`/`Administrators` would leave the service unable to read its own config file in modes 2/3, and the collector would fail to start (`assert_nrdot_status_ok` would then time out with exit 31).

Something like this, granting whichever account the service will actually run as:
```powershell
$serviceAccount = if ($mode -eq "1") { $null }
                   elseif ($mode -eq "2") { $env:NR_CLI_MSSQL_WIN_ACCOUNT }
                   else { $env:NR_CLI_MSSQL_GMSA_ACCOUNT }

$grants = @("SYSTEM:(F)", "BUILTIN\Administrators:(F)")
if ($serviceAccount) { $grants += "${serviceAccount}:(R)" }

icacls $configPath /inheritance:r /grant:r $grants
```

recipes/newrelic/infrastructure/nrdot/mssql-otel/windows.yml
RamanaReddy8801: Fixed. `WIN_ACCOUNT/WIN_PASSWORD/GMSA_ACCOUNT` now read from `$env:` instead of being substituted into the script text, same as the **SA_PASSWORD** fix — applied everywhere they're used (`assert_auth_inputs, configure_database_user, configure_service_identity`), not just this spot.

Also swapped `cmd /c "sc config ...`" for a direct `& sc.exe config ... `call, so the password isn't re-parsed by cmd.exe's metacharacter rules (% triggers env-var expansion even inside quotes, `^/&/| `are shell operators) — that was the actual risk for domain passwords with special characters.

recipes/newrelic/infrastructure/nrdot/mssql-otel/windows.yml
```diff
@@ -0,0 +1,1047 @@
...
+            BEGIN CATCH
+              PRINT 'Error on ' + @name + ': ' + ERROR_MESSAGE();
+            END CATCH
+            FETCH NEXT FROM db_cursor INTO @name;
+          END
+          CLOSE db_cursor;
+          DEALLOCATE db_cursor;
+          GO
+          "@
+            $sql | Set-Content -Path $sqlFile
```
Nandu-pns: Same issue as #1412, on Windows — `$sqlFile` gets the plaintext monitoring password written to it but never gets an `icacls` restriction like `$configPath` does. And if `Invoke-GrantScript` fails, it exits before `Remove-Item -Path $sqlFile` runs, so the file is left behind with the password in it. `$pwFile` has the same gap — no `icacls` call for it anywhere.

Could `$sqlFile` and `$pwFile` get the same `icacls` treatment as `$configPath`, and get cleaned up even on failure (e.g. try/finally around the grant step)?

RamanaReddy8801: Fixed. `$sqlFile` and `$pwFile` now go through the same `Protect-TempFile` helper as `$configPath — icacls /inheritance:r /grant:r "SYSTEM:(F)" "BUILTIN\Administrators:(F)"`, called right after each is written

recipes/newrelic/infrastructure/nrdot/mssql-otel/windows.yml
```diff
@@ -0,0 +1,870 @@
...
+
+          $sql = "USE [master];`nGO`nIF NOT EXISTS (SELECT 1 FROM sys.server_principals WHERE name = '$loginName')`n  CREATE LOGIN [$loginName] WITH PASSWORD = '$NrPassword';`nELSE`n  ALTER LOGIN [$loginName] WITH PASSWORD = '$NrPassword';`nGO`nGRANT VIEW SERVER STATE TO [$loginName];`nGRANT VIEW ANY DEFINITION TO [$loginName];`nGRANT VIEW ANY DATABASE TO [$loginName];`nGO`nDECLARE @name SYSNAME;`nDECLARE db_cursor CURSOR READ_ONLY FORWARD_ONLY FOR`nSELECT [name]`nFROM [master].[sys].[databases]`nWHERE [name] NOT IN ('master', 'msdb', 'model', 'rdsadmin', 'distribution')`nAND [state] = 0;`nOPEN db_cursor;`nFETCH NEXT FROM db_cursor INTO @name;`nWHILE @@FETCH_STATUS = 0`nBEGIN`n  BEGIN TRY`n    EXEC('USE [' + @name + '];`n    IF NOT EXISTS (SELECT 1 FROM sys.database_principals WHERE name = ''$loginName'')`n    BEGIN`n      CREATE USER [$loginName] FOR LOGIN [$loginName];`n    END;`n    GRANT VIEW DATABASE STATE TO [$loginName];');`n  END TRY`n  BEGIN CATCH`n    PRINT 'Error on ' + @name + ': ' + ERROR_MESSAGE();`n  END CATCH`n  FETCH NEXT FROM db_cursor INTO @name;`nEND`nCLOSE db_cursor;`nDEALLOCATE db_cursor;`nGO"
```
Nandu-pns, line 254: `$loginName` comes straight from `NR_CLI_MSSQL_LOGIN_NAME` with nothing validating it before it's embedded in `[$loginName]` throughout this block (`CREATE LOGIN`, `ALTER LOGIN`, three `GRANT`s, `CREATE USER ... FOR LOGIN` twice, `GRANT VIEW DATABASE STATE`). T-SQL bracket-quoted identifiers escape a literal `]` by doubling it — a single `]` in the username breaks out of the identifier and whatever comes after runs as `sa`. Same class of issue is up on the Linux side in #1412. Could we validate `$loginName` (e.g. `$loginName -match '^[A-Za-z][A-Za-z0-9_]*$'`) before this block runs?

recipes/newrelic/infrastructure/nrdot/mssql-otel/windows-rds.yml
```diff
@@ -0,0 +1,913 @@
...
+          $loginName = '{{.NR_CLI_MSSQL_LOGIN_NAME}}'
+          $masterUser = '{{.NR_CLI_MSSQL_MASTER_USER}}'
+          $chars = (48..57) + (65..90) + (97..122)
+          $NrPassword = -join ($chars | Get-Random -Count 24 | ForEach-Object {[char]$_})
+
+          $sql = @"
+          USE [master];
+          GO
+          IF NOT EXISTS (SELECT 1 FROM sys.server_principals WHERE name = '$loginName')
+            CREATE LOGIN [$loginName] WITH PASSWORD = '$NrPassword';
```
Nandu-pns, line 264: Same as `windows.yml` — `$loginName` is unvalidated and reused unescaped in `[$loginName]` throughout this block. A `]` in the username breaks out of the bracket-quoted identifier and runs as the RDS master user. Same fix needed here as on `windows.yml` and the Linux recipes in #1412.

recipes/newrelic/infrastructure/nrdot/mssql-otel/windows-winauth.yml
```diff
@@ -0,0 +1,918 @@
...
+          function Protect-TempFile($path) {
+            icacls $path /inheritance:r /grant:r "SYSTEM:(F)" "BUILTIN\Administrators:(F)" | Out-Null
+            if ($LASTEXITCODE -ne 0) {
+              Write-Host -ForegroundColor Red "Failed to restrict permissions on $path."
+              exit 1
+            }
+          }
+
+          if ($mode -eq "1") {
+            $account = '{{.NR_CLI_MSSQL_WIN_ACCOUNT}}'
```
Nandu-pns, line 262: Both branches here take user input straight into bracket-quoted identifiers with no validation — `$account` (this line) and `$gmsa` (line 265) both land in `[$account]`/`[$gmsa]` unescaped, repeated across `CREATE LOGIN ... FROM WINDOWS`, three `GRANT`s, `CREATE USER ... FOR LOGIN`, etc. A stray `]` in either breaks out of the identifier and runs as `sa`-equivalent. Domain/gMSA account names are more constrained than free-text usernames in practice, but nothing in this script actually enforces that — worth validating before this runs, same as the other recipes in this feature.

recipes/newrelic/infrastructure/nrdot/mssql-otel/windows-rds-winauth.yml
```diff
@@ -0,0 +1,989 @@
...
+          function Protect-TempFile($path) {
+            icacls $path /inheritance:r /grant:r "SYSTEM:(F)" "BUILTIN\Administrators:(F)" | Out-Null
+            if ($LASTEXITCODE -ne 0) {
+              Write-Host -ForegroundColor Red "Failed to restrict permissions on $path."
+              exit 1
+            }
+          }
+
+          if ($mode -eq "1") {
+            $account = '{{.NR_CLI_MSSQL_WIN_ACCOUNT}}'
```
Nandu-pns, line 264: Same as `windows-winauth.yml` — `$account`/`$gmsa` (here and line 303) go straight into `[$account]`/`[$gmsa]` with nothing validating them first, repeated unescaped through the whole grant block. Same fix as the other three recipes in this feature.

recipes/newrelic/infrastructure/nrdot/mssql-otel/windows.yml
RamanaReddy8801: Added the validation in `assert_auth_inputs` for `windows.yml`, `$loginName -notmatch '^[A-Za-z][A-Za-z0-9_]*$'`, exits before `configure_database_user` runs

recipes/newrelic/infrastructure/nrdot/mssql-otel/windows-rds.yml
RamanaReddy8801: Added the validation in `assert_auth_inputs` for `windows-rds.yml`, `$loginName -notmatch '^[A-Za-z][A-Za-z0-9_]*$'`, exits before `configure_database_user` runs

recipes/newrelic/infrastructure/nrdot/mssql-otel/windows-winauth.yml
RamanaReddy8801: In `windows.yml/windows-rds.yml`, `$loginName` is a brand-new SQL login that this recipe creates — the customer is free to pick any name they want (default `newrelic`), so it's reasonable to restrict it to a safe character set.

In `windows-winauth.yml/windows-rds-winauth.yml`, `$winAccount/$gmsaAccount` are **pre-existing AD-provided identities** — the recipe doesn't create them, it just grants permissions to whatever already exists in the domain. The format is fixed by AD conventions, not customer choice:

Windows Auth: `DOMAIN\username`
gMSA: `DOMAIN\gMSAName$`

recipes/newrelic/infrastructure/nrdot/mssql-otel/windows-rds-winauth.yml
RamanaReddy8801: In `windows.yml/windows-rds.yml`, `$loginName` is a brand-new SQL login that this recipe creates — the customer is free to pick any name they want (default `newrelic`), so it's reasonable to restrict it to a safe character set.

In `windows-winauth.yml/windows-rds-winauth.yml`, `$winAccount/$gmsaAccount` are **pre-existing AD-provided identities** — the recipe doesn't create them, it just grants permissions to whatever already exists in the domain. The format is fixed by AD conventions, not customer choice:

Windows Auth: `DOMAIN\username`
gMSA: `DOMAIN\gMSAName$`

Nandu-pns (approved):
Changes LGTM  @RamanaReddy8801 — approving to unblock testing these recipes via `Ansible/Chef` tools as discussed. One open item: the winauth SQL injection I flagged above still stands. Since you all know this stuff better than I do, so I'll leave the right fix to you — just wanted to flag it again so it doesn't get lost, and to revisit whenever you get a chance.

PR #1404, opened 2026-07-21
[ebpf] Release 1.5.0 changes

burhan-nr:
- New SUSE/SLES recipe added for the eBPF agent
- Extended OS/platform support: RHEL 10, Oracle Linux (incl. UEK kernel headers), Ubuntu 26.04, Debian 13
- Replaced reportAiMetrics/genAICaptureMessageContent with a nested ai_monitoring config block
- Added httpPathNormalizationPatterns and httpBodyLimitBytes config options
- Relaxed Ubuntu install failure check — only fails if the eBPF agent package itself didn't install
- Renamed and re-defaulted eBPF agent reporting variables in the Kubernetes recipe
- Commented out hard requirement for `DEPLOYMENT_NAME` during installation

### DIFF
--- a/recipes/newrelic/ebpf/awslinux.yml
+++ b/recipes/newrelic/ebpf/awslinux.yml
@@ -344,7 +344,7 @@ install:
             # 'auto' (send only when neither APM nor OTel agent is attached)
             enabled: ${AI_MONITORING_ENABLED:-false}
             # samplingLatency represents the sampling latency threshold for the spans to export.
-            # Options: p1, p10, p50, p90, p99.
+            # Options: any percentile from p0 to p99 (e.g., p0, p1, p5, p10, p50, p90, p99).
             samplingLatency: "${AI_MONITORING_SAMPLING_LATENCY:-p50}"
             # -- samplingErrorRate represents the error rate threshold
             # for an HTTP route where surpassing it would mean the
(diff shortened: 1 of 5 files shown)


recipes/newrelic/ebpf/awslinux.yml
```diff
@@ -206,13 +206,13 @@ install:
...
-            exit 1
-          fi
-          echo "Deployment name validated."
+          # if [[ -z "${DEPLOYMENT_NAME:-}" && "${NEW_RELIC_AGENT_CONTROL:-}" != "true" ]]; then
+          #   echo ""
+          #   echo -e "\e[31mERROR:\e[0m DEPLOYMENT_NAME is required but not set. Aborting installation."
+          #   echo ""
+          #   exit 1
+          # fi
+          # echo "Deployment name validated."
```
Nandu-pns, line 215: I see this isn't mentioned in the PR description. Previously, an install without `DEPLOYMENT_NAME` (outside Agent Control mode) would hard-fail; now it silently proceeds with an empty deploymentName in the generated config. Is this intentional, or a leftover local debug comment-out that should be uncommented before merge?

burhan-nr: This is an intentional change. The hard requirement for `DEPLOYMENT_NAME` has been removed. I'll update the comment.

Nandu-pns (approved):

PR #1412, opened 2026-07-28
feat(nrdot-mssql): add debian.yml/rhel.yml recipes for Linux SQL Server monitoring

RamanaReddy8801:
Jira Ticket: https://new-relic.atlassian.net/browse/NR-595360
## Summary
- Adds `recipes/newrelic/infrastructure/nrdot/mssql-otel/debian.yml` and `rhel.yml` — New Relic CLI install recipes for monitoring Microsoft SQL Server on Linux via the NRDOT OpenTelemetry collector, driven from https://docs.newrelic.com/docs/opentelemetry/database/mssql/linux-hosted/

I’ve attached the test [[documentation](https://docs.google.com/document/d/1GmdXanWcy4_ZVMtUCZmG9rMVHQbV3gBTUHktLljzsCQ/edit?tab=t.0)] covering several key integration scenarios for your review.


Nandu-pns:
It looks like SA_PASSWORD and a few other vars are getting inserted straight into the script text using `{{.NR_CLI_MSSQL_SA_PASSWORD}}`, instead of being read as an environment variable when the script runs. I think this is bit risky: if the password has characters like `` ` ``, `$(...)`, or `"` in it, they get treated as shell code, not just as password text — that's an injection risk. It also means the actual secret is written out in the generated script instead of staying in the environment.

The `redis` recipe already does this the safer way — `REDIS_PASSWORD="${NR_CLI_REDIS_PASSWORD:-}"` — it just reads the value from the environment at runtime. Could we update `debian.yml` and `rhel.yml` to do the same for `SA_PASSWORD`, `SERVER`, `PORT`, and `LOGIN_NAME`? Should be a simple swap with no behavior change. (Same issue is in the Windows recipe too, just worth fixing there separately.)

Nandu-pns:
One more thing — in `configure_database_user`, there's an `if/else` meant to let someone pass in a custom password for the monitoring login instead of a random one:

```bash
if [ -z "{{.NR_CLI_MSSQL_LOGIN_PASSWORD}}" ]; then
  NR_PASSWORD=$(tr -dc 'A-Za-z0-9' < /dev/urandom | head -c 24)
else
  NR_PASSWORD="{{.NR_CLI_MSSQL_LOGIN_PASSWORD}}"
fi
```

`NR_CLI_MSSQL_LOGIN_PASSWORD` is never added to `inputVars`, so it always resolves to an empty string — which means the `else` branch (the custom password) can never actually run. Every install just takes the `if` branch and generates a random password, so the `else` here is dead code.

Not a security issue, just dead code right now. The `rabbitmq` recipe shows how to wire this up properly if a custom password is actually wanted:
```yaml
inputVars:
  - name: NR_CLI_RABBITMQ_PASSWORD
    prompt: "RabbitMQ password: "
    secret: true
```
It's declared in `inputVars` with `secret: true`, so the CLI prompts for it and the value is real by the time the script runs.

If a custom password isn't actually needed, simplest is to just drop the `if/else` and always generate the random one — up to you which way you'd rather go.

recipes/newrelic/infrastructure/nrdot/mssql-otel/debian.yml
```diff
@@ -0,0 +1,615 @@
...
+          fi
+
+          case "$REGION" in
+            staging) OTLP_ENDPOINT="https://staging-otlp.nr-data.net" ;;
+            EU)      OTLP_ENDPOINT="https://otlp.eu01.nr-data.net" ;;
+            JP)      OTLP_ENDPOINT="https://otlp.jp.nr-data.net" ;;
+            *)       OTLP_ENDPOINT="https://otlp.nr-data.net" ;;
+          esac
+
+          cat > "$CONFIG_PATH" << EOF
```
Nandu-pns, line 299: One more thing, in `create_collector_config` (same in both `debian.yml` and `rhel.yml`):

```bash
CONFIG_PATH="${CONFIG_DIR}/mssql-config.yaml"
...
mkdir -p "$CONFIG_DIR"
...
cat > "$CONFIG_PATH" << EOF
receivers:
  nrsqlserver:
    ...
    password: ${NR_PASSWORD}
```

`mssql-config.yaml` ends up with the SQL Server monitoring password in plaintext, plus the New Relic license key (`api-key: ${LICENSE_KEY}`) — but the file and its directory never get a `chmod`, so on a normal umask it's world-readable. Any local user on the box can read both.

The `redis` NRDOT recipe already in this repo does this the right way:
```bash
chmod 600 "$CONFIG_FILE"
...
api-key: ${env:NEW_RELIC_LICENSE_KEY}
```
It `chmod 600`s the config file, and passes the license key through the collector's own `${env:...}` substitution instead of writing the literal value into the file. The `chmod` is the important part here — could we add `chmod 600 "$CONFIG_PATH"` after it's written, same as redis does?

RamanaReddy8801: Fixed in both `debian.yml` and `rhel.yml` — added chmod 600 "$CONFIG_PATH" right after the config file is fully written (same spot as the redis recipe)

recipes/newrelic/infrastructure/nrdot/mssql-otel/debian.yml
```diff
@@ -0,0 +1,617 @@
...
+          PORT="${NR_CLI_MSSQL_PORT:-1433}"
+          SA_PASSWORD="${NR_CLI_MSSQL_SA_PASSWORD:-}"
+          SQL_FILE="/tmp/nr-mssql-grant.sql"
+          PW_FILE="/tmp/nr-mssql-newrelic-pw.tmp"
+
+          NR_PASSWORD=$(tr -dc 'A-Za-z0-9' < /dev/urandom | head -c 24)
+          printf '%s' "$NR_PASSWORD" > "$PW_FILE"
+          chmod 600 "$PW_FILE"
+
+          cat > "$SQL_FILE" << EOF
```
Nandu-pns, line 200: Similar issue as the config file, but on `SQL_FILE="/tmp/nr-mssql-grant.sql"` (same in both `debian.yml` and `rhel.yml`): it's written with the new monitoring password in plain text but never gets `chmod 600` like `PW_FILE` does — so it's readable by any user on the box. Worse, if `sqlcmd` fails, the script exits before `rm -f "$SQL_FILE"` ever runs, so the file (with the password) is left behind permanently.

Could you add `chmod 600 "$SQL_FILE"` right after it's written, and clean up both temp files even on failure (e.g. a `trap ... EXIT`)? Or use `mktemp` instead of a fixed name — other recipes here (`agent-control`, PHP agent) already do that, and it gets you `600` perms + an unguessable filename in one step.

RamanaReddy8801: Fixed in both `debian.yml` and `rhel.yml`. Went with `mktemp` for `SQL_FILE `(**unguessable name + 600 perms in one step**, as you suggested), plus a trap cleanup **EXIT** that removes it on every exit path — including a failed **sqlcmd** run.

recipes/newrelic/infrastructure/nrdot/mssql-otel/debian.yml
```diff
@@ -0,0 +1,627 @@
...
+          export PATH="/opt/mssql-tools18/bin:/opt/mssql-tools/bin:$PATH"
+
+          SERVER="${NR_CLI_MSSQL_SERVER:-localhost}"
+          PORT="${NR_CLI_MSSQL_PORT:-1433}"
+          SA_PASSWORD=$(cat << 'NR_SA_PW_EOF'
+          {{.NR_CLI_MSSQL_SA_PASSWORD}}
+          NR_SA_PW_EOF
+          )
+
+          RESULT=$(sqlcmd -S "${SERVER},${PORT}" -U sa -P "${SA_PASSWORD}" -C -b -h -1 -Q "SET NOCOUNT ON; SELECT SERVERPROPERTY('ProductMajorVersion');" 2>&1)
```
Nandu-pns: Every `sqlcmd` call in this file passes the `sa` password straight on the command line via `-P "${SA_PASSWORD}"` (here, plus lines 233 and 266 for the grant script and the `ALTER LOGIN` fallback). That's visible to any other local user via `ps aux`/`/proc/<pid>/cmdline` for as long as each call is running.

This isn't just us being extra careful — Microsoft's own `sqlcmd` docs say it outright: *"Using `-P` is insecure. Avoid giving the password on the command line... Alternatively, use the `SQLCMDPASSWORD` environment variable."* And the fix is already sitting in this same feature — the Windows recipes in #1403 do exactly that (`$env:SQLCMDPASSWORD = $saPassword`, no `-P` anywhere). Can we bring these three calls in line with that?

recipes/newrelic/infrastructure/nrdot/mssql-otel/rhel.yml
```diff
@@ -0,0 +1,627 @@
...
+          export PATH="/opt/mssql-tools18/bin:/opt/mssql-tools/bin:$PATH"
+
+          SERVER="${NR_CLI_MSSQL_SERVER:-localhost}"
+          PORT="${NR_CLI_MSSQL_PORT:-1433}"
+          SA_PASSWORD=$(cat << 'NR_SA_PW_EOF'
+          {{.NR_CLI_MSSQL_SA_PASSWORD}}
+          NR_SA_PW_EOF
+          )
+
+          RESULT=$(sqlcmd -S "${SERVER},${PORT}" -U sa -P "${SA_PASSWORD}" -C -b -h -1 -Q "SET NOCOUNT ON; SELECT SERVERPROPERTY('ProductMajorVersion');" 2>&1)
```
Nandu-pns: Same as `debian.yml` — `-P "${SA_PASSWORD}"` puts the `sa` password on the command line here too (again at lines 233 and 266). Microsoft's own `sqlcmd` docs call `-P` insecure for exactly this reason and point at `SQLCMDPASSWORD` instead — which is already how the Windows recipe in #1403 handles this. Same fix would close this out here.

recipes/newrelic/infrastructure/nrdot/mssql-otel/rds-debian.yml
```diff
@@ -0,0 +1,822 @@
...
+
+          SERVER="{{.NR_CLI_MSSQL_SERVER}}"
+          PORT="${NR_CLI_MSSQL_PORT:-1433}"
+          MASTER_USER="{{.NR_CLI_MSSQL_MASTER_USER}}"
+          MASTER_PASSWORD=$(cat << 'NR_MASTER_PW_EOF'
+          {{.NR_CLI_MSSQL_MASTER_PASSWORD}}
+          NR_MASTER_PW_EOF
+          )
+
+          RESULT=$(sqlcmd -S "${SERVER},${PORT}" -U "${MASTER_USER}" -P "${MASTER_PASSWORD}" -C -b -h -1 -Q "SET NOCOUNT ON; SELECT SERVERPROPERTY('ProductMajorVersion');" 2>&1)
```
Nandu-pns: Same thing as `debian.yml`/`rhel.yml`, just with the RDS master creds — `-P "${MASTER_PASSWORD}"` puts the RDS master password on the command line here (and again at lines 238 and 272). Microsoft's `sqlcmd` docs explicitly warn against `-P` for this and recommend `SQLCMDPASSWORD` instead — which is already what the Windows RDS recipe in #1403 does (`$env:SQLCMDPASSWORD = $masterPassword`). Worth the same fix here.

recipes/newrelic/infrastructure/nrdot/mssql-otel/rds-rhel.yml
```diff
@@ -0,0 +1,822 @@
...
+
+          SERVER="{{.NR_CLI_MSSQL_SERVER}}"
+          PORT="${NR_CLI_MSSQL_PORT:-1433}"
+          MASTER_USER="{{.NR_CLI_MSSQL_MASTER_USER}}"
+          MASTER_PASSWORD=$(cat << 'NR_MASTER_PW_EOF'
+          {{.NR_CLI_MSSQL_MASTER_PASSWORD}}
+          NR_MASTER_PW_EOF
+          )
+
+          RESULT=$(sqlcmd -S "${SERVER},${PORT}" -U "${MASTER_USER}" -P "${MASTER_PASSWORD}" -C -b -h -1 -Q "SET NOCOUNT ON; SELECT SERVERPROPERTY('ProductMajorVersion');" 2>&1)
```
Nandu-pns: Same as `rds-debian.yml` — RDS master password on the command line via `-P "${MASTER_PASSWORD}"` (here, plus lines 238 and 272). Same `SQLCMDPASSWORD` fix as the others.

recipes/newrelic/infrastructure/nrdot/mssql-otel/debian.yml
```diff
@@ -0,0 +1,627 @@
...
+          SQL_FILE=$(mktemp /tmp/nr-mssql-grant.sql.XXXXXX)
+          trap 'rm -f "$SQL_FILE"' EXIT
+
+          NR_PASSWORD=$(tr -dc 'A-Za-z0-9' < /dev/urandom | head -c 24)
+
+          cat > "$SQL_FILE" << EOF
+          USE [master];
+          GO
+          IF NOT EXISTS (SELECT 1 FROM sys.server_principals WHERE name = '${LOGIN_NAME}')
+            CREATE LOGIN [${LOGIN_NAME}] WITH PASSWORD = '${NR_PASSWORD}';
```
Nandu-pns, line 204: `LOGIN_NAME` comes straight from `NR_CLI_MSSQL_LOGIN_NAME` (default `newrelic`) with nothing validating it before it lands in `[${LOGIN_NAME}]` — and it's not just this line, the same unescaped bracket shows up 8 times in this task (`CREATE LOGIN`, `ALTER LOGIN`, three `GRANT`s, `CREATE USER ... FOR LOGIN` twice, `GRANT VIEW DATABASE STATE`). T-SQL escapes a literal `]` inside a bracket-quoted identifier by doubling it (`]]`) — a single `]` in the username breaks out of the identifier and whatever follows runs as `sa`. Could we validate `LOGIN_NAME` (something like `^[A-Za-z][A-Za-z0-9_]*$`) right after it's read, before any of this runs?

recipes/newrelic/infrastructure/nrdot/mssql-otel/rhel.yml
```diff
@@ -0,0 +1,627 @@
...
+          SQL_FILE=$(mktemp /tmp/nr-mssql-grant.sql.XXXXXX)
+          trap 'rm -f "$SQL_FILE"' EXIT
+
+          NR_PASSWORD=$(tr -dc 'A-Za-z0-9' < /dev/urandom | head -c 24)
+
+          cat > "$SQL_FILE" << EOF
+          USE [master];
+          GO
+          IF NOT EXISTS (SELECT 1 FROM sys.server_principals WHERE name = '${LOGIN_NAME}')
+            CREATE LOGIN [${LOGIN_NAME}] WITH PASSWORD = '${NR_PASSWORD}';
```
Nandu-pns, line 204: Same as `debian.yml` — `LOGIN_NAME` is unvalidated and reused unescaped in `[${LOGIN_NAME}]` across 8 statements in this task. A `]` in the username breaks out of the bracket-quoted identifier and runs as `sa`. Same fix needed here.

recipes/newrelic/infrastructure/nrdot/mssql-otel/rds-debian.yml
```diff
@@ -0,0 +1,822 @@
...
+          SQL_FILE=$(mktemp /tmp/nr-mssql-grant.sql.XXXXXX)
+          trap 'rm -f "$SQL_FILE"' EXIT
+
+          NR_PASSWORD=$(tr -dc 'A-Za-z0-9' < /dev/urandom | head -c 24)
+
+          cat > "$SQL_FILE" << EOF
+          USE [master];
+          GO
+          IF NOT EXISTS (SELECT 1 FROM sys.server_principals WHERE name = '${LOGIN_NAME}')
+            CREATE LOGIN [${LOGIN_NAME}] WITH PASSWORD = '${NR_PASSWORD}';
```
Nandu-pns, line 209: Same issue as on `debian.yml`/`rhel.yml`, just with higher stakes — `LOGIN_NAME` is unvalidated and gets reused unescaped in `[${LOGIN_NAME}]` across 8 statements in this task. A `]` in the username breaks out of the bracket-quoted identifier and whatever follows runs as the RDS master user. Could we validate `LOGIN_NAME` (e.g. `^[A-Za-z][A-Za-z0-9_]*$`) before any of this runs?

recipes/newrelic/infrastructure/nrdot/mssql-otel/rds-rhel.yml
```diff
@@ -0,0 +1,822 @@
...
+          SQL_FILE=$(mktemp /tmp/nr-mssql-grant.sql.XXXXXX)
+          trap 'rm -f "$SQL_FILE"' EXIT
+
+          NR_PASSWORD=$(tr -dc 'A-Za-z0-9' < /dev/urandom | head -c 24)
+
+          cat > "$SQL_FILE" << EOF
+          USE [master];
+          GO
+          IF NOT EXISTS (SELECT 1 FROM sys.server_principals WHERE name = '${LOGIN_NAME}')
+            CREATE LOGIN [${LOGIN_NAME}] WITH PASSWORD = '${NR_PASSWORD}';
```
Nandu-pns, line 209: Same as `rds-debian.yml` — unvalidated `LOGIN_NAME` reused unescaped in `[${LOGIN_NAME}]` across this task, and a `]` in it breaks out and runs as the RDS master user. Same validation fix needed here.

recipes/newrelic/infrastructure/nrdot/mssql-otel/debian.yml
RamanaReddy8801: Fixed — switched to `SQLCMDPASSWORD` as suggested, `-P` removed from all three `sqlcmd` calls in each file. Verified on a real RDS SQL Server instance by polling `/proc/<pid>/cmdline` during an actual install — captured argv is `sqlcmd -S ... -U admin -C -b ..., zero -P`, zero password anywhere.

recipes/newrelic/infrastructure/nrdot/mssql-otel/rds-rhel.yml
RamanaReddy8801: Fixed — switched to `SQLCMDPASSWORD` as suggested, `-P` removed from all three `sqlcmd` calls in each file. Verified on a real RDS SQL Server instance by polling `/proc/<pid>/cmdline` during an actual install — captured argv is `sqlcmd -S ... -U admin -C -b ..., zero -P`, zero password anywhere.

recipes/newrelic/infrastructure/nrdot/mssql-otel/rds-debian.yml
RamanaReddy8801: Fixed — switched to `SQLCMDPASSWORD` as suggested, `-P` removed from all three `sqlcmd` calls in each file. Verified on a real RDS SQL Server instance by polling `/proc/<pid>/cmdline` during an actual install — captured argv is `sqlcmd -S ... -U admin -C -b ..., zero -P`, zero password anywhere.

recipes/newrelic/infrastructure/nrdot/mssql-otel/rhel.yml
RamanaReddy8801: Fixed — switched to `SQLCMDPASSWORD` as suggested, `-P` removed from all three `sqlcmd` calls in each file. Verified on a real RDS SQL Server instance by polling `/proc/<pid>/cmdline` during an actual install — captured argv is `sqlcmd -S ... -U admin -C -b ..., zero -P`, zero password anywhere.

recipes/newrelic/infrastructure/nrdot/mssql-otel/debian.yml
RamanaReddy8801: Fixed — added `^[A-Za-z][A-Za-z0-9_]*$` validation right after `LOGIN_NAME` is read, before it touches the CREATE LOGIN line or any of the other 7 bracket-quoted statements in the task.

Also found and fixed a separate bug while verifying this on real infra: `LOGIN_NAME` (and `PORT`) were using `${NR_CLI_MSSQL_X:-default}` shell env-var expansion instead of Task's `{{.NR_CLI_MSSQL_X}}` template substitution that `SERVER/MASTER_USER/MASTER_PASSWORD` already correctly use. Since the CLI never exports these into the process environment, `LOGIN_NAME` was silently falling back to `newrelic` regardless of what was typed — which meant the injection test looked like a pass for the wrong reason (the payload never reached the check at all). Fixed to use `{{.}}` templating consistently; re-tested on real RDS afterward with `newrelic1]; DROP LOGIN admin; `-- — now correctly rejected with Invalid monitoring username: ... before any SQL runs.

recipes/newrelic/infrastructure/nrdot/mssql-otel/rhel.yml
RamanaReddy8801: Fixed — added `^[A-Za-z][A-Za-z0-9_]*$` validation right after `LOGIN_NAME` is read, before it touches the CREATE LOGIN line or any of the other 7 bracket-quoted statements in the task.

Also found and fixed a separate bug while verifying this on real infra: `LOGIN_NAME` (and `PORT`) were using `${NR_CLI_MSSQL_X:-default}` shell env-var expansion instead of Task's `{{.NR_CLI_MSSQL_X}}` template substitution that `SERVER/MASTER_USER/MASTER_PASSWORD` already correctly use. Since the CLI never exports these into the process environment, `LOGIN_NAME` was silently falling back to `newrelic` regardless of what was typed — which meant the injection test looked like a pass for the wrong reason (the payload never reached the check at all). Fixed to use `{{.}}` templating consistently; re-tested on real RDS afterward with `newrelic1]; DROP LOGIN admin; `-- — now correctly rejected with Invalid monitoring username: ... before any SQL runs.

recipes/newrelic/infrastructure/nrdot/mssql-otel/rds-debian.yml
RamanaReddy8801: Fixed — added `^[A-Za-z][A-Za-z0-9_]*$` validation right after `LOGIN_NAME` is read, before it touches the CREATE LOGIN line or any of the other 7 bracket-quoted statements in the task.

Also found and fixed a separate bug while verifying this on real infra: `LOGIN_NAME` (and `PORT`) were using `${NR_CLI_MSSQL_X:-default}` shell env-var expansion instead of Task's `{{.NR_CLI_MSSQL_X}}` template substitution that `SERVER/MASTER_USER/MASTER_PASSWORD` already correctly use. Since the CLI never exports these into the process environment, `LOGIN_NAME` was silently falling back to `newrelic` regardless of what was typed — which meant the injection test looked like a pass for the wrong reason (the payload never reached the check at all). Fixed to use `{{.}}` templating consistently; re-tested on real RDS afterward with `newrelic1]; DROP LOGIN admin; `-- — now correctly rejected with Invalid monitoring username: ... before any SQL runs.

recipes/newrelic/infrastructure/nrdot/mssql-otel/rds-rhel.yml
RamanaReddy8801: Fixed — added `^[A-Za-z][A-Za-z0-9_]*$` validation right after `LOGIN_NAME` is read, before it touches the CREATE LOGIN line or any of the other 7 bracket-quoted statements in the task.

Also found and fixed a separate bug while verifying this on real infra: `LOGIN_NAME` (and `PORT`) were using `${NR_CLI_MSSQL_X:-default}` shell env-var expansion instead of Task's `{{.NR_CLI_MSSQL_X}}` template substitution that `SERVER/MASTER_USER/MASTER_PASSWORD` already correctly use. Since the CLI never exports these into the process environment, `LOGIN_NAME` was silently falling back to `newrelic` regardless of what was typed — which meant the injection test looked like a pass for the wrong reason (the payload never reached the check at all). Fixed to use `{{.}}` templating consistently; re-tested on real RDS afterward with `newrelic1]; DROP LOGIN admin; `-- — now correctly rejected with Invalid monitoring username: ... before any SQL runs.

Nandu-pns (approved):
LGTM

PR #1416, opened 2026-08-03
feat(nrdot-oracle): scaffold debian.yml/rhel.yml metadata and inputVars

RamanaReddy8801:
**Jira Ticket:**

Summary
Adds recipes/newrelic/infrastructure/nrdot/oracle-otel/debian.yml and rhel.yml — New Relic CLI install recipes for monitoring Oracle  on Linux via the NRDOT OpenTelemetry collector, driven from https://docs.newrelic.com/docs/opentelemetry/database/otel-oracledb

- I’ve attached the test [documentation](https://docs.google.com/document/d/1P7TQG6-pWrTkWRQQgZCVXbAtyHQQgrDUXf8o7D3cF4Y/edit?tab=t.0) covering several key integration scenarios for your review


recipes/newrelic/infrastructure/nrdot/oracle-otel/oci-liunx.yml
```diff
@@ -0,0 +1,1024 @@
...
...N V_\$SYSMETRIC TO ${DB_USER} CONTAINER=ALL;\nGRANT SELECT ON V_\$CON_SYSMETRIC TO ${DB_USER} CONTAINER=ALL;\nGRANT SELECT ON V_\$SESSION TO ${DB_USER} CONTAINER=ALL;\nGRANT SELECT ON V_\$RESOURCE_LIMIT TO ${DB_USER} CONTAINER=ALL;\nGRANT SELECT ON V_\$OSSTAT TO ${DB_USER} CONTAINER=ALL;\nGRANT SELECT ON V_\$SGAINFO TO ${DB_USER} CONTAINER=ALL;\nGRANT SELECT ON V_\$ROWCACHE TO ${DB_USER} CONTAINER=ALL;\nGRANT SELECT ON V_\$PARAMETER TO ${DB_USER} CONTAINER=ALL;\nGRANT SELECT ON CDB_TABLESPACE_USAGE_METRICS TO ${DB_USER} CONTAINER=ALL;\nGRANT SELECT ON CDB_TABLESPACES TO ${DB_USER} CONTAINER=ALL;\nGRANT SELECT ON DBA_DATA_FILES TO ${DB_USER} CONTAINER=ALL;\nGRANT SELECT ON DBA_FREE_SPACE TO ${DB_USER} CONTAINER=ALL;\nGRANT SELECT ON DBA_RECYCLEBIN TO ${DB_USER} CONTAINER=ALL;\nGRANT SELECT ON V_\$SQL TO ${DB_USER} CONTAINER=ALL;\nGRANT SELECT ON V_\$SQL_PLAN TO ${DB_USER} CONTAINER=ALL;\nGRANT SELECT ON V_\$LOCK TO ${DB_USER} CONTAINER=ALL;\nGRANT SELECT ON V_\$SESSION_EVENT TO ${DB_USER} CONTAINER=ALL;\nGRANT SELECT ON DBA_PROCEDURES TO ${DB_USER} CONTAINER=ALL;\nGRANT SELECT ON DBA_OBJECTS TO ${DB_USER} CONTAINER=ALL;\nEXIT;\n' | sudo su - oracle -c 'sqlplus -S / as sysdba'" 2>&1)
```
Nandu-pns: `DB_USER` and `NR_PASSWORD` get dropped straight into the SSH command as plain text:
```bash
RESULT=$(ssh -o StrictHostKeyChecking=no -o BatchMode=yes "${SSH_USER}@${HOST}" \
  "printf 'WHENEVER SQLERROR EXIT SQL.SQLCODE\nALTER SESSION SET CONTAINER = CDB\$ROOT;\nCREATE USER ${DB_USER} IDENTIFIED BY \"${NR_PASSWORD}\" CONTAINER=ALL;\n...' | sudo su - oracle -c 'sqlplus -S / as sysdba'")
```
The whole line is one double-quoted string, so both variables get filled in locally before the command is even sent over SSH — nothing escapes their contents first. `DB_USER` can come straight from the `NEW_USERNAME` prompt a few lines up, and nothing stops someone from typing a single quote in there. If they do, it breaks out of the remote `printf`'s quoting and whatever comes after runs as a real shell command on the Oracle box, as `${SSH_USER}` — before it even reaches `sudo su - oracle`. Same issue applies to the PDB-mode block right below (line 314). Could we escape these before they go in, or better, write the SQL out to a local file and ship that over instead of building it as one long inline string?

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-debian.yml
```diff
@@ -0,0 +1,1004 @@
...
+
+          if [ -z "{{.NR_CLI_ORACLE_LOGIN_PASSWORD}}" ]; then
+            NR_PASSWORD=$(tr -dc 'A-Za-z0-9' < /dev/urandom | head -c 24)
+          else
+            NR_PASSWORD="{{.NR_CLI_ORACLE_LOGIN_PASSWORD}}"
+          fi
+          printf '%s' "$NR_PASSWORD" > "$PW_FILE"
+          chmod 600 "$PW_FILE"
+
+          RESULT=$(sqlplus -S "${ADMIN_USER}/${ADMIN_PASSWORD}@${HOST}:${PORT}/${SERVICE_NAME}" << SQL 2>&1
```
Nandu-pns: The RDS master password is being passed right on the `sqlplus` command line:
```bash
sqlplus -S "${ADMIN_USER}/${ADMIN_PASSWORD}@${HOST}:${PORT}/${SERVICE_NAME}" << SQL
```
(same pattern at line 108 and line 202 too). Anything on a command line is visible to every other user on the box via `ps aux` or `/proc/<pid>/cmdline` while the process is running. So for as long as this task is executing, any other local user can just read the RDS master password off the process list — no file permissions to bypass, nothing to guess. That's a step worse than the file-permission issues we caught in #1412.

We're already handling this correctly elsewhere in the repo — `mysql/rds.yml` does basically the same thing (RDS admin creds + a CLI client) but keeps the password out of the command line entirely:
```bash
CONNECT_OUTPUT=$(MYSQL_PWD="${RDS_ADMIN_PASS}" mysql --host="${RDS_ENDPOINT}" --port="${RDS_PORT}" --user="${RDS_ADMIN_USER}" -N -s -e "SELECT 1;" 2>&1)
```
`sqlplus` doesn't have a direct equivalent to `MYSQL_PWD`, but we can get the same effect with `sqlplus -S /nolog` and a `CONNECT` line inside the heredoc, so the password only ever shows up in heredoc content, never in argv. Can we switch to that here (or use a wallet if that's already the standard for Oracle elsewhere)?

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-rhel.yml
```diff
@@ -0,0 +1,1004 @@
...
+
+          if [ -z "{{.NR_CLI_ORACLE_LOGIN_PASSWORD}}" ]; then
+            NR_PASSWORD=$(tr -dc 'A-Za-z0-9' < /dev/urandom | head -c 24)
+          else
+            NR_PASSWORD="{{.NR_CLI_ORACLE_LOGIN_PASSWORD}}"
+          fi
+          printf '%s' "$NR_PASSWORD" > "$PW_FILE"
+          chmod 600 "$PW_FILE"
+
+          RESULT=$(sqlplus -S "${ADMIN_USER}/${ADMIN_PASSWORD}@${HOST}:${PORT}/${SERVICE_NAME}" << SQL 2>&1
```
Nandu-pns: Same as the comment on `rds-debian.yml` — the RDS master password is passed right on the `sqlplus` command line here too:
```bash
sqlplus -S "${ADMIN_USER}/${ADMIN_PASSWORD}@${HOST}:${PORT}/${SERVICE_NAME}" << SQL
```
(same pattern at line 112 and line 202). Anything on a command line is visible to every other user on the box via `ps aux` or `/proc/<pid>/cmdline` while the process is running, so this leaks the RDS master password to any other local user for as long as the task runs.

`mysql/rds.yml` avoids this exact thing for the same RDS-admin-creds-plus-CLI-client scenario:
```bash
CONNECT_OUTPUT=$(MYSQL_PWD="${RDS_ADMIN_PASS}" mysql --host="${RDS_ENDPOINT}" --port="${RDS_PORT}" --user="${RDS_ADMIN_USER}" -N -s -e "SELECT 1;" 2>&1)
```
Can we do the equivalent here with `sqlplus -S /nolog` + a `CONNECT` line inside the heredoc, so the password stays out of argv?

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-debian.yml
```diff
@@ -0,0 +1,1004 @@
...
+          WHENEVER SQLERROR EXIT SQL.SQLCODE
+          CREATE USER ${LOGIN_NAME} IDENTIFIED BY "${NR_PASSWORD}";
```
Nandu-pns, line 279: `LOGIN_NAME` goes straight into the `CREATE USER` statement as-is:
```bash
CREATE USER ${LOGIN_NAME} IDENTIFIED BY "${NR_PASSWORD}";
GRANT CONNECT TO ${LOGIN_NAME};
```
`LOGIN_NAME` can come from the `NEW_USERNAME` prompt earlier in this task, and there's no check on what someone types there. A username with a stray `;` in it would end the `CREATE USER` statement early and let whatever follows run as the RDS master user. Since this whole block runs with admin privileges, could we lock the username down to something like `^[A-Za-z][A-Za-z0-9_]*$` before it touches any SQL?

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-rhel.yml
```diff
@@ -0,0 +1,1004 @@
...
+          WHENEVER SQLERROR EXIT SQL.SQLCODE
+          CREATE USER ${LOGIN_NAME} IDENTIFIED BY "${NR_PASSWORD}";
```
Nandu-pns, line 279: Same issue as `rds-debian.yml` — `LOGIN_NAME` goes straight into the `CREATE USER` statement with no validation:
```bash
CREATE USER ${LOGIN_NAME} IDENTIFIED BY "${NR_PASSWORD}";
GRANT CONNECT TO ${LOGIN_NAME};
```
It can come from the `NEW_USERNAME` prompt, and a stray `;` in there would end the statement early and let whatever follows run as the RDS master user. Could we validate the username against something like `^[A-Za-z][A-Za-z0-9_]*$` before it's used in SQL?

recipes/newrelic/infrastructure/nrdot/oracle-otel/oci-linux.yml
```diff
@@ -0,0 +1,1024 @@
...
+          SERVICE_NAME="{{.NR_CLI_ORACLE_SERVICE_NAME}}"
+          REGION="{{.NEW_RELIC_REGION}}"
+          LICENSE_KEY="{{.NEW_RELIC_LICENSE_KEY}}"
+          CONFIG_DIR="/etc/nrdot-collector"
+          CONFIG_PATH="${CONFIG_DIR}/oracle-config.yaml"
+          ENV_FILE="${CONFIG_DIR}/nrdot-collector.conf"
+          PW_FILE="/tmp/nr-oracle-newrelic-pw.tmp"
+          USER_FILE="/tmp/nr-oracle-db-user.tmp"
+
+          mkdir -p "$CONFIG_DIR"
```
Nandu-pns, line 439: `oracle-config.yaml` ends up holding both the monitoring password and the license key in plain text under `/etc/nrdot-collector`, but we never lock down its permissions after writing it — on a normal umask that means any local user can just open it and read both. Same gap we already found and fixed on the MSSQL side.

The `redis` recipe already gets this right:
```bash
# 600: only root should read sensitive credentials (license key, redis password).
chmod 600 "$CONFIG_FILE"
```
Can we add `chmod 600 "$CONFIG_PATH"` right after it's written here?

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-debian.yml
```diff
@@ -0,0 +1,1004 @@
...
+          SERVICE_NAME="{{.NR_CLI_ORACLE_SERVICE_NAME}}"
+          REGION="{{.NEW_RELIC_REGION}}"
+          LICENSE_KEY="{{.NEW_RELIC_LICENSE_KEY}}"
+          CONFIG_DIR="/etc/nrdot-collector"
+          CONFIG_PATH="${CONFIG_DIR}/oracle-config.yaml"
+          ENV_FILE="${CONFIG_DIR}/nrdot-collector.conf"
+          PW_FILE="/tmp/nr-oracle-rds-newrelic-pw.tmp"
+          USER_FILE="/tmp/nr-oracle-rds-db-user.tmp"
+
+          mkdir -p "$CONFIG_DIR"
```
Nandu-pns, line 351: Same as `oci-liunx.yml` — `oracle-config.yaml` holds the monitoring password and license key in plain text but never gets its permissions locked down, so it's world-readable on a normal umask.

The `redis` recipe already does this correctly:
```bash
# 600: only root should read sensitive credentials (license key, redis password).
chmod 600 "$CONFIG_FILE"
```
Can we add `chmod 600 "$CONFIG_PATH"` right after it's written here too?

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-rhel.yml
```diff
@@ -0,0 +1,1004 @@
...
+          SERVICE_NAME="{{.NR_CLI_ORACLE_SERVICE_NAME}}"
+          REGION="{{.NEW_RELIC_REGION}}"
+          LICENSE_KEY="{{.NEW_RELIC_LICENSE_KEY}}"
+          CONFIG_DIR="/etc/nrdot-collector"
+          CONFIG_PATH="${CONFIG_DIR}/oracle-config.yaml"
+          ENV_FILE="${CONFIG_DIR}/nrdot-collector.conf"
+          PW_FILE="/tmp/nr-oracle-rds-newrelic-pw.tmp"
+          USER_FILE="/tmp/nr-oracle-rds-db-user.tmp"
+
+          mkdir -p "$CONFIG_DIR"
```
Nandu-pns, line 351: Same as `oci-liunx.yml`/`rds-debian.yml` — `oracle-config.yaml` holds the monitoring password and license key in plain text but never gets its permissions locked down.

The `redis` recipe already does this correctly:
```bash
# 600: only root should read sensitive credentials (license key, redis password).
chmod 600 "$CONFIG_FILE"
```
Can we add `chmod 600 "$CONFIG_PATH"` right after it's written here too?

recipes/newrelic/infrastructure/nrdot/oracle-otel/oci-liunx.yml
```diff
@@ -0,0 +1,1024 @@
...
+
+name: nrdot-collector-oracle-linux
+displayName: NRDOT Oracle Database (RHEL/OEL)
+description: New Relic install recipe for Oracle Database monitoring via NRDOT Collector on RHEL/OEL self-hosted environments
+repository: https://github.com/newrelic/nrdot-collector-releases
+
+installTargets:
+  - type: host
+    os: linux
+    platform: "oracle"
```
Nandu-pns, line 13: Small thing but worth catching before this ships — `platform: "oracle"` here only matches Oracle Linux (OEL) hosts. `displayName`/`description` both say this covers RHEL/OEL, but as written a plain RHEL or CentOS box will never see this recipe show up.

`redis/rhel.yml` handles this same split correctly with four separate `installTargets` entries (`amazon`/`centos`/`oracle`/`redhat`), and `rds-rhel.yml` right here in this PR does the same for `centos`/`redhat`/`oracle`. Can we give `oci-liunx.yml` the same treatment? Right now it silently drops half the OS support it advertises.

recipes/newrelic/infrastructure/nrdot/oracle-otel/oci-linux.yml
```diff
@@ -0,0 +1,1024 @@
+# Visit our schema definition for additional information on this file format.
+# https://github.com/newrelic/open-install-library/blob/main/docs/recipe-spec/recipe-spec.md#schema-definition
+# WORK IN PROGRESS - not for use.
+
+name: nrdot-collector-oracle-linux
```
Nandu-pns, line 5: Unrelated to the logic, but the filename itself — `oci-liunx.yml` — looks like a typo for `oci-linux.yml`. Every other nrdot recipe (`redis`, `elasticsearch-otel`, `mssql-otel`, `nginx`...) sticks to the plain `debian.yml`/`rhel.yml` naming, and even this PR's own title/description refer to `debian.yml`/`rhel.yml`. Worth renaming before merge — recipe filenames end up baked into the catalog path, so this isn't just cosmetic.

recipes/newrelic/infrastructure/nrdot/oracle-otel/oci-linux.yml
```diff
@@ -0,0 +1,1024 @@
...
+
+    configure_database_user:
+      cmds:
+        - |
+          CONTAINER_TYPE="{{.NR_CLI_ORACLE_CONTAINER_TYPE}}"
+          PDB_NAME="{{.NR_CLI_ORACLE_PDB_NAME}}"
+          LOGIN_NAME="{{.NR_CLI_ORACLE_LOGIN_NAME}}"
+          HOST="{{.NR_CLI_ORACLE_HOST}}"
+          SSH_USER="{{.NR_CLI_ORACLE_SSH_USER}}"
+          PW_FILE="/tmp/nr-oracle-newrelic-pw.tmp"
```
Nandu-pns, line 225: `PW_FILE`/`USER_FILE` get a `chmod 600` right after they're written, which is good, but the path itself is a fixed, guessable name in `/tmp` (`/tmp/nr-oracle-newrelic-pw.tmp`), and there's no `trap ... EXIT` to guarantee cleanup. If the install gets interrupted (Ctrl-C) or a task fails outside the one `rm -f` path that's already handled, the plaintext monitoring password can be left sitting in `/tmp` indefinitely — and in the brief window before the `chmod` runs, a pre-planted symlink at that exact path would follow through too.

This is the same thing we flagged (and fixed) on the MSSQL PR for `SQL_FILE` — went with `mktemp` there instead of a fixed name, which gets you an unguessable filename + 600 perms in one step, plus a `trap EXIT` to clean it up on every exit path. Could we do the same here? Same issue applies to `rds-debian.yml` and `rds-rhel.yml` too.

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-debian.yml
```diff
@@ -0,0 +1,1004 @@
...
+    configure_database_user:
+      cmds:
+        - |
+          LOGIN_NAME="{{.NR_CLI_ORACLE_LOGIN_NAME}}"
+          HOST="{{.NR_CLI_ORACLE_HOST}}"
+          PORT="{{.NR_CLI_ORACLE_PORT}}"
+          SERVICE_NAME="{{.NR_CLI_ORACLE_SERVICE_NAME}}"
+          ADMIN_USER="{{.NR_CLI_ORACLE_ADMIN_USER}}"
+          ADMIN_PASSWORD="{{.NR_CLI_ORACLE_ADMIN_PASSWORD}}"
+          PW_FILE="/tmp/nr-oracle-rds-newrelic-pw.tmp"
```
Nandu-pns, line 200: Same fixed-name-in-`/tmp` + no-`trap` issue as `oci-liunx.yml` — `PW_FILE`/`USER_FILE` here (`/tmp/nr-oracle-rds-newrelic-pw.tmp`) get `chmod 600` after being written, but the name is predictable and there's no `trap ... EXIT`, so an interrupted run can leave the plaintext RDS monitoring password sitting in `/tmp`. Same fix as the MSSQL PR — `mktemp` + `trap EXIT` cleanup would close this.

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-rhel.yml
```diff
@@ -0,0 +1,1004 @@
...
+    configure_database_user:
+      cmds:
+        - |
+          LOGIN_NAME="{{.NR_CLI_ORACLE_LOGIN_NAME}}"
+          HOST="{{.NR_CLI_ORACLE_HOST}}"
+          PORT="{{.NR_CLI_ORACLE_PORT}}"
+          SERVICE_NAME="{{.NR_CLI_ORACLE_SERVICE_NAME}}"
+          ADMIN_USER="{{.NR_CLI_ORACLE_ADMIN_USER}}"
+          ADMIN_PASSWORD="{{.NR_CLI_ORACLE_ADMIN_PASSWORD}}"
+          PW_FILE="/tmp/nr-oracle-rds-newrelic-pw.tmp"
```
Nandu-pns, line 200: Same as `rds-debian.yml`/`oci-liunx.yml` — `PW_FILE`/`USER_FILE` (`/tmp/nr-oracle-rds-newrelic-pw.tmp`) use a fixed, guessable `/tmp` path and rely on `chmod 600` alone with no `trap ... EXIT`. An interrupted run can leave the plaintext password behind. Can we bring this one in line with the `mktemp` + `trap EXIT` fix from the MSSQL PR as well?

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-debian.yml
Nandu-pns: Worth flagging while we're in here — the blast radius on this is bigger than just the `CREATE USER`/`GRANT CONNECT` lines. The same tainted `LOGIN_UPPER` gets reused in ~28 `EXEC rdsadmin.rdsadmin_util.grant_sys_object('...', '${LOGIN_UPPER}', 'SELECT', false)` calls right after this, all running with RDS-admin-equivalent privileges. So whatever validation goes on `LOGIN_NAME` needs to happen before any of that block runs, not just before the `CREATE USER` line — otherwise the same injection is still reachable a few lines further down, just through a different statement.

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-rhel.yml
Nandu-pns: Same note as on `rds-debian.yml` — this isn't just the `CREATE USER`/`GRANT CONNECT` lines. `LOGIN_UPPER` (derived straight from the unvalidated `LOGIN_NAME`) also feeds ~28 `EXEC rdsadmin.rdsadmin_util.grant_sys_object('...', '${LOGIN_UPPER}', 'SELECT', false)` calls right below, all running with RDS-admin-equivalent privileges. Whatever validation gets added needs to sit before that whole block, not just before `CREATE USER`.

recipes/newrelic/infrastructure/nrdot/oracle-otel/oci-linux.yml
RamanaReddy8801: Tested this, and running chmod 600 breaks the service for two reasons:

- **Process User**: The `nrdot-collector` process runs as user nrdot-collector, not root.

- **Permission Block**: Setting 600 on a root-owned file prevents non-root users from reading it, blocking the collector from reading `oracle-config.yaml`

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-debian.yml
RamanaReddy8801: Tested this, and running chmod 600 breaks the service for two reasons:

- **Process User**: The `nrdot-collector` process runs as user nrdot-collector, not root.
- **Permission Block**: Setting 600 on a root-owned file prevents non-root users from reading it, blocking the collector from reading `oracle-config.yaml`

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-rhel.yml
RamanaReddy8801: Tested this, and running chmod 600 breaks the service for two reasons:

- **Process User**: The `nrdot-collector` process runs as user nrdot-collector, not root.
- **Permission Block**: Setting 600 on a root-owned file prevents non-root users from reading it, blocking the collector from reading `oracle-config.yaml`

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-debian.yml
RamanaReddy8801: Fixed — switched to sqlplus -S /nolog with CONNECT as the first line inside the heredoc, so the password only ever appears in stdin content, never in argv. Verified on a real box by polling /proc/<pid>/cmdline during the actual install — no credentials visible.

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-rhel.yml
RamanaReddy8801: Fixed — switched to sqlplus -S /nolog with CONNECT as the first line inside the heredoc, so the password only ever appears in stdin content, never in argv. Verified on a real box by polling /proc/<pid>/cmdline during the actual install — no credentials visible.

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-debian.yml
RamanaReddy8801: Fixed — added `^[A-Za-z][A-Za-z0-9_]*$` validation right after the username is finalized (covers both the initial prompt value and the "create new username" override), before it touches the `CREATE USER` line or any of the `grant_sys_object` calls after it — addresses the follow-up about the whole block, not just the first statement. Tested on real RDS with `newrelic1`; `DROP USER admin` as the payload — rejected immediately, no SQL executed, `admin` confirmed untouched.

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-rhel.yml
RamanaReddy8801: Fixed — added `^[A-Za-z][A-Za-z0-9_]*$` validation right after the username is finalized (covers both the initial prompt value and the "create new username" override), before it touches the `CREATE USER` line or any of the `grant_sys_object` calls after it — addresses the follow-up about the whole block, not just the first statement. Tested on real RDS with `newrelic1`; `DROP USER admin` as the payload — rejected immediately, no SQL executed, `admin` confirmed untouched.

recipes/newrelic/infrastructure/nrdot/oracle-otel/oci-linux.yml
RamanaReddy8801: fixed, rename to `oci-linux.yml`

recipes/newrelic/infrastructure/nrdot/oracle-otel/oci-liunx.yml
RamanaReddy8801: Intentionally scoped to platform: oracle for now — this has only been tested against a real Oracle Linux (OCI) instance. The script itself has no OS-specific logic (same RPM install, same sqlplus/ssh flow across RHEL/CentOS/OEL).

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-rhel.yml
RamanaReddy8801: Fixed — added an `NR_SUCCESS` flag with `trap ... EXIT` on `PW_FILE/USER_FILE`: deleted if the task exits without reaching success, kept in place if it completes normally (needed for handoff to `create_collector_config`). Tested on real RDS with a forced `CREATE USER` failure (`SYSTEM` name conflict) temp files confirmed removed afterward.

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-debian.yml
RamanaReddy8801: Fixed — added an `NR_SUCCESS` flag with `trap ... EXIT` on `PW_FILE/USER_FILE`: deleted if the task exits without reaching success, kept in place if it completes normally (needed for handoff to `create_collector_config`). Tested on real RDS with a forced `CREATE USER` failure (`SYSTEM` name conflict) temp files confirmed removed afterward.

recipes/newrelic/infrastructure/nrdot/oracle-otel/oci-linux.yml
RamanaReddy8801: Fixed — added an `NR_SUCCESS` flag with `trap ... EXIT` on `PW_FILE/USER_FILE`: deleted if the task exits without reaching success, kept in place if it completes normally (needed for handoff to `create_collector_config`). Tested on real RDS with a forced `CREATE USER` failure (`SYSTEM` name conflict) temp files confirmed removed afterward.

recipes/newrelic/infrastructure/nrdot/oracle-otel/oci-liunx.yml
RamanaReddy8801: Fixed — the SQL is now written to a local heredoc variable and shipped over `ssh ... <<< "$SQL"` against a fixed remote command string (`sudo su - oracle -c 'sqlplus -S / as sysdba'`), instead of being spliced into the double-quoted remote command itself. Verified with a mock ssh that logs its exact argv — confirmed zero user-controlled content reaches the command line now.

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-debian.yml
Nandu-pns: Validation only guards the second query though. `LOGIN_UPPER` is built from the raw `LOGIN_NAME` at line 205 and gets used in the `EXISTS_RESULT` lookup (`SELECT username FROM dba_users WHERE username='${LOGIN_UPPER}'`) at line 210 — that query runs before the `^[A-Za-z][A-Za-z0-9_]*$` check, which only sits in front of the `CREATE USER`/`grant_sys_object` block further down. So the same injection window is still open on the very first statement that touches the DB, just with a narrower payload shape (breaking out of a single-quoted string in a SELECT instead of a CREATE USER). Worth moving the regex check up above line 205, before `LOGIN_UPPER` is ever built.

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-rhel.yml
Nandu-pns: Same gap as rds-debian.yml — the validation added lands after the `EXISTS_RESULT` existence-check query (which also interpolates the unvalidated `LOGIN_UPPER`), so that first SELECT is still injectable even though CREATE USER/grant_sys_object are now covered. Suggest moving the regex check to right after `LOGIN_UPPER` is computed, before it's used anywhere.

recipes/newrelic/infrastructure/nrdot/oracle-otel/oci-linux.yml
Nandu-pns: Fair point on plain chmod 600 — that would lock root out the collector too. But the collector RPM (installed a few tasks earlier in install_nrdot) creates the nrdot-collector system user/group as part of packaging, so by the time create_collector_config writes oracle-config.yaml, that user already exists. That opens up chown nrdot-collector:nrdot-collector "$CONFIG_PATH" && chmod 600 "$CONFIG_PATH" right after the file is written — keeps the plaintext DB password out of reach of every other local user/process on the box, while still letting the collector (running as that same user) read its own config. Would rather see that than leaving it on default permissions, since this file has a real credential in it.

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-debian.yml
Nandu-pns: Fair point on plain chmod 600 — that would lock root out the collector too. But the collector RPM (installed a few tasks earlier in install_nrdot) creates the nrdot-collector system user/group as part of packaging, so by the time create_collector_config writes oracle-config.yaml, that user already exists. That opens up chown nrdot-collector:nrdot-collector "$CONFIG_PATH" && chmod 600 "$CONFIG_PATH" right after the file is written — keeps the plaintext DB password out of reach of every other local user/process on the box, while still letting the collector (running as that same user) read its own config. Would rather see that than leaving it on default permissions, since this file has a real credential in it.

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-rhel.yml
Nandu-pns: Fair point on plain chmod 600 — that would lock root out the collector too. But the collector RPM (installed a few tasks earlier in install_nrdot) creates the nrdot-collector system user/group as part of packaging, so by the time create_collector_config writes oracle-config.yaml, that user already exists. That opens up chown nrdot-collector:nrdot-collector "$CONFIG_PATH" && chmod 600 "$CONFIG_PATH" right after the file is written — keeps the plaintext DB password out of reach of every other local user/process on the box, while still letting the collector (running as that same user) read its own config. Would rather see that than leaving it on default permissions, since this file has a real credential in it.

recipes/newrelic/infrastructure/nrdot/oracle-otel/oci-liunx.yml
Nandu-pns: displayName/description say "RHEL/OEL" but installTargets only lists `oracle` (OCI) — could you narrow the description to match (e.g. "Oracle Linux (OCI)") so the catalog listing isn't misleading? Can broaden both together later once tested on plain RHEL/CentOS.

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-debian.yml
RamanaReddy8801: Fixed — moved the validation up to fire right after `LOGIN_NAME` is read, before `LOGIN_UPPER` is computed or the `EXISTS_RESULT` lookup runs. The original check after the case-1/case-2 block stays too, since it still needs to independently validate a fresh value entered via "create new username." Verified on real RDS: an injection payload at the very first prompt is now rejected before `sqlplus` is invoked at all (confirmed via `/proc/<pid>/cmdline` polling — zero `sqlplus` processes ever spawned).

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-rhel.yml
RamanaReddy8801: Fixed — moved the validation up to fire right after `LOGIN_NAME` is read, before `LOGIN_UPPER` is computed or the `EXISTS_RESULT` lookup runs. The original check after the case-1/case-2 block stays too, since it still needs to independently validate a fresh value entered via "create new username." Verified on real RDS: an injection payload at the very first prompt is now rejected before `sqlplus` is invoked at all (confirmed via `/proc/<pid>/cmdline` polling — zero `sqlplus` processes ever spawned).

recipes/newrelic/infrastructure/nrdot/oracle-otel/oci-linux.yml
RamanaReddy8801: Added `chown nrdot-collector:nrdot-collector + chmod 600 `as suggested. Verified on real RDS this time: `ls -la `shows `-rw------- nrdot-collector nrdot-collector, journalctl `shows a clean startup with no permission errors (`connected to PDB, Everything is ready`), `systemctl status `shows it actively running, and a plain local user (`ubuntu`, no sudo) correctly gets `Permission denied` trying to read it directly.

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-debian.yml
RamanaReddy8801: Added `chown nrdot-collector:nrdot-collector + chmod 600 `as suggested. Verified on real RDS this time: `ls -la `shows `-rw------- nrdot-collector nrdot-collector, journalctl `shows a clean startup with no permission errors (`connected to PDB, Everything is ready`), `systemctl status `shows it actively running, and a plain local user (`ubuntu`, no sudo) correctly gets `Permission denied` trying to read it directly.

recipes/newrelic/infrastructure/nrdot/oracle-otel/rds-rhel.yml
RamanaReddy8801: Added `chown nrdot-collector:nrdot-collector + chmod 600 `as suggested. Verified on real RDS this time: `ls -la `shows `-rw------- nrdot-collector nrdot-collector, journalctl `shows a clean startup with no permission errors (`connected to PDB, Everything is ready`), `systemctl status `shows it actively running, and a plain local user (`ubuntu`, no sudo) correctly gets `Permission denied` trying to read it directly.

recipes/newrelic/infrastructure/nrdot/oracle-otel/oci-liunx.yml
RamanaReddy8801: Fixed — `displayName` is now `NRDOT Oracle Database (Oracle Linux) `and the description reads "...on Oracle Linux (OCI) self-hosted environments," matching the current `platform: oracle`-only `installTargets`.

recipes/newrelic/infrastructure/nrdot/oracle-otel/oci-linux.yml
```diff
@@ -0,0 +1,1124 @@
...
+            SSH_AUTH_SOCK=$(ls /tmp/ssh-*/agent.* 2>/dev/null | head -1)
+            export SSH_AUTH_SOCK
+          fi
+          if [ -z "${SSH_AUTH_SOCK:-}" ]; then
+            echo "SSH agent socket not found." >&2
+            echo "Please connect to this host using SSH agent forwarding: ssh -A <user>@<this-host-ip>" >&2
+            exit 131
+          fi
+
+          VERSION_OUTPUT=$(ssh -o StrictHostKeyChecking=no -o BatchMode=yes "${SSH_USER}@${HOST}" \
```
Nandu-pns: All three `ssh` calls use `-o StrictHostKeyChecking=no`, which skips host-key verification entirely — the connection accepts whatever key the remote end presents, no `known_hosts` check, even if the key changes between runs. That's a MITM risk on a real network, and this connection carries SYS-privileged SQL, so it's worth protecting.

No other recipe in the repo uses SSH like this, so there's no existing pattern to match. One option I see is `-o StrictHostKeyChecking=accept-new`. In general, I tried to check how other tools handle this and i see OpenSSH added it in [7.6](https://www.openssh.org/txt/release-7.6) specifically as a safer alternative to `no` (see the [`ssh_config` man page](https://man.openbsd.org/ssh_config.5#StrictHostKeyChecking) for the exact semantics), and it's also why [Ansible has an open feature request](https://github.com/ansible/ansible/issues/77318) to support it natively. That said, could you double check what fits best here for how this recipe is actually run — accept-new might not be the only reasonable answer depending on constraints I might not be seeing.

RamanaReddy8801: Fixed, switched all three `ssh` calls to `-o StrictHostKeyChecking=accept-new`. Stays fully non-interactive alongside the existing `-o BatchMode=yes `(auto-accepts on first connection, no prompt), but now correctly refuses the connection if the host key changes on a later run - closes the gap `=no` had. Verified end-to-end on a real OCI Oracle Linux instance: all three SSH calls (version check, exists-check, and the full 27-grant `CREATE USER` script) succeeded, collector started cleanly, and metrics confirmed flowing into New Relic.

Nandu-pns (approved):
LGTM

PR #1422, opened 2026-08-31
feat(nrdot-mysql): add MySQL NRDOT recipes (Linux self-hosted + RDS)

RamanaReddy8801:
## Summary
- Adds `nrdot-collector-mysql` recipes for Debian/Ubuntu and RHEL/CentOS, both self-hosted and RDS/Aurora, using the `nrmysql` receiver — mirrors the existing mssql-otel pattern (Standard/Full-feature presets, scoped monitoring-user provisioning, injection-safe credential handling).
- MySQL NRDOT support is Linux-only per docs (no Windows/WinAuth), so this ships 4 recipe files instead of mssql-otel's 8.
- Adds test definitions (`test/definitions/nrdot/mysql-otel/`) and Ansible deploy roles (`test/deploy/linux/nrdot-mysql/`) for CI provisioning.
- Design spec: `docs/superpowers/specs/2026-08-31-mysql-otel-recipe-design.md` (local/gitignored, per repo convention).

All 4 recipes marked `WORK IN PROGRESS - not for use`, consistent with the current mssql-otel recipes' status.

I’ve attached the test [documentation](https://docs.google.com/document/d/1nRgqw-w1_-lAIal17GwUQsR9pATpR1VN6wm02ALOPiQ/edit?tab=t.0) covering several key integration scenarios for your review

### DIFF
--- a/recipes/newrelic/infrastructure/nrdot/mysql-otel/debian.yml
+++ b/recipes/newrelic/infrastructure/nrdot/mysql-otel/debian.yml
@@ -396,11 +396,8 @@ install:
               collection_interval: 10s
               initial_delay: 1s
               tls:
-                insecure: false
-                insecure_skip_verify: false
-                # Required for Amazon RDS/Aurora with SSL/TLS enforcement -- see
-                # Prerequisites -> Network access -> Amazon RDS / Aurora for MySQL.
-                # ca_file: /path/to/global-bundle.pem
+                insecure: true
+                insecure_skip_verify: true
               # "procedure" collects EXPLAIN plans for write statements (UPDATE/DELETE/INSERT/REPLACE)
               # without granting DML to the monitoring user, via a SQL SECURITY DEFINER procedure.
               # Falls back to "inline" (default) automatically if the procedure is absent for a schema.
@@ -652,9 +649,8 @@ install:
               collection_interval: 10s
               initial_delay: 1s
               tls:
-                insecure: false
-                insecure_skip_verify: false
-                # ca_file: /path/to/global-bundle.pem
+                insecure: true
+                insecure_skip_verify: true
               explain_mode: procedure
               statement_events:
                 digest_text_limit: 4096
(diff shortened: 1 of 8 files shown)


Nandu-pns:
Nice work on the credential heredoc and the grant scoping — matches the docs exactly, no more privilege than needed. Two blockers:

1. `NR_CLI_MYSQL_DATABASE` and `TLS_CA_FILE` aren't declared with a default like every other var in this file (and in mssql-otel) — so when unset they render as the literal string `"<no value>"`, not empty. Breaks `database:` on all 4 recipes, breaks TLS on both RDS variants. Just needs a `default:` line like the rest.
2. `password=${ROOT_PASSWORD}` in the generated `my.cnf` is unquoted — MySQL's parser treats `#` as a comment and trims whitespace, so passwords with either get silently mangled (verified against a real 8.0 instance). mssql-otel avoids this entirely by passing `sa`'s password via `SQLCMDPASSWORD` env var instead of writing a file — might be worth doing the same here instead of just quoting.

Nandu-pns:
I see CI checks are failing on the latest commit (da8e275) — both Test Deploy Recipe jobs (RHEL9, Ubuntu22) error out with:

`install error encountered: no default value for environment variable NR_CLI_MYSQL_DATABASE and none provided`

Looks like this is because default: "" on `NR_CLI_MYSQL_DATABASE` doesn't actually count as a default to newrelic-cli — it seems to treat an empty string the same as no default at all.

oracle-otel's `NR_CLI_ORACLE_PDB_NAME` is the working pattern for an optional/blank var — no default: key at all, just name:/prompt:, with the script doing its own [ -z "$PDB_NAME" ] check. Worth dropping the default: "" line and following that instead.

Everything else here (credential heredoc, grant scoping, RDS variant) looks good — just this one and we should be good to go.

Nandu-pns:
Self note: As discussed offline with @RamanaReddy8801, the `NR_CLI_MYSQL_DATABASE` has been entirely dropped owing to few installation issues, and instead this shall be handled in the Install Flow UI adding a note detailing this.

Nandu-pns (approved):

PR #1423, opened 2026-08-31
feat(nrdot-postgresql): add PostgreSQL NRDOT recipes (Linux self-hosted + RDS)

RamanaReddy8801:
## Summary
- Adds `nrdot-collector-postgresql` recipes for Debian/Ubuntu and RHEL/CentOS, both self-hosted and RDS/Aurora, using the `nrpostgresql` receiver — mirrors the mysql-otel/mssql-otel pattern.
- PostgreSQL NRDOT support is Linux-only per docs (no Windows), so this ships 4 recipe files, same shape as mysql-otel.
- Adds test definitions (`test/definitions/nrdot/postgresql-otel/`) and Ansible deploy roles (`test/deploy/linux/nrdot-postgresql/`) — RHEL role explicitly configures `pg_hba.conf` for password-based host auth, which PGDG's default doesn't permit (Debian's default already does).

All 4 recipes marked `WORK IN PROGRESS - not for use`.

- I’ve attached the test [documentation](https://docs.google.com/document/d/12h0LnnpTX7T11yclit4P5QlAttLcaQ1UaV9v2wgMJC8/edit?tab=t.0) covering several key integration scenarios for your review

### DIFF
--- a/recipes/newrelic/infrastructure/nrdot/postgresql-otel/debian.yml
+++ b/recipes/newrelic/infrastructure/nrdot/postgresql-otel/debian.yml
@@ -323,13 +323,7 @@ install:
             exit 1
           fi
 
-          if [ "$PRESET" = "2" ]; then
-            COLLECTION_INTERVAL="30s"
-            LIMIT_MIB=500
-          else
-            COLLECTION_INTERVAL="15s"
-            LIMIT_MIB=200
-          fi
+          COLLECTION_INTERVAL="15s"
 
           DATABASES_YAML=""
           OLD_IFS="$IFS"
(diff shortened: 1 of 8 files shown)


Nandu-pns:
This one's clean — pg_hba edit is correct, credentials never hit argv/disk, grants match docs. One blocker:

`SERVER="${NR_CLI_POSTGRES_SERVER:-localhost}"` on the self-hosted recipes is shell expansion, not the Go template — `newrelic-cli` never exports recipe vars into the shell env, so this always falls back to `localhost` no matter what's entered at the prompt. Your own RDS file does it right: `rds-debian.yml:113` uses `{{.NR_CLI_POSTGRES_SERVER}}`. Same swap needed in 3 spots each on debian.yml/rhel.yml.

Nandu-pns (approved):

PR #1434, opened 2026-09-15
add multi-instance RDS support to database NRDOT recipes

RamanaReddy8801:
## Summary
Adds multi-instance support across NRDOT database recipes so one recipe run can configure a single collector to monitor multiple database instances, instead of requiring a separate install per instance.

Scope, tracked as it lands on this branch:
- [x] Oracle RDS (`rds-debian.yml`, `rds-rhel.yml`) — instances YAML + secrets file
  input, per-instance skip-and-continue, multi-receiver config generation following
  New Relic's documented YAML anchor/merge-key pattern
  (docs.newrelic.com/.../otel-oracledb/#rds-multi-receiver-config).
- [x] Oracle self-hosted (`oci-linux.yml`) — multiple co-located instances on one host
- [x] MySQL (RDS + self-hosted)
- [x] PostgreSQL (RDS + self-hosted)
- [x] MSSQL (RDS + self-hosted)

## Design notes (Oracle RDS)
- Two new prompts replace the old single-instance ones: an instances file (host/port/service/login_name per instance, non-secret) and a secrets file (per-instance RDS master credentials, read directly rather than pre-exported).
- A bad instance (wrong password, unsupported version, failed user setup) is skipped with a logged reason; the rest still get configured. The install only hard-fails if zero instances survive.
- Collector config: 1 instance gets a plain receiver; 2+ instances share the metrics/events block via a YAML anchor (`&nroracledb-rds-common`) with later receivers using `<<: *nroracledb-rds-common` plus only their own endpoint/username/password/service — matching New Relic's own documented pattern.

## Test plan
- [x] Local dry-run of every task's shell logic against fixture instances/secrets files with mocked `sqlplus`: happy path, one bad instance (skip-and-continue), zero survivors, both config presets, interactive existing-user prompt, single vs. multi-instance config generation — verified via `yaml.safe_load` including real merge-key resolution.
- [x] Verified against 2 real AWS RDS Oracle instances via `newrelic install -c <path>`; found and fixed a real `mvdan/sh` shell incompatibility (`unhandled redirect op: <&`) that only surfaced on the live run.
- [x] Re-run full manual scenario list against real infra for each additional recipe as it's added.

I’ve attached the test [documentation](https://docs.google.com/document/d/1KTJPtdAy6Zv1LvA_mnU6ZIHjx--boymHHjiqipTWxRU/edit?tab=t.0) covering several key integration scenarios for your review


Nandu-pns (changes requested):
Nice pattern overall — instances-file + secrets-file + skip-and-continue is the right shape for this, and the multi-receiver YAML anchor setup checks out. Two things need fixing before we can proceed further:

**1. The secrets file gets `source`d as root instead of parsed — that's arbitrary code execution.**
In Oracle, MySQL, PostgreSQL, and MSSQL-on-Linux, we do this (root-only recipes, so it runs as root):
```sh
# shellcheck disable=SC1090
. "$SECRETS_FILE"
```
If that file (or its directory) is writable by anyone but the current user, that's root code exec, not just a bad secrets format. The permission check we have only warns and proceeds anyway. `windows.yml` already parses this safely (`if ($_ -match '^([^=]+)=(.*)$') {...}`) — can we do the same `KEY=VALUE` parsing on the Linux side instead of sourcing, and make the permission check a hard failure?

**2. All the old single-instance recipe inputs were removed with no default/fallback — that's an unannounced breaking change.**
`_SERVER`/`_PORT`/`_PASSWORD`/`_LOGIN_NAME`/`_MASTER_USER` etc. are gone across all four DB families, replaced by two new required file-based prompts with no default. Any existing scripted/`-y` install using the old vars gets them silently ignored, then hard-fails on "instances file not found." Can we either keep the old vars as a deprecated-but-supported path, or call this out explicitly (release notes / README) so it's not a silent break for existing customers?

Happy to go another round once these two are addressed.

Nandu-pns:
One more thought, non-blocking — probably still worth a look though: there are likely a lot of users just monitoring a single DB, and with this change they now have to author and manage two files instead of answering a few prompts. Could we keep the interactive prompts as a fast path when there's only one instance, and only require the file-based flow once there's more than one? Just my take though — happy to defer to you all on what's actually best for the single-instance install experience.

Nandu-pns (approved):

PR #1437, opened 2026-09-21
fix(test): widen nginx-plus AMI name filter version wildcard

Nandu-pns:
## Summary

Fixes the `nginx-linux2-ami` smoke test (`test/definitions/ohi/linux/nginx-linux2-ami.json` and its EU/JP copies), which was failing on every run because its `ami_name` filter matched zero AMIs.

## Changes

**1. Target the real, currently-published NGINX Plus AMI family.**

```diff
- "ami_name": "nginx-plus-amazon-linux-2-v1.??-x86_64-developer-*"
+ "ami_name": "nginx-plus-amazon-linux-20*-v*-amd64-developer-prod-*"
```

NGINX Plus no longer publishes AMIs for plain `amazon-linux-2` — the current catalog only has `amazon-linux-2023`, and versions `v2.1`+ use `amd64` instead of `x86_64` for the architecture label. Verified against the real AWS Marketplace AMI catalog (`aws ec2 describe-images`) rather than assumed. The OS-release year is wildcarded (`20*`) rather than hardcoded to `2023`, so a future Amazon Linux release year won't reproduce this same failure. The `amd64` architecture segment stays literal on purpose — wildcarding it (e.g. `*64`) would also match `arm64`, which can't boot on this test's `t3.medium` (x86-only) instance.

**2. Disambiguate the resource id across the 3 copies of this test.**

```diff
- "id": "host1"
+ "id": "host1-eu"   (EU copy)
+ "id": "host1-jp"   (JP copy)
+ "id": "host1-base" (base ohi/linux copy)
```

All 3 copies used the identical resource id and land in the same underlying VPC when run together in CI, causing a security-group name collision (`InvalidGroup.Duplicate`) between whichever two happened to race closest together. `debian12-infra.json` already established this exact pattern for its own EU/JP duplication (`host1-eu`/`host1-jp`); this extends the same convention here, plus a distinct suffix for the base copy too, since any pair of the three can collide, not just EU vs JP.

**3. Tolerate a yum repo that no longer exists on the new AMI.**

```diff
  - name: Disable nginx-amplify permanently (if present)
    shell: "yum-config-manager --disable nginx-amplify"
    become: yes
+   register: disable_amplify
+   changed_when: disable_amplify.rc == 0
+   failed_when: disable_amplify.rc != 0 and 'No matching repo to modify' not in disable_amplify.stderr
```

The old `amazon-linux-2` NGINX Plus AMI shipped a `nginx-amplify` yum repo that this test's `yum-fix` role explicitly disables. The new `amazon-linux-2023` AMI (from change #1) doesn't ship that repo at all, so the unconditional disable command failed outright (`Error: No matching repo to modify: nginx-amplify.`). Now only fails on an unexpected error — "nothing to disable" is treated as success.

**4. Add Amazon Linux 2023 support to the `nginx-open-source-integration` production recipe.**

```diff
    - type: host
      os: linux
      platform: "amazon"
      platformVersion: "2"
+   - type: host
+     os: linux
+     platform: "amazon"
+     platformVersion: "(2023\\.*)"
    - type: host
      os: linux
      platform: "redhat"
```

`recipes/newrelic/infrastructure/ohi/nginx/linux.yml` — this is a **production recipe**, used by real customers, not test-only config. Its `installTargets` restricted `platform: "amazon"` to `platformVersion: "2"` only, so on Amazon Linux 2023 hosts the recipe was never even evaluated (confirmed via CI logs: only `infrastructure-agent-installer` showed up in `ScriptEvaluation`, never `nginx-open-source-integration`). This is why the base `ohi/linux/nginx-linux2-ami.json` job kept failing even after changes #1–#3 got EU/JP green — those two copies install via a hardcoded recipe URL and validate on a generic "installation complete" string, while the base copy specifically validates that the NGINX Integration was installed and running.

Before applying this, checked real customer install activity for `platformVersion: "2023"` via NRQL against `VirtuosoInstallRecipe` (account `10827034`, 90-day window, faceted by `hostPlatformVersion`) rather than assuming it's safe to add — AL2023 install volume is substantial, comparable to or exceeding AL2 itself. The regex pattern mirrors the exact precedent already shipped for the same `platform: "amazon"` + AL2023 case in `recipes/newrelic/infrastructure/ohi/mysql/rhel.yml`.

## Blocker resolved: AWS Marketplace subscription

`RunInstances` was also failing with `OptInRequired` even after the fix above, since this account had never subscribed to the `NGINX Plus Developer - Amazon Linux 2023` Marketplace product (`prod-2zvigd3ltowyy` — confirmed this is exactly the product backing our AMI, matching the `image_location` on the resolved image). Subscribed as of Sep 21, 2026 (usage-based, free 30-day trial, cancel-anytime — no annual commitment).

## Test plan

- [x] AMI resolution confirmed working in CI (real image found, matches a genuine NGINX Plus AMI)
- [x] Security-group collision confirmed resolved (no longer fails with `InvalidGroup.Duplicate`)
- [x] Marketplace subscription confirmed active
- [x] EU and JP `Test Deploy Recipe` jobs confirmed fully passing
- [x] Base `Test Deploy Recipe` job confirmed fully passing with the recipe fix (change #4)

---

**P.S. — why subscribing, rather than dropping NGINX Plus for a free base image**

Checked whether there's a way to avoid needing the subscription at all before going this route:
- This is the **only** test in the entire repo that uses `nginx-plus-...` (the commercial product) — every other nginx test here (manual, NRDOT, OHI) uses a free base image (`amzn2-ami-hvm`, `al2023-ami`, `ubuntu-pro-server`, etc.) and installs open-source nginx itself via the package manager. So there's no already-subscribed fallback anywhere else in the repo to borrow.
- This specific test's `yum-fix` role explicitly disables the `nginx-amplify` yum repo — that repo only exists because it's pre-baked into the official NGINX Plus AMI. That's a strong signal this test was deliberately built to validate against the commercial NGINX Plus product (and a known repo conflict specific to it), not "nginx in general." Swapping to a free base image would mean adding an nginx-install step to the role and would change what this test actually covers, not just its AMI source.
- Given that, subscribing preserves the test's original intent. The offer itself is low-risk for this use case: usage-based pricing (~$0.14/hr), a free 30-day trial, no end date, cancelable anytime — not an annual commitment.


Nandu-pns:
FYI — this PR also unblocks 2 of the 5 currently-failing checks on [newrelic-cli#1890](https://github.com/newrelic/newrelic-cli/pull/1890):

- `EU-nginx-linux2-ami.json`
- `JP-nginx-linux2-ami.json`

Confirmed via the actual job logs (not assumed) — both fail with the exact same root cause fixed here: the deploy config resolves `test/definitions-eu(-jp)/smoke/nginx-linux2-ami.json` and `recipes/newrelic/infrastructure/ohi/nginx/linux.yml` straight from `open-install-library`'s `main` branch at test-run time (no pinned commit/submodule), still carrying the pre-fix `ami_name` filter (`nginx-plus-amazon-linux-2-v1.??-x86_64-developer-*`). `ec2_ami_info` returns `"images": []`, and the same `.last` on an empty sequence crash happens (`No last item, sequence was empty.`) that this PR's change #1 fixes. Since that repo is cloned live from `main` rather than pinned, no changes are needed on the CLI side — merging this PR resolves those 2 checks there once `main` picks it up.

The other 3 failing checks on that PR (`EU-windows2019.json`, `JP-windows2019.json`, `infra-win2016.json`) are **not** related to this fix — they fail on a distinct Windows Infra Agent install-validation issue (`Failed to validate that Infrastructure Agent\s+(installed) is in output`), unrelated to nginx/AMI resolution.

pranav-new-relic (approved):

